PR #110, opened 2016-01-14
Do not fetch artifacts with a classifier from Ivy repositories if not asked to do so

alexarchambault:


alexarchambault:
Note to self: merge after having added non regression tests for that

alexarchambault:
Included in https://github.com/alexarchambault/coursier/pull/124.

PR #132, opened 2016-02-04
Better handling of info attributes

alexarchambault:
(These are attributes - things added to organization / name in `Module` - starting with `info.`.)

- Drop them from dependencies originating from Ivy metadata if asked to do so (dropInfoAttributes field of IvyRepository)
- Enable dropping them by default on ~/.ivy2/{cache,local}, and Ivy repositories in SBT plugin
- Add --drop-info-attr CLI option to enable dropping them from the command-line tools
- Test these with org.scala-js:sbt-scalajs;sbtVersion=0.13;scalaVersion=2.10:0.6.6


alexarchambault:
Allows to load SBT projects with the scalajs 0.6.6 plugin enabled (it was fine up to 0.6.5).

PR #142, opened 2016-02-16
Fix in launcher class loader isolation

alexarchambault:
Fixes https://github.com/alexarchambault/coursier/issues/141

### DIFF
--- a/cli/src/main/scala-2.11/coursier/cli/Coursier.scala
+++ b/cli/src/main/scala-2.11/coursier/cli/Coursier.scala
@@ -230,7 +230,7 @@ case class Launch(
   val contextLoader = Thread.currentThread().getContextClassLoader
 
   val parentLoader0: ClassLoader =
-    if (Try(contextLoader.loadClass("coursier.Launch")).isSuccess)
+    if (Try(contextLoader.loadClass("coursier.cli.Launch")).isSuccess)
       Launch.mainClassLoader(contextLoader)
         .flatMap(cl => Option(cl.getParent))
         .getOrElse {


paulp:
If I may be so bold, this sort of bug is an inevitability when there are class names in string literals. It also defeats all kinds of useful static analysis like proguard unless you have a means to inform proguard of the class names which may be created reflectively, which implies organized management of those names.

If you can't avoid working in terms of a class name, then at least isolating the reflective code and working with something stronger than a string literal would help a lot. And of course anything which must be create by name needs a test behind the name.

(Sorry that this commentary is 6x as long as the commit, especially given that you're fixing the bug which I reported! But I think it's more important that it remain fixed a month from now than that it be fixed today.)

alexarchambault:
I agree that this name should be obtained in a better way... (at compile time, by a macro, say).

I don't want it to be preserved by proguard though. The modified line looks for this class with a fqn, in order to know whether the launcher was proguarded (the fqn should have been obfuscated, so the class can't be found, and there's no need to isolate class loaders) or not (coursier and its libs are in the classpath, non obfuscated, so we need isolation). Reflection is kept at a minimum here.

paulp:
I mentioned proguard only as a common example of why it's important/useful to keep a tight leash on all logic which circumvents the normal communication paths amongst bytecode, and so not having classnames sprinkled around in source code string literals. Sorry, it didn't make for the clearest example.

alexarchambault:
@paulp I'm merging this for now, but I think I'll try to ensure no literals hang around for reflection, before 1.0 (https://github.com/alexarchambault/coursier/issues/144).

alexarchambault:
@paulp Were you having a precise idea on ways to circumvent that? I was thinking of writing a tiny lib in Java, with interfaces having the methods looked for by reflection, along with some helpers to do that (find the methods by reflection). The problem being that I don't want to try to cast things as the interface, as it may be loaded by different class loaders.

paulp:
My objection is the way the string containing the class name is decoupled from the actual name of the class, allowing one to change without the other and no way to know it. It looks to me like this bug was/is avoidable if instead of `contextLoader.loadClass("coursier.cli.Launch")` it was `contextLoader.loadClass(this.getClass.getName)`.

alexarchambault:
Ah, I see. The goal here is to detect when the fully qualified class name, as it is known at compile time (`"coursier.cli.Launch"`), differs from the runtime one (`getClass.getName`). They should be identical, except when the coursier launcher is proguarded. I now think I should have tested that with just `getClass.getName != "coursier.cli.Launch"`, instead of a `loadClass` attempt.

PR #158, opened 2016-02-22
Extract npm dependencies to package.json file and use it for travis

easel:
- Allows developers to init simply using `npm install`


alexarchambault:
That's great, thanks!

PR #172, opened 2016-02-24
Force versions of the usual org.scala-lang modules, from the SBT plugin

alexarchambault:


alexarchambault:
As underlined by @dwijnand, this amounts to https://github.com/sbt/sbt/issues/2286. This is of more usefulness here, as coursier tends to pick higher versions as SBT. (coursier keeps the highest required by its dependees, no matter their depth from root dependencies.)

alexarchambault:
@dwijnand Do you know of particular reasons not do that? It seems natural to me, I actually thought SBT was doing it by default.

alexarchambault:
A SBT key could possibly be added to disable that.

dwijnand:
What you've doesn't look insane to me.

Rather than provide a blunt sbt key to revert back I recommend exploring @paulp's idea around allowing users to configure on a dependency-by-dependency basis (or perhaps the set result of applying a predicate) how version conflict resolution happens (if I understood him right).

paulp:
I'd certainly hope to avoid a dependency-by-dependency basis except in pathological scenarios - but it ought to be possible, and it would be if the logic could be provided by the project.

The notion that all the scala-* jars ought to have the same version is the sort of constraint which is a precondition for correctness and easy to express if you can write code, yet impossible to express if the form is limited to "group:artifact:id". At that point you're down to hoping the ad hoc hooks offered by the dependency manager are sufficient to at least reject invalid combinations (as being able to fix the versions across the board is too much to hope for) or crossing your fingers it will be special cased in a way which doesn't burn you later.

alexarchambault:
Ok, I'm going to merge this then. And keep in mind to add support for more detailed reports of these possibly non obvious choices that were made during resolution (so including possible discardings of versions of these org.scala-lan artifacts).

Thanks to the both of you!

PR #189, opened 2016-03-05
Add support for custom URL protocols, change the way cache directory is specified

alexarchambault:
The latter is a source / binary compatibility breaking change in particular. It makes specifying the cache directory easier, which makes it worth the compatibility breaking.


alexarchambault:
A point about the custom URL handlers should be added in the README. Namely, to add a handler for protocol `test`, one should ensure a class `coursier.cache.protocol.TestHandler` extending `java.net.URLStreamHandler` and having a no-arg constructor is present in the classpath.

Adding this handler under `sun.net.www.protocol`, as suggested by the `java.net.URL` sources doesn't seem to work as expected. In particular, it requires the handlers to be available in the same `ClassLoader` as `java.net.URL`, which should not be the case for handlers added as dependencies with the coursier launcher for example.

PR #212, opened 2016-03-31
Add support for fallback artifact (from) and excludeDependencies in SBT plugin

alexarchambault:


alexarchambault:
Fixed https://github.com/alexarchambault/coursier/issues/196, https://github.com/alexarchambault/coursier/issues/199.

PR #213, opened 2016-03-31
Basic property substitution in version in POMs

alexarchambault:


alexarchambault:
Fixes https://github.com/alexarchambault/coursier/issues/210.

alexarchambault:
@pfn I had to ensure coursier gives the maven-plugin packages the .jar extension (https://github.com/alexarchambault/coursier/pull/213/commits/f167b36c1c3416641817b9e2fe7f30d581c87c54), for https://github.com/pfn/android-javacv-sample to be fine. I guess that's their actual extension, right?

pfn:
@alexarchambault yes, that's correct, `maven-plugin` packaging = `jar`; however, they do not get exported to classpath by default (but if coursier is only doing resolution/fetching, shouldn't matter); they only get exported to classpath if one sets `classpathTypes += "maven-plugin"`

alexarchambault:
Seriously? I didn't know about that one :). Opened https://github.com/alexarchambault/coursier/issues/214 for it.

PR #223, opened 2016-04-09
Ensure returned resolution artifacts are unique

alexarchambault:
Should hopefully fix issues sometimes seen on CIs using sbt-coursier


alexarchambault:
Should fix https://github.com/alexarchambault/coursier/issues/208, also seen on https://ci.appveyor.com/project/alexarchambault/coursier/build/536, https://ci.appveyor.com/project/alexarchambault/coursier/build/538, https://ci.appveyor.com/project/alexarchambault/coursier/build/539. The problem originates from coursier trying to download several times in parallel a given URL, and failing because of locks (there are no re-attempt to download / check things if they are locked for now).

PR #238, opened 2016-04-22
Add cross version support with excludeDependencies

### LINKED ISSUE #235: Support cross versions in excludeDependencies
The recent addition of excludeDependencies is lacking support for cross versioning. In particular, the syntactic sugar of defining an exclusion as `"io.kamon" %% "kamon-autoweave"` results in a message stating that it isn't supported. While it is easy to workaround this, it's a bit of a hassle and can subtly break current sbt builds.

jalaziz:
Fixes #235


alexarchambault:
Looks great, thanks!

PR #259, opened 2016-05-25
Prefer version from dependency rather than from POM...

alexarchambault:
...when building artifact URL


alexarchambault:
There are actually good reasons to prefer the one from the project metadata (the version from the dependency may be an interval for example, which wouldn't be fine as is in the path of the artifact). I'll re-open with a better fix.

PR #289, opened 2016-06-27
Add early internals document draft

alexarchambault:


alexarchambault:
I'd like to fill the holes and illustrate with inlined examples next (with tut).

PR #296, opened 2016-07-05
Update to sbt 0.13.11

VEINHORN:


alexarchambault:
The SBT version is still 0.13.8 to ensure the SBT plugin is compatible at least from this version (e.g. it's not compatible with SBT 0.13.5). I put this a bit on hold as I'm not sure we'll still be compatible with it with this change, and no tests ensure that. I think I'll add integration tests for it. I'll merge after that.

VEINHORN:
@alexarchambault :+1:

antonkulaga:
There is already 0.13.2 avaliable

alexarchambault:
I finally switched to the latest sbt (`0.13.15` currently), to bump to a more recent scala `2.10` (for the auto type class derivations now happening in [`Travis.scala`](https://github.com/coursier/coursier/blob/e6c28f02d6f096029c800cbb83233fac5edf2c57/project/Travis.scala) to be fine). It's safer now, during the RC cycle (and I guess more people switched to the subsequent versions of sbt too...)

PR #343, opened 2016-09-04
[WIP] Feature/better spark submit

darkjh:
Inject extra classpath needed to run spark on yarn in runtime. Use a direct call to `SparkSubmit`.

WIP, need to be tested.

### DIFF
--- a/spark/src/main/scala/coursier/cli/SparkSubmit.scala
+++ b/spark/src/main/scala/coursier/cli/SparkSubmit.scala
@@ -1,7 +1,8 @@
 package coursier.cli
 
-import java.io.{ BufferedReader, File, InputStream, InputStreamReader, OutputStream }
-import java.nio.file.{ Files, Paths }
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/coursier/cli/SparkSubmit.scala
```diff
@@ -29,19 +33,30 @@ case class SparkSubmit(
...
     // must be fine with spark 2.10 too, ...)
-    val dir = new File(sparkHome + "/assembly/target/scala-2.11")
-    Option(dir.listFiles()).getOrElse(Array.empty).filter { f =>
-      f.isFile && f.getName.endsWith(".jar")
-    } match {
-      case Array(assemblyFile) =>
-        assemblyFile.getAbsolutePath
+    // TODO(han) maybe a conf or sys env ???
+    val dirs = List(
+      new File(sparkHome + "/assembly/target/scala-2.11"),
```
darkjh: Actually we could have multiple locations depending on the build.

PR #346, opened 2016-09-08
Rework spark submit command

alexarchambault:


alexarchambault:
cc @darkjh

alexarchambault:
Note to self: release a 1.0.0-M14-1 with the current spark-submit command before merging that.

PR #348, opened 2016-09-12
Publish some test helpers / resources

alexarchambault:


alexarchambault:
Goal of this was to make it easier to possibly re-use the test stuff from other repos (possibly @rtfpessoa 's https://github.com/rtfpessoa/coursier-s3). I may add some S3 ITs using fm-sbt-s3-resolver in this repo instead.

Issue #325, opened 2016-08-15
Can not resolve hbase


hujianxin:
The error message is:

```
[error] 
[error] 1 error(s):
[error]   org.apache.hbase:${compat.module}:1.1.5:
[error]     not found: /home/hujianxin/.sbt/0.13/plugins/target/resolution-cache/org.apache.hbase/${compat.module}/1.1.5/resolved.xml.xml
[error]     Not found
[error]     not found: /home/hujianxin/.ivy2/local/org.apache.hbase/${compat.module}/1.1.5/ivys/ivy.xml
[error]     not found: https://repo1.maven.org/maven2/org/apache/hbase/$%7Bcompat.module%7D/1.1.5/$%7Bcompat.module%7D-1.1.5.pom
[error]     not found: /home/hujianxin/.m2/repository/org/apache/hbase/$%7Bcompat.module%7D/1.1.5/$%7Bcompat.module%7D-1.1.5.pom
[trace] Stack trace suppressed: run last *:coursierResolution for the full output.
[error] (*:coursierResolution) java.lang.Exception: Encountered 1 error(s) in dependency resolution
[error] Total time: 2 s, completed Aug 15, 2016 8:49:33 AM
```

hujianxin:
I ran it in `1.0.0-M13` and `v1.0.0-M12-1`, got the same error.

hujianxin:
But it works on hbase 1.2.2, weird

alexarchambault:
I quickly tried to reproduce that, without success. Do you have more details? Like the dependency that transitively pulled hbase?

longhua:
I got a similar error when I used coursier with the following configuration. However, it worked if I removed coursier. Is there any suggestion? Thanks in advance.

```
libraryDependencies ++= Seq(
  "org.apache.spark" %% "spark-streaming" % "1.6.2" % "provided",
  "org.apache.spark" %% "spark-sql" % "1.6.2" % "provided",
  "org.apache.spark" %% "spark-streaming-kafka" % "1.6.2" exclude("org.spark-project.spark", "unused"),
  "org.elasticsearch" %% "elasticsearch-spark" % "2.1.2" intransitive(),
  "org.apache.phoenix" % "phoenix-spark" % "4.7.0-HBase-1.1" % "provided",
  "com.typesafe" % "config" % "1.3.0",
  "com.github.nscala-time" %% "nscala-time" % "2.6.0",
  "joda-time" % "joda-time" % "2.9.1"
)
[...]

PR #360, opened 2016-10-16
Add support for property absence directives in profile activation

eboto:
Addresses the lack of support described by @alexarchambault in #325 for profiles activations like: 

```xml
<activation>
  <property>
    <name>!hadoop.profile</name>
  </property>
</activation>
```


alexarchambault:
Looks great, thanks!

FYI, I think I'll make a release in the coming days.

alexarchambault:
@eboto `1.0.0-M14-4` with that just published. I'll tell the world more loudly about those changes, and the other since `1.0.0-M14`, for `1.0.0-M15`. (`1.0.0-M??-?` releases are just for convenience).

PR #381, opened 2016-11-07
Fixes

alexarchambault:


alexarchambault:
Seems to fix https://github.com/alexarchambault/coursier/issues/379 (from the CLI tools at least).

PR #384, opened 2016-11-09
Added Quill to the list of projects using coursier.

mxl:


alexarchambault:
Thanks for that!

I wasn't too keen in putting plugin users under this section until now. I may move it to a separate one.

PR #389, opened 2016-11-13
scaladex cli api change & use scalaVersion from options

MasseGuillaume:
* Scaladex was updated to allow developers to specify cli artifact. It
  will only return projects and artifacts with cli enable

https://github.com/scalacenter/scaladex/commit/c2b865ebd10a280d8bf085d80f7a62e89b5778a2


alexarchambault:
Thanks! I think I'll still add / keep a fallback for when a project has no cli modules. That kind of stuff can be detected automatically too.

PR #396, opened 2016-11-18
Add support for Spark 2 in spark-submit command

alexarchambault:


jbpettitTeads:
unrelated but related @alexarchambault I get this Execption while trying to save a file from a notebook within spark, could it be a missing jar here? 
```
java.lang.RuntimeException: java.lang.RuntimeException: java.lang.ClassNotFoundException: Class org.apache.hadoop.mapred.DirectFileOutputCommitter not found (java.lang.RuntimeException: java.lang.ClassNotFoundException: Class org.apache.hadoop.mapred.DirectFileOutputCommitter not found)
```

PR #426, opened 2017-01-23
Add shading plugin

alexarchambault:
This PR adds an extra SBT plugin that allows to shade dependencies.

Use like
```scala
lazy val myProj = project
  .enablePlugins(coursier.ShadingPlugin)
  .settings(
    shadingNamespace := "myproj.shaded",
    libraryDependencies += "com" %% "foo" % "0.2.3" % "shaded"
  )
```

`sbt myProj/shading:publish` then publishes `myProj` with dependency towards `com::foo` shaded .

Once published, coursier would use it to shade `fastparse`, to prevent conflicts with other versions of it (like in https://github.com/alexarchambault/coursier/issues/395), and not to have coursier `1.0` final depend on a `0.x` dependency.

It would be cool to have `sbt-ensime` use it to shade coursier itself, to prevent conflicts like https://github.com/alexarchambault/coursier/issues/417. (ping @fommil)


fommil:
+1

appveyor.yml
```diff
@@ -14,15 +14,27 @@ install:
   - cmd: SET PATH=C:\sbt\sbt\bin;%JAVA_HOME%\bin;%PATH%
   - cmd: SET SBT_OPTS=-XX:MaxPermSize=2g -Xmx4g
```
fommil: sbt now supports `.sbtopts` on windows as of 0.13.13

fommil: feel free to take our config https://github.com/ensime/ensime-server/blob/2.0/appveyor.yml

alexarchambault: Good to know! I've been trying to stick to sbt 0.13.8 for now though, to be sure sbt-coursier is fine with it.

fommil: it's the 0.13.13 launcher that matters, you can still use it to launch 0.13.8

PR #428, opened 2017-01-31
Update README

alexarchambault:


benmccann:
"an" is actually the correct usage because "sbt" is pronounced "ess bee tee" and it makes more sense to say "an ess bee tee" than "a ess bee tee"

See http://english.stackexchange.com/questions/1016/do-you-use-a-or-an-before-acronyms

alexarchambault:
@benmccann Thanks for the reminder, that's why this wasn't merged already!

The discrepancy comes from things that were committed in `README.md`, but not in `doc/README.md` (the former being generated from the latter by tut), I'll fix it.

PR #430, opened 2017-02-03
Tweak artifact type handling for Maven repos

alexarchambault:
Commit `Better handling of optional dependencies` fixes https://github.com/alexarchambault/coursier/issues/274, https://github.com/alexarchambault/coursier/issues/378.


alexarchambault:
Note to self: add non reg tests for the last commit (check that depending on the `optional` config of `org.scala-lang:scala-compiler` indeed bring `jline:jline`).

PR #434, opened 2017-02-14
Ensure publishing is done with Java 6 with sbt-javaversioncheck

alexarchambault:


alexarchambault:
Bad idea actually, Java 7 is required to compile coursier (even though it won't use Java 7 specific stuff if Java 6 is detected, there are [tests](https://github.com/alexarchambault/coursier/blob/b6295b2719f3dad71f30639404d2149e381a28de/scripts/travis.sh#L170-L176) checking just that).

Issue #157, opened 2016-02-22
Unable to resolve dependencies with no version specified


easel:
Simple to reproduce:

    [coursier] ./coursier fetch  "net.openhft":"chronicle-map":"2.4.12"  

Results in this:
```
  Dependencies:issing file(s) / 14
net.openhft:chronicle-map:2.4.12:default(compile)
https://repo1.maven.org/maven2/org/slf4j/slf4j-api/1.7.6/slf4j-api-1.7.6.pom
  100.0% [##########] 2.6 KiB (14.1 KiB / s)
https://repo1.maven.org/maven2/org/slf4j/slf4j-parent/1.7.6/slf4j-parent-1.7.6.pom
  100.0% [##########] 11.6 KiB (401.0 KiB / s)

Error:
  org.ops4j.pax.url:pax-url-aether::
    Not found: /Users/erik/.ivy2/local/org.ops4j.pax.url/pax-url-aether/ivys/ivy.xml
    Not found: https://repo1.maven.org/maven2/org/ops4j/pax/url/pax-url-aether//pax-url-aether-.pom
  org.ow2.asm:asm::
    Not found: /Users/erik/.ivy2/local/org.ow2.asm/asm/ivys/ivy.xml
    Not found: https://repo1.maven.org/maven2/org/ow2/asm/asm//asm-.pom
  org.xerial.snappy:snappy-java::
    Not found: /Users/erik/.ivy2/local/org.xerial.snappy/snappy-java/ivys/ivy.xml
    Not found: https://repo1.maven.org/maven2/org/xerial/snappy/snappy-java//snappy-java-.pom
  org.codehaus.jettison:jettison::
    Not found: /Users/erik/.ivy2/local/org.codehaus.jettison/jettison/ivys/ivy.xml
    Not found: https://repo1.maven.org/maven2/org/codehaus/jettison/jettison//jettison-.pom
  com.intellij:annotations::
    Not found: /Users/erik/.ivy2/local/com.intellij/annotations/ivys/ivy.xml
    Not found: https://repo1.maven.org/maven2/com/intellij/annotations//annotations-.pom
  com.sun.java:tools::
    Not found: /Users/erik/.ivy2/local/com.sun.java/tools/ivys/ivy.xml
    Not found: https://repo1.maven.org/maven2/com/sun/java/tools//tools-.pom
  com.thoughtworks.xstream:xstream::
[...]

PR #437, opened 2017-02-18
Create Resolution#errors from its errorCache

kzys:
Apparently using Resolution#dependencies hides some errors (e.g. #157).

----

Is this because dependencies are direct dependencies and errorCache contains transitive dependencies?

### DIFF
--- a/cli/src/main/scala-2.11/coursier/cli/Helper.scala
+++ b/cli/src/main/scala-2.11/coursier/cli/Helper.scala
@@ -467,13 +467,13 @@ class Helper(
     errPrintln("\nMaximum number of iterations reached!")
   }
 
-  if (res.errors.nonEmpty) {
+  if (res.metadataErrors.nonEmpty) {
     anyError = true
     errPrintln(
-      s"\nError:\n" +
-      res.errors.map {
-        case (dep, errs) =>
-          s"  ${dep.module}:${dep.version}:\n${errs.map("    " + _.replace("\n", "    \n")).mkString("\n")}"
+      "\nError:\n" +
+      res.metadataErrors.map {
+        case ((module, version), errors) =>
+          s"  $module:$version\n${errors.map("    " + _.replace("\n", "    \n")).mkString("\n")}"
       }.mkString("\n")
     )
   }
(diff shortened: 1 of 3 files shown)


alexarchambault:
👍 , that should always have been done this way... There might have been reasons for this to be this way in the very very early stages of coursier IIRC (I can't recall them though), but I don't think they should apply anymore.

This should unearth errors where the parent POM of a POM is errored, but not the (child) POM itself. It should be possible to add non regression tests for that in [`ResolutionTests`](https://github.com/alexarchambault/coursier/blob/master/tests/shared/src/test/scala/coursier/test/ResolutionTests.scala), for this exact case (POM with a parent POM that can't be found).

Also, it'd be better to keep a method with the former method signature, but deprecated. Its result could even be built in a smart way, by creating `Dependency` instances from the `(Module, String)` that don't have a corresponding `Dependency` in `dependencies`.

kzys:
Okay. Let me add some tests and restore the original `errors`. Should the new method be named as `errors` as well, or something more explicit such as `transitiveErrors` or `allErrors`?

I'm not sure it can be built from `dependencies` because the errors are errorCache elements which *have* a corresponding Dependency in `dependencies`.

kzys:
Ah, there is no way to have 2 errors. The new one has to have the different signature.

kzys:
Back to the original issue (#157), third-party-bom which has a checksum issue is in the direct dependencies.
http://central.maven.org/maven2/net/openhft/chronicle-map/2.4.12/chronicle-map-2.4.12.pom

kzys:
Okay. Now I understand that 1) `dependencies` have transitive ones at the end of `ResolutionProcess`, but 2) these don't contain POM dependencies. That's why `third-party-bom` was not reported correctly.

alexarchambault:
@kzys Using `errorCache` straightaway is what should be done, yes. Indeed, it contains errors not directly associated to dependencies, because of ~pom dependencies (parent POMs, dependency management, in particular).

IMO, the current `errors` methods should be deprecated. A new one, solely based on `errorCache` and returning a `Seq[((Module, String), Seq[String])]` should be added and used instead - it could be called `metadataErrors` say (and simply be renamed to `errors` in subsequent versions, replacing the deprecated method).

Then tests using `metadataErrors` can be added.

alexarchambault:
I see you updated this PR. Looks great! Merging, thanks!

PR #439, opened 2017-02-21
Add sbt launcher

alexarchambault:


dwijnand:
In .travis.yml you're curling an ancient version of sbt-extras.

https://github.com/paulp/sbt-extras/commit/478a364a2d43d6e1ac4451f8b84cdafe43a2e43f

You should either update the SHA you use or drop the curl to rely on what Travis CI provides you.

alexarchambault:
Yeah, it wasn't supposed to be polled from there anyway. That should be fixed now.

PR #457, opened 2017-03-14
Tweak launcher

alexarchambault:


alexarchambault:
Goal of the header in the proguarded JAR is to allow to launch it straightaway as an executable (on Unix systems). This should be used by the [homebrew formula to come](https://github.com/coursier/homebrew-formulas/blob/master/coursier.rb).

alexarchambault:
The `echo` module should be launched during the homebrew formula tests (to ensure the installed launcher works fine).

PR #462, opened 2017-03-18
Profiling...

alexarchambault:


sbt-coursier/src/main/scala-2.10/coursier/ToSbt.scala
```diff
@@ -1,90 +1,103 @@
...
 import coursier.maven.MavenSource
-
 import sbt._
 
 object ToSbt {
 
-  def moduleId(dependency: Dependency): sbt.ModuleID =
+  private def caching[K, V](f: K => V): K => V = {
+
+    val cache = new ConcurrentHashMap[K, V]
```
Jentsch: Just a suggestion, but I think you could use the [TrieMap](http://www.scala-lang.org/api/current/scala/collection/concurrent/TrieMap.html) from the Scala std-lib here. The `getOrElseUpdate` should have the same behaviour as the code below.

alexarchambault: Thanks for the hint. There are people raising good concerns about `scala.collection.concurrent.TrieMap`, like [here](http://42ways.de/blog/2015/02/21/scala-pitfalls-concurrent-hashmap/) or [there](http://stackoverflow.com/questions/29499381/what-is-a-triemap-and-what-is-its-advantages-disadvantages-compared-to-a-hashmap/38700133#38700133) (which should explain why it's not that known). The risk of using it wouldn't be that high here though, I'll probably update the PR with it...

Jentsch: And I was wondering why nobody use this data structure. In particular the performance characteristics in comparison to `ConcurrentHashMap` are interesting.

jvican: Note that the issue with `getOrElseUpdate` was fixed in 2.11.x. https://issues.scala-lang.org/browse/SI-7943

alexarchambault: Not sure the fix made it into `2.10.x` (`2.10.4` in particular, that sbt `0.13.8` depends on).

jvican:
Has this profiling yielded any interesting result?

alexarchambault:
I only did a quick stack analysis when using sbt-coursier from a project with plenty of modules and dependencies. I don't remember the exact numbers, just that way too many of the now cached things from `ToSbt.scala` were created. The first commit addresses that.

The second commit can't hurt, it should avoid some IO. That one isn't motivated by profiling though.

Overall, that seemed to address some issues that @vil1 was seeing on a big project of his.

PR #463, opened 2017-03-18
Use parent's projects resolution result as project cache

vil1:


alexarchambault:
Thanks! I'm going to merge this in the corresponding `topic/profiling` branch, and toy a bit with its then two commits / do some heap stack analysis to see their impact.

PR #477, opened 2017-04-02
Enable sbt plugin hack in Maven repo paths by default

alexarchambault:


alexarchambault:
@jvican Didn't you mention something related on the Gitter of coursier? I've been wanting to enable that by default.

jvican:
Indeed, I did :). Good to see this change.

I finally did this change to my code as well as adding sbt repos by hand.

PR #481, opened 2017-04-04
Support --repository bintray-ivy:org/repo/ in cli.

olafurpg:
This commit adds support for a `bintray-ivy:` prefix to the `--repository`
flag in the cli. This option is equivalent to the `Resolver.bintrayIvyRepo` helper in sbt.
With this new helper, it's possible to write `-r bintray-ivy:scalameta/maven/` instead of

```
-r ivy:https://dl.bintray.com/scalameta/maven/[organisation]/[module]/(scala_[scalaVersion]/)(sbt_[sbtVersion]/)[revision]/[type]s/[artifact](-[classifier]).[ext]
```

### DIFF
--- a/cli/src/main/scala-2.11/coursier/cli/Options.scala
+++ b/cli/src/main/scala-2.11/coursier/cli/Options.scala
@@ -29,6 +29,7 @@ final case class CommonOptions(
   @Short("N")
     maxIterations: Int = 100,
   @Help("Repository - for multiple repositories, separate with comma and/or add this option multiple times (e.g. -r central,ivy2local -r sonatype-snapshots, or equivalently -r central,ivy2local,sonatype-snapshots)")
+  @Value("maven|sonatype:repo|ivy2local|bintray:org/repo|bintray-ivy:org/repo|typesafe-ivy:releases|ivy:pattern")
   @Short("r")
     repository: List[String] = Nil,
   @Help("Source repository - for multiple repositories, separate with comma and/or add this option multiple times")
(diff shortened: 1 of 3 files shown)


olafurpg:
Some questions:

1. Currently, it's necessary to suffix the pattern with `/`, should we handle that automatically so it's optional to leave out the `/` suffix, like `bintray-ivy:scalameta/maven`?
2. How's the best way to test this?
3. Should I document this change in the `--help` page?

alexarchambault:
Thanks!

About your questions:
1. It's already handled automatically for the Maven repos (by `MavenRepository`), it would be better to handle it automatically here too.
2. A small test suite can be added in the `tests` module (`ParseTests`), testing the `bintray-ivy:` syntax (tests for other syntaxes could / should be added later).
3. The `--help` messages are a bit a mess for now... (as is the README IMO). You can add a word about it anywhere if you feel like it, else I'll mention it later if I rework them.

alexarchambault:
Btw, isn't it something weird in your example? It vaguely says it's a Maven repo (`scalameta/maven`), although the pattern shows it isn't...

olafurpg:
> Btw, isn't it something weird in your example? It vaguely says it's a Maven repo (scalameta/maven), although the pattern shows it isn't...

We messed it up by publishing ivy-style to the "maven" repo 😅 We'll fix it soon by publishing maven style.

I added a commit addressing all three points.

alexarchambault:
About the name `bintray-ivy`, maybe something like `bintray-default-ivy` would be better. The corresponding pattern is only the default one (it might be specific to sbt, but I'm not sure about it...), the point of Ivy is being able to support arbitrary custom patterns.

olafurpg:
It's still possible to use a custom ivy pattern with `ivy:`, so I would prefer to keep `bintray-ivy:` which is the parallel to `bintrayIvyRepo` in sbt. However, I can change if you strongly prefer `bintray-default-ivy`.

I've squashed the commits.

alexarchambault:
Ok, let's keep `bintray-ivy`. Merging, thanks again!

Issue #444, opened 2017-02-28
Dependency classifiers are ignored


omervk:
The dependency `"org.apache.spark" %% "spark-sql" % "2.0.2" classifier "tests"` resolves to the same dependency as `"org.apache.spark" %% "spark-sql" % "2.0.2"`.

This is a regression (worked in M15, doesn't in M15-2).

alexarchambault:
Adding
```scala
classpathTypes += "test-jar"
```
fixes that for me.

I don't know if this is actually a bug. Asking for type `test-jar` is something that seems to be done from Maven ([here](https://maven.apache.org/guides/mini/guide-attached-tests.html) or [there](https://maven.apache.org/plugins/maven-jar-plugin/test-jar-mojo.html)). I don't know if this is something Maven requires all the time for these.

This behavior appeared in coursier in `1.0.0-M15-2` after the Maven classifiers and artifact types handling was reworked a bit, hopefully to make it more robust. JARs with classifier `tests` and extension `jar` are now assumed to have type `test-jar` rather than `jar`. `test-jar` type has thus to be added to the classpath to add these JARs. IIRC, I saw the `test-jar` around [this method](https://github.com/sbt/sbt/blob/0.13/sbt-maven-resolver/src/main/scala/sbt/mavenint/MavenRepositoryResolver.scala#L498-L509) in the sources of sbt.

omervk:
Thanks, this worked for me!

trentontrees:
I've created a PR with a test around this: #510
From my testing, this is not resolved in 1.0.0-RC1

PR #510, opened 2017-04-23
Add sbt-test for classifier verfication

trentontrees:
#444 

This is just a test to reproduce the issue of classifiers not being resolved as expected.

### DIFF
--- a/sbt-coursier/src/sbt-test/sbt-coursier/classifiers/build.sbt
+++ b/sbt-coursier/src/sbt-test/sbt-coursier/classifiers/build.sbt
@@ -1,0 +1,13 @@
+scalaVersion := "2.11.8"
+
+libraryDependencies += ("org.jclouds.api" % "nova" % "1.5.9")
+  .classifier("tests")
+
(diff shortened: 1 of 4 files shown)


trentontrees:
This fails the test on my local machine.

I've also tried supplying `classpathTypes += "test-jar"` to the test build.sbt.  That doesnt resolve the build issue.

alexarchambault:
With `classpathTypes += "test-jar"`, the test jar of `org.jclouds.api:nova` lands in the classpath as expected.

The test fails anyway because `org.jclouds.openstack.nova.compute.NovaComputeServiceLiveTest` needs other test jars, to be loaded (the one `org.jclouds:jclouds-compute:1.5.9`, among others). This is  not specific to coursier.

trentontrees:
Confirmed that the test passes as expected when the test dependencies are specifically declared.

`libraryDependencies += ("org.jclouds" % "jclouds-compute" % "1.5.9").classifier("tests")

libraryDependencies += ("org.jclouds" % "jclouds-core" % "1.5.9").classifier("tests")

classpathTypes += "test-jar"`

PR #513, opened 2017-04-24
Switch to scala 2.11.11, 2.12.2

alexarchambault:


alexarchambault:
Split into https://github.com/coursier/coursier/pull/517 and https://github.com/coursier/coursier/pull/518.

PR #527, opened 2017-05-02
Fix excludeDependencies handling

alexarchambault:


alexarchambault:
@fommil That should fix the problem with the `excludeDependencies` you were seeing in ensime-sbt when bumping sbt-sensible.

fommil:
Awesome!

PR #533, opened 2017-05-08
List Support for 0.13.15 in Readme

ChristopherDavenport:
Pretty Simple Readme Edit, but I believe this is already present.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Enable the SBT plugin by adding
 addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.0-RC2")
 ```
 to `~/.sbt/0.13/plugins/build.sbt` (enables it globally), or to the `project/plugins.sbt` file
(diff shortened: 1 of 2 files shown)


alexarchambault:
Thanks! sbt `0.13.16-M1` and `1.0.0-M5` are also supported now.

Can you make the edit in `doc/README.md` first? (Then run `sbt tut` to have it reported to the `README.md` at the root.)

ChristopherDavenport:
A little worried that a binary file ended up in there. How would that have happened?

alexarchambault:
Because you (git-)added it? 😉 It's created by a scripted test, but should be [removed by it too](https://github.com/coursier/coursier/blob/7874f62c419f953b0e399ed0df2c3255e8fb1cc9/sbt-coursier/src/sbt-test/sbt-coursier/from/test#L4).

ChristopherDavenport:
I'll reapply, my apologies

alexarchambault:
Looks good. Merging, thanks!

Issue #424, opened 2017-01-13
Support for dynamic revisions different from SBT / ivy


avdv:
It seems coursier only supports dynamic revisions in major, minor and revision positions of a version, but not anywhere else.

For example `2.5.10-+` is correctly resolvable by vanilla SBT, but coursier spits out:

```
[error]   com.example:base_2.11:2.5.10-+:
[error]     not found:
[error]       /home/claudio/.ivy2/local/com.example/base_2.11/2.5.10-+/ivys/ivy.xml
[error]       https://jcenter.bintray.com/com/example/base_2.11/2.5.10-+/base_2.11-2.5.10-+.pom
[error]       https://repo1.maven.org/maven2/com/example/base_2.11/2.5.10-+/base_2.11-2.5.10-+.pom
```

For the following minimal build file:

```scala
scalaVersion := "2.11.8"
libraryDependencies += "com.example" %% "base" % "2.5.10-+" intransitive()
```

jedesah:
This is probably related to this, `latest.integration` (as in `addSbtPlugin("organization" % "artifact" % "latest.integration")`) doesn't seem to work with coursier.

PR #556, opened 2017-05-31
Support latest subrevision

avdv:
Hi.

Here's my initial try at fixing #424, kindly requesting some thoughts. 

As you can see, the test is still failing.

Why is the `-a1` appended to the upper version bound? Should the revision `1.2.3-+` not be equivalent to `[1.2.3,1.2.4[`?

Note, I will squash the commits when the open questions are resolved.

### DIFF
--- a/core/shared/src/main/scala/coursier/core/Parse.scala
+++ b/core/shared/src/main/scala/coursier/core/Parse.scala
@@ -11,23 +11,28 @@ object Parse {
     else Some(Version(trimmed))
   }
 
+  // matches revisions with a '+' appended, e.g. "1.2.+", "1.2+" or "1.2.3-+"
+  private val latestSubRevision = "(.*[^.-])[.-]?[+]".r
+
   def ivyLatestSubRevisionInterval(s: String): Option[VersionInterval] =
-    if (s.endsWith(".+")) {
-      for {
-        from <- version(s.stripSuffix(".+"))
-        if from.rawItems.nonEmpty
-        last <- Some(from.rawItems.last).collect { case n: Version.Numeric => n }
-        // a bit loose, but should do the job
-        if from.repr.endsWith(last.repr)
-        // appending -a1 to the next version, so has not to include things like
-        // nextVersion-RC1 in the interval - nothing like nextVersion* should be included
-        to <- version(from.repr.stripSuffix(last.repr) + last.next.repr + "-a1")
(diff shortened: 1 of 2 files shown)


alexarchambault:
The `a` stands for alpha (discovered that in the code from aether that the version parsing code here is based on). The goal of the `-a1` is to have `1.2.3+` not match things like `1.2.4-M1`, `1.2.4-RC2`, or `1.2.4-a1`.  The latter versions are all `< 1.2.4`, so would be included in `[1.2.3,1.2.4)`. But they aren't in `[1.2.3,1.2.4-a1)`.

alexarchambault:
(I can squash while merging from the github UI btw.)

avdv:
> The a stands for alpha (discovered that in the code from aether that the version parsing code here is based on). The goal of the -a1 is to have 1.2.3+ not match things like 1.2.4-M1, 1.2.4-RC2, or 1.2.4-a1. The latter versions are all < 1.2.4, so would be included in [1.2.3,1.2.4). But they aren't in [1.2.3,1.2.4-a1).

OK, that was my speculation.

According to https://ant.apache.org/ivy/history/2.2.0/settings/latest-strategies.html the "latest-revision" strategy compares versions similar to [PHP's version_compare](http://php.net/manual/en/function.version-compare.php).

In that case, `dev` is even "lower than" `a` / `alpha`.

And a string different from `[dev, alpha, a, beta, b, RC, rc, pl, p]`has the lowest priority.

> (I can squash while merging from the github UI btw.)

I know, but I'm not sure you can edit the commit message after squashing the commits together... and I would want to clean those up a little after the dust has settled. :smile:

avdv:
WDYT? Is this fix good enough (just fix the test) and try to solve the version comparison later?

alexarchambault:
Yeah, let's just update the test with a `-a1`. This is the expected behavior.

avdv:
Sorry for the delay. I have updated the test and squashed the commits.

alexarchambault:
Thanks!

About the version qualifiers, I think I read somewhere (can't find back the reference) than versions with qualifiers (`dev`, `alpha`, `a`, etc.) just shouldn't be matched by version intervals. Version intervals should only match non-pre-releases. That way, `2.3.+` corresponding to `[2.3.0,2.4.0)` would work as expected, without needing the `-a1` hack. `MavenRepository` should probably be adjusted to match that behavior...

That can be done later, merging this, thanks again!

PR #560, opened 2017-06-01
Switch to sbt 1.0.0-M6 (WIP)

alexarchambault:


project/Settings.scala
```diff
@@ -178,11 +178,91 @@ object Settings {
...
+      }
+    }.evaluated
+  }
+
   lazy val plugin =
     javaScalaPluginShared ++
     divertThingsPlugin ++
     withScriptedTests ++
     Seq(
+      sbtLauncher := {
```
jvican: Why is this required?

sbt-coursier/src/main/scala/coursier/Tasks.scala
```diff
@@ -861,19 +863,24 @@ object Tasks {
 
       val verbosityLevel = coursierVerbosity.value
 
+      val classifiersRes = coursierSbtClassifiersResolution.value
+      val mainRes = coursierResolutions.value
+
       val res =
         if (withClassifiers && sbtClassifiers)
-          Seq(coursierSbtClassifiersResolution.value)
```
jvican: :smile:

project/Settings.scala
alexarchambault: The [default `sbtLauncher` implem](https://github.com/sbt/sbt/blob/6ce09cfed6f3353326981a4e6e07ffba1f15252f/scripted/plugin/src/main/scala/sbt/ScriptedPlugin.scala#L102) expects the sbt launcher to appear first in the update report.  But coursier currently gives no guaranties about the order of the modules and artifacts in its output, so not in the update report either. For whatever reasons, it was in first position anyway before, but isn't anymore now...

sbt-coursier/src/main/scala/coursier/Tasks.scala
alexarchambault: Your fault 😛 But yeah, it's better this way overall.

project/Settings.scala
jvican: Hm, I wonder why it's not the first anymore. Thanks for explaining!

alexarchambault:
(Failing test on Travis is fine locally and on Windows CI.)

Issue #169, opened 2016-02-23
Maven Time based snapshotting is not fully supported


jasongilanfarr:
./coursier fetch org.jitsi:jitsi-videobridge:1.0-SNAPSHOT -r https://github.com/jitsi/jitsi-maven-repository/raw/master/releases/ -r https://github.com/jitsi/jitsi-maven-repository/raw/master/snapshots
  Dependencies:
org.jitsi:jitsi-videobridge:1.0-SNAPSHOT:default(compile)

Error:
  org.igniterealtime:tinder::
    Not found: /Users/jason/.ivy2/local/org.igniterealtime/tinder/ivys/ivy.xml
    Not found: https://repo1.maven.org/maven2/org/igniterealtime/tinder//tinder-.pom
    Not found: https://github.com/jitsi/jitsi-maven-repository/raw/master/releases/org/igniterealtime/tinder//tinder-.pom
    Not found: https://github.com/jitsi/jitsi-maven-repository/raw/master/snapshots/org/igniterealtime/tinder//tinder-.pom
  org.jitsi:jitsi-protocol-jabber:2.9-20160222.210637-7:
    Not found: /Users/jason/.ivy2/local/org.jitsi/jitsi-protocol-jabber/2.9-20160222.210637-7/ivys/ivy.xml
    Not found: https://repo1.maven.org/maven2/org/jitsi/jitsi-protocol-jabber/2.9-20160222.210637-7/jitsi-protocol-jabber-2.9-20160222.210637-7.pom
    Not found: https://github.com/jitsi/jitsi-maven-repository/raw/master/releases/org/jitsi/jitsi-protocol-jabber/2.9-20160222.210637-7/jitsi-protocol-jabber-2.9-20160222.210637-7.pom
    Not found: https://github.com/jitsi/jitsi-maven-repository/raw/master/snapshots/org/jitsi/jitsi-protocol-jabber/2.9-20160222.210637-7/jitsi-protocol-jabber-2.9-20160222.210637-7.pom
  org.jitsi:jicoco:1.0-20160202.222531-17:
    Not found: /Users/jason/.ivy2/local/org.jitsi/jicoco/1.0-20160202.222531-17/ivys/ivy.xml
    Not found: https://repo1.maven.org/maven2/org/jitsi/jicoco/1.0-20160202.222531-17/jicoco-1.0-20160202.222531-17.pom
[...]

PR #567, opened 2017-06-05
Allow pom-only snapshot dependencies to be found

rfkm:
Fix for #169 

The change was suggested by @chikei in #169.
While this looks a bit hacky and I'm not also sure whether this will cause any regression, I confirmed this works for my issue.

### DIFF
--- a/core/shared/src/main/scala/coursier/maven/MavenRepository.scala
+++ b/core/shared/src/main/scala/coursier/maven/MavenRepository.scala
@@ -12,7 +12,7 @@ object MavenRepository {
   val SnapshotTimestamp = "(.*-)?[0-9]{8}\\.[0-9]{6}-[0-9]+".r
 
   def isSnapshot(version: String): Boolean =
-    version.endsWith("SNAPSHOT") || version.matches(SnapshotTimestamp.regex)
+    version.endsWith("SNAPSHOT") || SnapshotTimestamp.findFirstIn(version).nonEmpty
 
   def toBaseVersion(version: String): String = version match {
       case SnapshotTimestamp(null) => "SNAPSHOT"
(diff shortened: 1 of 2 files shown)


alexarchambault:
That doesn't seem to work for me, I'm still getting the errors of https://github.com/coursier/coursier/issues/169 when running
```
$ coursier fetch org.jitsi:jitsi-videobridge:1.0-SNAPSHOT -r https://github.com/jitsi/jitsi-maven-repository/raw/master/releases/ -r https://github.com/jitsi/jitsi-maven-repository/raw/master/snapshots
```

You can test with this branch by running `sbt ++2.11.11 cli/pack` from its sources, then running coursier via `cli/target/pack/bin/coursier`.

rfkm:
Ah, then, my issue might be different from the original problem raised in #169. I confirmed it works with our in-house project. Let me check what is the actual problem.

rfkm:
@alexarchambault 
OK, I confirmed #169 is fixed with a9a8a02. 
Now `coursier fetch` should be able to resolve `org.jitsi:jitsi-videobridge:1.0-SNAPSHOT`, but it seems to be necessary to add ` -r https://jitpack.io`.

alexarchambault:
Thanks a lot for that PR!

I was initially under the idea that https://github.com/coursier/coursier/issues/169 was requiring more work, but it turned out I was wrong.

I'm about to rework a bit how the tests handle things from actual repositories... Currently, things are downloaded from Central (and some other repositories) during the tests. I'll probably create a separate github repo with the things from Central et al., so that these don't have to be downloaded. Then, what's used to resolve `org.jitsi:jitsi-videobridge:1.0-SNAPSHOT` could be added there too, so that non-reg tests can be added for it.

I'll merge this PR after that.

rfkm:
Sounds good! I'm looking forward to the next release :-)

alexarchambault:
This is included in https://github.com/coursier/coursier/pull/581 (merged). Thanks again!

PR #568, opened 2017-06-05
Latest developments

alexarchambault:


cache/src/main/scala/coursier/Cache.scala
```diff
@@ -1119,6 +1233,48 @@ object Cache {
...
+              override def downloadLength(url: String, length: Long, alreadyDownloaded: Long) =
+                logger.downloadLength(url, length, alreadyDownloaded)
+
+              override def downloadProgress(url: String, downloaded: Long) =
+                logger.downloadProgress(url, downloaded)
+
+              override def downloadedArtifact(url: String, success: Boolean) =
+                logger.downloadedArtifact(url, success)
+              override def checkingUpdates(url: String, currentTimeOpt: Option[Long]) =
+                checkingUpdates(url, currentTimeOpt)
```
larsrh: I think this is an infinite recursion here. Same for `checkingUpdatesResult` below.

larsrh: #638

PR #572, opened 2017-06-07
Test resolution for distinct configurations is run separately.

dotta:
The original issue was likely resolved in 8002910a433fa21e40faa072408ad2448b440383


alexarchambault:
Thanks, merging!

PR #574, opened 2017-06-09
Tree: Fix the pipe symbol to match the other UTF-8 symbols used

sschuberth:
The "branching" symbols used to print the tree are UTF-8 characters.
Make the pipe symbol be the matching UTF-8 character to close the tiny gap
between symbols visible in the tree before.


alexarchambault:
Good idea!

Looks like [this test](https://github.com/coursier/coursier/blob/master/tests/shared/src/test/scala/coursier/util/TreeTests.scala) has to be updated.

sschuberth:
Will do that!

sschuberth:
@alexarchambault Thanks for fixing up the test! I didn't realize the first column of pipe chars was actually the special chars used by `stripMargin`.

alexarchambault:
Merging, thanks!

PR #577, opened 2017-06-09
Try to re-enable osx ci (WIP)

alexarchambault:
The OSX build was probably trapped when https://github.com/travis-ci/travis-ci/issues/4681 was fixed.


alexarchambault:
There used to be an OS X job via Travis CI, but it seems not to be triggered any more. See [this discussion](https://twitter.com/alxarchambault/status/873220443947360259).

If anyone feels like trying to make that work again and / or reaching the Travis support about it, please do!

Issue #548, opened 2017-05-20
Dynamic revision redux


emanresusername:
might fit better as a continuation of [this](https://github.com/coursier/coursier/issues/351) but wasn't sure if commenting on that would go to  `/dev/null/that/issue/was/already/closed` 😅 

using version `1.0.0-RC3` i had various `#.#.+` style version numbers in [my global sbt plugins](https://github.com/emanresusername/.sbt/tree/60286f8f615d7909c5ab6ba1f175a20ecc101c6c/0.13) that caused `sbt compile` to fail with coursier (also enabled globally)

alexarchambault:
Your dot-files seem to work fine for me... Small docker-based repro [here](https://github.com/alexarchambault/coursier-issue-548) (run `./run.sh` from this repo).

Do you have more details about the error you're getting? (sbt stacktrace, etc.)

emanresusername:
yeah, so looking at it again, it looks like this is happening when it tries to get the scala 2.10 version of the plugins. not sure if that's a coursier issue, or if i need to set something up differently in my env?
```text
[error] (*:coursierResolution) coursier.ResolutionException: Encountered 3 error(s) in dependency resolution:
[error]   org.ensime:sbt-ensime;sbtVersion=0.13;scalaVersion=2.10:1.12.+:
[error]     not found:
[error]       https://repo.typesafe.com/typesafe/ivy-releases/org.ensime/sbt-ensime/scala_2.10/sbt_0.13/
[error]       https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.ensime/sbt-ensime/scala_2.10/sbt_0.13/
[error]       ~/.ivy2/local/org.ensime/sbt-ensime/scala_2.10/sbt_0.13
[error]       https://repo1.maven.org/maven2/org/ensime/sbt-ensime_2.10_0.13/maven-metadata.xml
[error]       ~/.sbt/preloaded/org.ensime/sbt-ensime
[error]       ~/.sbt/preloaded/org/ensime/sbt-ensime_2.10_0.13/maven-metadata.xml
[error]   com.softwaremill.clippy:plugin-sbt;sbtVersion=0.13;scalaVersion=2.10:0.5.+:
[error]     not found:
[error]       https://repo.typesafe.com/typesafe/ivy-releases/com.softwaremill.clippy/plugin-sbt/scala_2.10/sbt_0.13/
[...]

Issue #566, opened 2017-06-05
Failure on version wildcard in ~/.sbt/0.13/plugins/plugins.sbt


keirlawson:
My `~/.sbt/0.13/plugins/plugins.sbt` looks iike:

```
addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.0-RC3")
addSbtPlugin("org.ensime" % "sbt-ensime" % "1.+")
```

When I run SBT, this results in:

```
coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
  org.ensime:sbt-ensime;sbtVersion=0.13;scalaVersion=2.10:1.+:
    not found:
      https://repo.typesafe.com/typesafe/ivy-releases/org.ensime/sbt-ensime/scala_2.10/sbt_0.13/
      https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.ensime/sbt-ensime/scala_2.10/sbt_0.13/
      /home/keir/.ivy2/local/org.ensime/sbt-ensime/scala_2.10/sbt_0.13
      https://repo1.maven.org/maven2/org/ensime/sbt-ensime_2.10_0.13/maven-metadata.xml
      /home/keir/.sbt/preloaded/org.ensime/sbt-ensime
      /home/keir/.sbt/preloaded/org/ensime/sbt-ensime_2.10_0.13/maven-metadata.xml
[error] (*:coursierResolution) coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
[error]   org.ensime:sbt-ensime;sbtVersion=0.13;scalaVersion=2.10:1.+:
[error]     not found:
[error]       https://repo.typesafe.com/typesafe/ivy-releases/org.ensime/sbt-ensime/scala_2.10/sbt_0.13/
[error]       https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.ensime/sbt-ensime/scala_2.10/sbt_0.13/
[error]       /home/keir/.ivy2/local/org.ensime/sbt-ensime/scala_2.10/sbt_0.13
[error]       https://repo1.maven.org/maven2/org/ensime/sbt-ensime_2.10_0.13/maven-metadata.xml
[error]       /home/keir/.sbt/preloaded/org.ensime/sbt-ensime
[error]       /home/keir/.sbt/preloaded/org/ensime/sbt-ensime_2.10_0.13/maven-metadata.xml
```

[...]

PR #580, opened 2017-06-13
Add support for version ranges for sbt plugins from Maven repositories

alexarchambault:
Fixes https://github.com/coursier/coursier/issues/548, https://github.com/coursier/coursier/issues/566.


alexarchambault:
Included in https://github.com/coursier/coursier/pull/581.

PR #598, opened 2017-07-02
Allow to generate native bootstraps

alexarchambault:
cc @densh @olafurpg I stumbled upon https://github.com/densh/scala-native-cli, and had the same idea some time ago :-) I pushed it here for you to see.

It's quite hackish in its current state, as it recompiles a few modules from the scala-native sources with scala 2.11, for the cli module of coursier to be able to use them.

`extra/src/main/scala-2.11/coursier/extra/Native.scala` tries to do what the scala-native sbt plugin does (namely, what the tasks listed [here](https://github.com/scala-native/scala-native/blob/c4eac0707bebc08f4979a0d0eaed16427b734171/sbt-scala-native/src/main/scala/scala/scalanative/sbtplugin/ScalaNativePluginInternal.scala#L396-L407) do), which it seems is what your scala-native-cli does too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,8 @@ matrix:
       os: linux
       jdk: oraclejdk8
       sudo: required
+      before_install:
+      - curl https://raw.githubusercontent.com/scala-native/scala-native/v0.3.1/bin/travis_setup.sh | bash -x
       services:
         - docker
     - env: SCALA_VERSION=2.10.6 PUBLISH=1
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,6 +15,16 @@ install:
   - cmd: SET SBT_OPTS=-XX:MaxPermSize=2g -Xmx4g
   - git submodule update --init --recursive
   - ps: |
+      if (!(Test-Path 'C:\Users\appveyor\.ivy2\local\sandbox\sandbox_native0.3_2.11\0.1-SNAPSHOT')) {
+        iex 'git clone https://github.com/coursier/scala-native'
+        Set-Location -Path scala-native
+        iex 'git checkout 550bf6e37d27'
+        iex 'sbt ++2.11.8 sandbox/publishLocal'
+        iex 'git checkout f8088aef6981'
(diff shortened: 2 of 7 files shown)


alexarchambault:
To test that locally, one needs to setup scala-native like done around [here](https://github.com/coursier/coursier/blob/topic/native-bootstrap/scripts/travis.sh#L60-L72).

Then run `sbt ++2.11.11 cli/pack`, and run coursier via `cli/target/pack/bin/coursier`, like
```
$ cli/target/pack/bin/coursier bootstrap -S sandbox::sandbox_native0.3:0.1-SNAPSHOT -o native
$ ./native
Hello, World!
```

alexarchambault:
Have you though about publishing the few scala-native modules used here and by the sbt plugin (`util`, `nir`, `tools`), but for `2.11`, or even `2.12`? Or also to simply publish an extra module doing what the sbt plugin does (roughly the content of [`Native.scala` here](https://github.com/coursier/coursier/blob/topic/native-bootstrap/extra/src/main/scala-2.11/coursier/extra/Native.scala))?

olafurpg:
Very cool! I'm personally a big fan of publishing JVM artifacts to maven and let users link the application locally. Beats pushing fat jars around. Extending this model to native would be awesome. Big 👍  from me

densh:
@alexarchambault This is pretty cool. [densh/scala-native-cli](https://github.com/densh/scala-native-cli) is very much a simple proof-of-concept experiment at this point but we'd like to see a tool along those lines to be available in the near future to simplify installation of Scala Native projects.

> Have you though about publishing the few scala-native modules used here and by the sbt plugin (util, nir, tools), but for 2.11, or even 2.12?

We can definitely publish those. Do you need both 2.11 and 2.12 or would just 2.12 suffice? We might skip 2.11 altogether.

> Or also to simply publish an extra module doing what the sbt plugin does (roughly the content of Native.scala in this branch)?

Yes, definitely. Please follow https://github.com/scala-native/scala-native/issues/895 for progress on this front.

PR #610, opened 2017-07-18
Switch to sbt 1.0.0-RC2

alexarchambault:
~~Blocked by https://github.com/sbt/sbt-pgp/issues/108.~~

### DIFF
--- a/project/Settings.scala
+++ b/project/Settings.scala
@@ -178,59 +178,6 @@ object Settings {
     )
   }
 
-  lazy val scripted100M6Workaround = {
-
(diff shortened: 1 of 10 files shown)


project/Settings.scala
```diff
@@ -180,6 +180,8 @@ object Settings {
 
   lazy val scripted100M6Workaround = {
 
+    // still needed with sbt 1.0.0-RC2?
```
dwijnand: Shouldn't be

alexarchambault: Cool :-)

PR #611, opened 2017-07-18
Fix links to generate-launcher.sh

jonas:


alexarchambault:
Thanks!

The build fails because changes in the README have to be made to `doc/README.md` first. A `sbt tut` then reports those to the README.md at the root.

jonas:
Amended to also update doc/README.md

alexarchambault:
Thanks, merging!

Issue #621, opened 2017-07-25
support of SHA-256 checksums


gabor-aranyossy:
I need to use SHA-256 hashes for checksums. Adding a custom `checksumURL` to `Artifact` would allow me to do this, but eventually the validation fails with a `ChecksumFormatError` as only md5 and sha1 hashes seem to be supported [here](https://github.com/coursier/coursier/blob/master/cache/src/main/scala/coursier/Cache.scala#L875).

Would it be possible to introduce support for this somehow?

alexarchambault:
Fixed by https://github.com/coursier/coursier/issues/625.

PR #625, opened 2017-07-28
introducing sha-256 checksum support

gaboraranyossy-da:
Implementing #621. This change introduces support for `SHA-256` hashes. The default checksums to use for validation has been changed from `SHA1 -> None` to `SHA-256 -> SHA1 -> None`


tests/shared/src/test/scala/coursier/test/CentralTests.scala
```diff
@@ -728,6 +728,7 @@ abstract class CentralTests extends TestSuite {
       val ver = "1.17"
 
       def hasSha1(a: Artifact) = a.checksumUrls.contains("SHA-1")
+      def hasSha256(a: Artifact) = a.checksumUrls.contains("SHA-256")
```
alexarchambault: That line isn't necessary any more (since your last commit).

gaboraranyossy-da: Correct, I will remove it soon.

gaboraranyossy-da:
Actually, since I use the library programatically and not in our build system I could just write simply `Cache.file(artifact, checksums = Seq(Some("SHA-256"))) ` at the final step when getting the file.

I don't think using SHA-256 checksums will be standard anytime soon either, so that way I would not need to change the defaults at all.

gaboraranyossy-da:
@alexarchambault would it be possible to merge this master and create a new release?

alexarchambault:
`1.0.0-RC9` will be released once https://github.com/coursier/coursier/pull/627 is merged.

gaboraranyossy-da:
thanks!

PR #652, opened 2017-09-08
Update version on README

benmccann:

PR #653, opened 2017-09-09
Use latest version in README.md examples

igstan:


alexarchambault:
Thanks!

Like in https://github.com/coursier/coursier/pull/652, the release process [should have bumped](https://github.com/coursier/coursier/blob/3df3980c5e436c3668fe9f56423c7d5be51d52b9/project/Release.scala#L347-L349) these versions, but it looks like it didn't :/

alexarchambault:
Merged via https://github.com/coursier/coursier/commit/ce1ebc0124a8b31e8ab8462fa40bc5cc8840246f.

PR #376, opened 2016-11-03
Fix #266 by stop recursively traversing a tree

kzys:
Thanks for making the great software! I've just fixed #266. Could you take a look?

PR #671, opened 2017-10-12
Fix cli print out completeness and cyclic handling issue

wisechengyi:
## Problem
Before resolve/fetch does not output the complete graph. E.g.
```
./coursier resolve -t org.apache.avro:avro:1.7.4 org.apache.commons:commons-compress:1.4.1
 Result:
├─ org.apache.avro:avro:1.7.4
│  ├─ com.thoughtworks.paranamer:paranamer:2.3
│  ├─ org.apache.commons:commons-compress:1.4.1
│  │  └─ org.tukaani:xz:1.0
│  ├─ org.codehaus.jackson:jackson-core-asl:1.8.8
│  ├─ org.codehaus.jackson:jackson-mapper-asl:1.8.8
│  │  └─ org.codehaus.jackson:jackson-core-asl:1.8.8
│  ├─ org.slf4j:slf4j-api:1.6.4
│  └─ org.xerial.snappy:snappy-java:1.0.4.1
└─ org.apache.commons:commons-compress:1.4.1
          // why is there nothing over here?
```

## Solution
Do not avoid jars have been printed out before. E.g.
```
]$ cli/target/pack/bin/coursier resolve  -t org.apache.avro:avro:1.7.4   org.apache.commons:commons-compress:1.4.1 
  Result:
├─ org.apache.avro:avro:1.7.4
│  ├─ com.thoughtworks.paranamer:paranamer:2.3
│  ├─ org.apache.commons:commons-compress:1.4.1
│  │  └─ org.tukaani:xz:1.0
│  ├─ org.codehaus.jackson:jackson-core-asl:1.8.8
│  ├─ org.codehaus.jackson:jackson-mapper-asl:1.8.8
│  │  └─ org.codehaus.jackson:jackson-core-asl:1.8.8
│  ├─ org.slf4j:slf4j-api:1.6.4
│  └─ org.xerial.snappy:snappy-java:1.0.4.1
└─ org.apache.commons:commons-compress:1.4.1
   └─ org.tukaani:xz:1.0
```

### DIFF
--- a/core/shared/src/main/scala/coursier/util/Tree.scala
+++ b/core/shared/src/main/scala/coursier/util/Tree.scala
@@ -15,11 +15,9 @@ object Tree {
       * @param acc       accumulation method on a string
       */
     def recursivePrint(elems: Seq[T], ancestors: Set[T], prefix: String, acc: String => Unit): Unit = {
-      for (elem <- elems) {
-        if (ancestors.contains(elem)) {
-          return
-        }
-        val isLast = elems.indexOf(elem) == elems.length - 1
+      val unseenElems: Seq[T] = elems.filterNot(ancestors.contains)
+      for (elem <- unseenElems) {
+        val isLast = unseenElems.indexOf(elem) == unseenElems.length - 1
         val tee = if (isLast) "└─ " else "├─ "
         acc(prefix + tee + show(elem))
 
--- a/core/shared/src/main/scala/coursier/util/Tree.scala
+++ b/core/shared/src/main/scala/coursier/util/Tree.scala
@@ -6,13 +6,20 @@ object Tree {
 
   def apply[T](roots: IndexedSeq[T])(children: T => Seq[T], show: T => String): String = {
 
-
-    def recursivePrint(roots: Seq[T], ancestors: Set[T], prefix: String, acc: String => Unit): Unit = {
-      for (root <- roots) {
+    /**
+      * Recursively go down the resolution for the elems to construct the tree for print out.
+      *
+      * @param elems     Seq of Elems that have been resolved
+      * @param ancestors a set of Elems to keep track for cycle detection
+      * @param prefix    prefix for the print out
+      * @param acc       accumulation method on a string
+      */
+    def recursivePrint(elems: Seq[T], ancestors: Set[T], prefix: String, acc: String => Unit): Unit = {
+      for (root <- elems) {
         if (ancestors.contains(root)) {
           return
         }
-        val isLast = roots.indexOf(root) == roots.length - 1
+        val isLast = elems.indexOf(root) == elems.length - 1
         val tee = if (isLast) "└─ " else "├─ "
         acc(prefix + tee + show(root))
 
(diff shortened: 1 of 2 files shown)


alexarchambault:
Thanks! I'm now realizing that https://github.com/coursier/coursier/pull/376, which fixed https://github.com/coursier/coursier/issues/266, added no non regression test for it... And I'm not sure the changes here are fine with it.

I think the code in question should simply remember what are the parents of the node about to be printed, and not print it if it has itself has a parent (which the current code does, but in a buggy way).

wisechengyi:
haha yeah I broke it again :D 
```
$ coursier resolve -t \
    -r http://cogcomp.cs.illinois.edu/m2repo \
    edu.illinois.cs.cogcomp:illinois-nlp-pipeline:0.1.21
```

wisechengyi:
To clarify what you mean by 
>  should simply remember what are the parents of the node about to be printed, and not print it if it has itself has a parent

Basically it is to see if any direct ancestor has appeared before, if so, do not recurse.

if 
```
a -> b -> c -> a
            -> e -> f
```
```
resolve a, e
```
should give:
```
a 
-b
--c 
// not printing 'a' again
---e
----f
e
-f
```

@alexarchambault could you kindly confirm?

alexarchambault:
@wisechengyi Yep, that's what I meant!

wisechengyi:
Reverting to the style pre https://github.com/coursier/coursier/pull/376/files, and add cycle detection.

wisechengyi:
hey @alexarchambault, do you know if there is a way to trigger a single test from sbt, like `coursier.util.TreeTests`? still scala newbie here

alexarchambault:
@wisechengyi Something like `testsJVM/testOnly coursier.util.TreeTests` at the sbt prompt, should do, like
```
$ sbt
…
> testsJVM/testOnly coursier.util.TreeTests
…
> ~testsJVM/testOnly coursier.util.TreeTests
…
```

(second one watches for changes)

alexarchambault:
@wisechengyi Tell me as soon as you want me to review.

wisechengyi:
@alexarchambault should be good for review now. thank you!

core/shared/src/main/scala/coursier/util/Tree.scala
```diff
@@ -16,8 +16,9 @@ object Tree {
       */
     def recursivePrint(elems: Seq[T], ancestors: Set[T], prefix: String, acc: String => Unit): Unit = {
       val unseenElems: Seq[T] = elems.filterNot(ancestors.contains)
-      for (elem <- unseenElems) {
-        val isLast = unseenElems.indexOf(elem) == unseenElems.length - 1
```
alexarchambault: I think the call to `indexOf` makes things quadratic here (calling `indexOf` for each element should be quadratic wrt to `unseenElems.length`). I pushed a quick fix.

wisechengyi: ok good catch. thanks!

alexarchambault:
Merging, thanks!

PR #672, opened 2017-10-12
Update homebrew installation command

Grogs:
The command to install via homebrew didn't work for me. Presumably this changed in a recent version of homebrew.

This is the error I received:
```
$ brew install coursier/formulas/coursier

Error: coursier/formulas/coursier is a head-only formula
Install with `brew install --HEAD coursier/formulas/coursier`
```

So this change just incorporates the fix it suggests, which works for me.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ $ curl -L -o coursier https://git.io/vgvpD && chmod +x coursier && ./coursier --
 
 Alternatively on OS X, install it via homebrew,
 ```
-$ brew install coursier/formulas/coursier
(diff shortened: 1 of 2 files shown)


alexarchambault:
@Grogs Thanks! That was required since https://github.com/coursier/homebrew-formulas/commit/848b7ce17f531ae34d3dba5921bfdee22d7b39d4, yes.

I pushed an extra commit reporting this change to `doc/README.md` (out of which the readme at the root is generated via a `sbt tut`), and also fixed the other brew command in the README.

PR #676, opened 2017-10-23
Introduce support for OS-specific standard cache directories

soc:
This cleans up the home directories of users and helps making sure that
Coursier's cache is not accidentically backed up by applications or OS
functionality.

The precedence is as follows:

- existing $COURSIER_CACHE environment variable
- existing coursier.cache Java property
- existing ~/.coursier/
- existing operating system specific standards:
  - Linux:   $XDG_CACHE_HOME/coursier, with fallback to ~/.cache/coursier
  - Windows: {SpecialFolder.LocalApplicationData}/cache/Coursier
  - macOS:   $HOME/Library/Caches/Coursier

### DIFF
--- a/paths/src/main/java/coursier/CoursierPaths.java
+++ b/paths/src/main/java/coursier/CoursierPaths.java
@@ -27,7 +27,7 @@ public final class CoursierPaths {
         if (path == null) {
         	File coursierDotFile = new File(System.getProperty("user.home") + "/.coursier");
         	if (coursierDotFile.isDirectory())
-                path = System.getProperty("user.home") + "/.coursier/cache/v1/";
+        		path = System.getProperty("user.home") + "/.coursier/cache/";
         }
 
         if (path == null) {
@@ -35,6 +35,11 @@ public final class CoursierPaths {
         	 path = coursierDirectories.projectCacheDir;
         }
 
-        return new File(path).getAbsoluteFile();
+        File coursierCacheDirectory = new File(path).getAbsoluteFile();
+
+        if (coursierCacheDirectory.getName().equals("v1"))
+        	coursierCacheDirectory = coursierCacheDirectory.getParentFile();
+        
+        return coursierCacheDirectory.getAbsoluteFile();
 	}
 }
--- a/paths/src/main/java/coursier/CoursierPaths.java
+++ b/paths/src/main/java/coursier/CoursierPaths.java
@@ -7,39 +7,66 @@ import io.github.soc.directories.ProjectDirectories;
 /**
  * Computes Coursier's directories according to the standard
  * defined by operating system Coursier is running on.
- * 
+ *
  * @implNote If more paths e. g. for configuration or application data is required,
- *           use {@link #coursierDirectories} and do not roll your own logic.
+ * use {@link #coursierDirectories} and do not roll your own logic.
  */
 public final class CoursierPaths {
-	private CoursierPaths() {
-		throw new Error();
-	}
-	
-	private static ProjectDirectories coursierDirectories;
-	
-	public static File cacheDirectory() {
+    private CoursierPaths() {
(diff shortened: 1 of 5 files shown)


alexarchambault:
@soc Shouldn't `dataDir` used here, instead of `cacheDir`? XDG base directory spec says that the cache dir is for "non-essential data files", where `dataDir` is for "user specific data files". I feel like the latter would be a better match (the "cache" of coursier consists of data files, but I wouldn't say they're "non-essential").

alexarchambault:
About the precedence, I think the last two points should be permuted, like
- if OS specific dir for coursier exists, use it,
- else if `~/.coursier` exists, use it,
- else, if none exists, fallback to OS specific dir.

soc:
@alexarchambault 

> Shouldn't dataDir used here, instead of cacheDir?

No, not in this case. There is no platform-independent way to avoid moving GBs of cached files across the network (only Windows supports the distinction with roaming vs. local application data). Cache is the right folder, because it perfectly fits the distinction of "the only thing users see after deleting the cache is that some things take more time".

On the other hand, packages that are published locally should go into the dataDir, as they are likely not recoverable from the online repositories.

This way, users have the huge benefit of being able to delete their caches without losing their locally published artifacts, and have their locally published artifacts synced across machines without dragging GBs of cached artifacts along with them.

> About the precedence

Makes sense, I can change that.

paths/src/main/java/coursier/CoursierPaths.java
```diff
@@ -27,19 +27,14 @@ public static File cacheDirectory() {
...
         	 coursierDirectories = ProjectDirectories.fromProjectName("Coursier");
         	 path = coursierDirectories.projectCacheDir;
         }
 
-        File coursierCacheDirectory = new File(path).getAbsoluteFile();
-
-        if (coursierCacheDirectory.getName().equals("v1"))
-        	coursierCacheDirectory = coursierCacheDirectory.getParentFile();
-        
-        return coursierCacheDirectory.getAbsoluteFile();
```
soc: I made these changes to separate figuring out the various directories in `CoursierPaths` from the way they are used `CachePath`.

This way dealing with "v1" makes it easy for Coursier to potentially upgrade to a "v2" layout later, without breaking everyone's environment settings pointing to the "v1" dir.

The idea is that you only need to adapt `defaultCacheDirectory` to use "v2" and be done without having to write migration code and documenting how users need to change their environment variables.

alexarchambault:
Ok, let's go for `cacheDir` then. (That's right that it makes no sense to somehow sync / save the cache dir of coursier, so it's better there.)

And, yep, I'm going to revert my commit about `v1`. It's better to put the cache in a sub dir of the `cacheDir` returned by directories (to separate it from other things that could be kept in `cacheDir`), and I can't think of a better name (and `v1` is shorter than e.g. `downloads`).

alexarchambault:
@soc Merging, thanks for having pushed for that!

soc:
@alexarchambault It has been a pleasure!

Did you have any chance yet to verify that the paths chosen by `directories` work fine on Winows? (Mac, too, but I have less concerns there, because the solution to derive the directories is substantially less complex than Window's "let's tell the JVM to tell the OS to run a new process, run PowerShell in that process and then call functions from .NET").

alexarchambault:
@soc The Windows CI didn't complain, I'm guessing things work fine there.

alexarchambault:
@soc Where cache files land on Windows 10 (in a [free VM](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/)):
coursier-cache-win10

soc:
@alexarchambault This agrees with http://doc.qt.io/qt-5/qstandardpaths.html, but I would love to know how they came up with this path in the first place. Can't find anything in the Microsoft documentation.

PR #679, opened 2017-10-27
Tweaking

alexarchambault:


alexarchambault:
@soc FYI, the second commit here changes again the handling of the `v1` component in cache path. The goal here is not to change the current behavior, when e.g. `COURSIER_CACHE` is set in the env. Before https://github.com/coursier/coursier/pull/676, things landed in `$COURSIER_CACHE/https/…`. Right after it, in `$COURSIER_CACHE/v1/https/…`. The latter is better in the long term, but I'd like not to change that right now.

PR #690, opened 2017-11-03
Implemented clean task + added integration test. Fixes #685

### LINKED ISSUE #685: clean doesn't clean the in-memory cache in sbt-coursier
sbt-coursier uses a few [in-memory caches](https://github.com/coursier/coursier/blob/eac0c2ce4528d12cdbd78ca5685a73b3df2e590a/sbt-coursier/src/main/scala/coursier/Tasks.scala#L417-L420) where it keeps a few results so that the corresponding tasks don't incur IO when they're re-run. When running `clean`, these caches should be wiped (as are the on-disk caches of sbt I think), which is not the case now.

olivierdeckers:


alexarchambault:
Green! Merging, thanks again!

Issue #665, opened 2017-09-23
Coursier not resilient to changing local ivy cache directory


Mocuto:
For resolving local cached dependencies, Coursier assumes they're in the $user.home/.ivy directory, even though that is absolutely not guaranteed.

See the following lines in Cache.scala:
```

  private lazy val ivy2HomeUri = {

    // a bit touchy on Windows... - don't try to manually write down the URI with s"file://..."

    val str = new File(sys.props("user.home") + "/.ivy2/").toURI.toString

    if (str.endsWith("/"))

      str

    else

      str + "/"

  }
```
It's very likely that a user may change their ivy home location via the SBT options:

```

-Dsbt.ivy.home=D:/mocut/.ivy2/

-Divy.home=D:/mocut/.ivy2/

```

Thus, Coursier should try `sys.props("ivy.home")` or `sys.props("sbt.ivy.home")` before deferring to `sys.props("user.home")`

alexarchambault:
It shouldn't be an issue when using coursier from sbt, as the ivy2 local repository comes from sbt itself, not from `coursier.Cache.ivy2Local`. The right Java properties should already be handled fine thanks to that.

It can only be a problem when using coursier via its API (and using directly `coursier.Cache.ivy2Local`), or from the CLI of coursier.

I'd be happy to merge a PR taking into account the `ivy.home` property for `coursier.Cache.ivy2*`.

alexarchambault:
Fixed by https://github.com/coursier/coursier/pull/705.

PR #705, opened 2017-12-03
Fix for: Coursier not resilient to changing local ivy cache directory #665 

Mocuto:
Made to address issue: https://github.com/coursier/coursier/issues/665
New priority for ivy home location is:
1. ivy.home property
2. sbt.ivy.home property
3. user.home property

One primary question with this fix: Will this address instances where coursier is used from CLI/API/Jar file?

### DIFF
--- a/cache/src/main/scala/coursier/Cache.scala
+++ b/cache/src/main/scala/coursier/Cache.scala
@@ -1075,7 +1075,7 @@ object Cache {
   private lazy val ivy2HomeUri = {
 
     val path =
-      sys.props("coursier.ivy.home")
+      sys.props.get("coursier.ivy.home")
         .orElse(sys.props.get("ivy.home"))
         .getOrElse(sys.props("user.home") + "/.ivy2/")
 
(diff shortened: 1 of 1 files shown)


alexarchambault:
@Mocuto Thanks! I'm not sure it works as expected as is. If `ivy.home` is `~/.ivy2`, the local repo should be under `~/.ivy2/local`. It seems to be `~/.ivy2/.ivy/local` with the current changes.

Also, this setting doesn't apply from the sbt plugin (sbt defines the repo itself). So I'm not sure `sbt.ivy.home` should be taken into account here. Possibly a `coursier.ivy.home` instead, but not `sbt.ivy.home` I think.

Mocuto:
Gotcha, so it is expected that the `val ivy2HomeUri` should point to the actual `.ivy2` directory, rather than the directory that CONTAINS the `.ivy2` directory?

Is that interpretation correct?

alexarchambault:
@Mocuto Yep, that's what I meant. I pushed some tweaking over your commit, I'll merge right after once CI is green.

Issue #695, opened 2017-11-22
[propose/discuss] Add Pants build tool to coursier


wisechengyi:
A few reasons why SBT isn't ideal: http://www.lihaoyi.com/post/SowhatswrongwithSBT.html

Personally speaking, I feel the build time is getting quite slow, so Pants is definitely going to help with more modularization and build cache. Scala is first class citizen in Pants https://www.pantsbuild.org/

Also an option to use something like bazel, but I am not familiar with that per se.

Curious what is your take on it @alexarchambault

wisechengyi:
I put up a sample at https://github.com/coursier/coursier/pull/697, since proguarded jar behaviors really werid and I couldn't find a way to build non-proguarded one. Sorry a bit messy with other changes, but the part matters is just adding some BUILD files and Pants bootstrap script.
```
$ git diff 659 --stat
 3rdparty/jvm/BUILD                                   |  96 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 cache/src/main/scala/BUILD                           |   9 +++++++++
 cli/src/main/scala-2.11/BUILD                        |  19 +++++++++++++++++++
 cli/src/main/scala-2.11/coursier/cli/Bootstrap.scala |  33 ++-------------------------------
 cli/src/main/scala-2.11/coursier/cli/Coursier.scala  |   2 +-
 core/BUILD                                           |  13 +++++++++++++
 extra/src/main/scala/coursier/extra/BUILD            |   7 +++++++
 pants                                                | 102 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 pants.ini                                            |  13 +++++++++++++
 paths/src/main/java/BUILD                            |   7 +++++++
 10 files changed, 269 insertions(+), 32 deletions(-)
```

You can do something like below:
```
# Run the binary target
$ ./pants run cli/src/main/scala-2.11:coursier-cli -- fetch --help

# Make a distributable bundle at dist/coursier-cli.jar
$ ./pants binary cli/src/main/scala-2.11:coursier-cli
```

The initial bootstrap may take a few seconds, but later runs are much faster than sbt.

godenji:
@wisechengyi would be interesting to compare to latest sbt (1.0.4 released today), which fixes a long standing issue with slow hashing leading to laggy incremental builds .

[...]

PR #709, opened 2017-12-08
Add Pants build tool

wisechengyi:
Per discussion https://github.com/coursier/coursier/issues/695

The only code change is at https://github.com/coursier/coursier/pull/709/files#diff-7ca3c74b5087813a533febc6b4a5a7a1R9 as coursier.properties is probably auto generated by sbt. Pants does support build properties but not in this fashion.

The commits are from repurposing other branches, but the actual change is purely declarative.


alexarchambault:
@wisechengyi You shoudn't need to add jackson as a dependency in this PR, do you?

That said, I can merge anyway if you'd like…

wisechengyi:
Ah thanks. Forgot to remove it since I was working backwards. Now removed.

alexarchambault:
Ok, so merging!

(A bit a pity that pants doesn't call its build files `PANTS` though…, like buck calls his, `BUCK`.)

Thanks @wisechengyi !

wisechengyi:
haha. actually there has been conversation about converging pants, bazel, and buck to use a common build language called skylark https://github.com/google/skylark. but no nothing immediate yet.

Issue #684, opened 2017-11-01
Add missing bits in sbt update reports


alexarchambault:
sbt-coursier overrides the default library management of sbt by somehow hijacking the `update` task. This task result is an "update report", which describes the transitive dependencies of the user added dependencies. The update reports from sbt-coursier are generated by [this method](https://github.com/coursier/coursier/blob/eac0c2ce4528d12cdbd78ca5685a73b3df2e590a/sbt-coursier/src/main/scala/coursier/ToSbt.scala#L199).

sbt-coursier currently generates reports that lack infos about exclusions or transitiveness of dependencies.

Missing exclusions result in the following error in particular:
```
$ cat project/build.properties
sbt.version=1.0.3
$ cat project/project/plugins.sbt
addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.0-RC13")
$ cat project/plugins.sbt
addSbtCoursier
addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.5")
$ sbt updateSbtClassifiers
…
[info] Updating coursier-sbt-assembly (sbt classifiers)
[info] Resolved coursier-sbt-assembly dependencies
[error] coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
[error]     org.jboss.interceptor:jboss-interceptor-api:1.1:
[error]         not found:
…
[error]         wrong checksum:
[error]             …/.coursier/cache/v1/https/repo1.maven.org/maven2/org/jboss/interceptor/jboss-interceptor-api/1.1/jboss-interceptor-api-1.1.pom
…
[error] (*:coursierSbtClassifiersResolution) coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
…
```

The dependency with a wrong checksum (`org.jboss.interceptor:jboss-interceptor-api:1.1`) shouldn't be pulled here. It's an excluded dependency of sbt-assembly. It gets pulled anyway because the update report of the ~meta project lacks the exclusions. `updateSbtClassifiers` resolves things again via the meta project update report, thus misses the exclusions.

PR #712, opened 2017-12-11
Write exclusions and transitiveness in sbt update reports, add sbt-shared module

alexarchambault:
Fixes https://github.com/coursier/coursier/issues/684.


leonardehrenfried:
From a quick glance this appears to put the FromSbt and ToSbt classes into a separate module, so my librarymanagement PR can use them.

Am I seeing this correctly?

alexarchambault:
@leonardehrenfried Yep, these are in the `sbt-shared` module now. To depend on it, something like
```scala
addSbtPlugin("io.get-coursier" % "sbt-shared" % "1.0.0-RC14")
```
should do.

dwijnand:
I think we should add anything that coursier needs and that is currently missing in https://github.com/dwijnand/sbt-compat, rather than have competing skim library..

@leonardehrenfried Do you think you could take that on?

alexarchambault:
@dwijnand That would be cool, indeed.

leonardehrenfried:
Maybe I don't get it, but as far as I see sbt-compat is for backporting APIs from 1.0 to 0.13.

What the new module is providing is conversions from coursier to sbt and vice versa.

dwijnand:
You're right, ToSbt and FromSbt are for that. But there's also a bunch of code (in scala-2.10 and scala-2.12) to handle sbt 1 vs sbt 0.13. That code should migrate to sbt-compat (imho).

leonardehrenfried:
Ah, I see. Yes I can make a PR for that.

leonardehrenfried:
@dwijnand I've made a PR at https://github.com/dwijnand/sbt-compat/pull/5. Would you be able to take a look?

dwijnand:
Sorry @leonardehrenfried I've been a bit busy with end of the year wrap ups. But I have that PR on the top of my list for the beginning of the year. :-/

PR #715, opened 2017-12-15
Update version numbers in CI scripts after 1.0.0 release

alexarchambault:


alexarchambault:
@wisechengyi Yep, that can be done later. That file could actually be rewritten in ammonite script, to make it easier to maintain…

PR #721, opened 2017-12-28
Remove sbt-launcher project

alexarchambault:
Now lives in its own repo at https://github.com/coursier/sbt-launcher


dwijnand:
Just out of curiosity, why the repo split?

alexarchambault:
I moved away the components that didn't really need to be in this repo (see also the echo and http-server projects under the coursier organization)…

Mainly because:
- they're likely to have different release cycles than the core of coursier (they're likely to change less than the core), and one usually doesn't develop on both them and the core now,
- they could have better tests without cluttering the CI of coursier itself (http-server in particular should have more tests),
- they might not make sense anymore in the future! (sbt-launcher might be superseded https://github.com/sbt/librarymanagement/pull/190)

alexarchambault:
I hesitated to open an issue first, to discuss it… I might do that in the future.

PR #724, opened 2017-12-28
Performance improvements for cli json report and minor refactoring

wisechengyi:
Major change:
* Identified hotspot in `JsonElem.hashCode` because it hashes children recursively. Hence simplified hashing scheme, reducing CPU time on `JsonReport.apply()` from 64.7 sec to 2.5 sec

Before:
screen shot 2017-12-28 at 1 26 20 pm

After:
screen shot 2017-12-28 at 1 26 27 pm

Other minor changes:
* Make `JsonRequirement` use immutable classes
* Simplify `JsonRequirement` composition

### DIFF
--- a/cli/src/main/scala-2.11/coursier/cli/Helper.scala
+++ b/cli/src/main/scala-2.11/coursier/cli/Helper.scala
@@ -716,12 +716,12 @@ class Helper(
           a.isOptional && notFound
       }
 
-    val artifactToFile: collection.mutable.Map[String, File] = collection.mutable.Map()
-    val files0 = results.collect {
+    val artifactToFile = results.collect {
       case (artifact: Artifact, \/-(f)) =>
-        artifactToFile.put(artifact.url, f)
-        f
-    }
+        (artifact.url, f)
+    }.toMap
+
+    val files0 = artifactToFile.values.toSeq
 
     logger.foreach(_.stop())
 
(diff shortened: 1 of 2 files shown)


wisechengyi:
That should be fine. GitHub has the feature to merge the PR squashed, so we don't have to do it manually

wisechengyi:
https://cloud.githubusercontent.com/assets/143418/14119749/7756ff60-f5a4-11e5-8597-12dcbb363746.png

alexarchambault:
Yep, I know that feature, I meant you could have squashed it in two commits, then merge as is.

wisechengyi:
Copying from gitter:

Hi @alexarchambault, by squashing into 2 commits, are you mostly concerned about the commit messages showing up in master?

Typically each commit message does not really master, and with Github's sqaush and merge, the description of the PR can go into the final commit message.

Maybe we can be more clear about how the PR should go. For example, https://github.com/pantsbuild/pants/blob/a439c87a1cef96f31bda6eb557716d5df3fd2a09/src/docs/howto_contribute.md#committing-a-change

Additional benefits if we keep the commits in the PR:
1. Github reviews are associated with the commit sha, so after force push after rebase/squash, they may be gone.
2. It allows for inspecting a range of shas on github to see what changes are made after the review for example.
3. Book keeping.
4. Personally I am bad at squashing commits manually :D

alexarchambault:
The way to merge PRs could be clarified, yes… I'm more for loose guidelines rather than hard ones.

That could be either squash-and-merge if the PR basically adds or changes one thing, or merge with merge commit if the succesive commits of the PR somehow make sense without the next ones (that is, in a PR with commits `a -> b -> c`, just adding `a` to master would have been fine, as would have adding `a -> b`).

I often tend to add a bunch of small things in one PR, so I usually follow the latter case.

wisechengyi:
Sure sounds good. Will squash then push in a bit. Thanks

wisechengyi:
squashed. thanks

wisechengyi:
Ah ok. But i don't think i am able to merge anymore. Could you double check?

alexarchambault:
Oh, possible github misconfig on my side. Is it better now?

wisechengyi:
thank you!

alexarchambault:
@wisechengyi Will you need a release that would include your recent PRs?

wisechengyi:
This change does not require a release. We are currently consuming directly off master. Thanks for asking :)

PR #729, opened 2018-01-09
Enable reflective class in the hackaround of tweakedScriptedTask

dwijnand:


alexarchambault:
@dwijnand Rebasing this one on the current master should make the failing test succeed.

alexarchambault:
Thanks!

PR #730, opened 2018-01-09
Update, extract & override scala versions for plugins

dwijnand:

PR #731, opened 2018-01-09
Replace most of SbtCompatibility with sbt-compat

dwijnand:
The sbt-compat version is the latest release from master. I'll cut a stable version if this goes well.


alexarchambault:
Looks cool!

alexarchambault:
If ever we can't get https://github.com/coursier/coursier/pull/730 to work, the current CI failures could be addressed by ensuring the dependency towards sbt-compat is only added if in scala 2.10 or 2.12.

dwijnand:
I don't things will compile without it, but we can try and see.

alexarchambault:
@dwijnand It should work thanks to https://github.com/coursier/coursier/blob/39c7708297a38d673072835e840a537b2faf2429/project/Settings.scala#L159-L166 😈

dwijnand:
This went green for both checks. So I published sbt-compat 1.2.0 and switched to that.

alexarchambault:
Merging, thanks @dwijnand!

Issue #717, opened 2017-12-21
Fine grain classifier specification on cli


wisechengyi:
Currently `--classifier` is applied to all modules requested on the cli. Ideally we should be able specify below in the same run.
```
fetch org:name:version \ # default classfieir
org2:name2:version2:classifierX
```
The same may also apply to packaging.

Quote from @baroquebobcat:

It looks like you can provide coursier with classifiers via the extra attribute syntax. I tried it locally with org.apache.kafka:kafka-streams:1.0.0;classifier=test, and it was accepted, but ignored. Maybe that's something we could file an issue about. I think it would clean this up a lot, and also allow projects with classified dependencies to function.

https://github.com/coursier/coursier/blob/8f0fef6039f1eb6321997446ab565020f9b8af97/core/shared/src/main/scala/coursier/util/Parse.scala#L24

wisechengyi:
@alexarchambault I plan to work on this soon. Let me know if you have any preference in term of 
`org2:name2:version2:<classifier>` vs `org2:name2:version2;classifier=<classifier>`, or some other format.

I think this should be the last blocking issue for us to fully adopt coursier internally.

alexarchambault:
Let's go for the latter, the first one conflicts with [the one for configurations](https://github.com/coursier/coursier/blob/70fa5f13cb36b2e30bfdddbc76566cbf878bbd79/core/shared/src/main/scala/coursier/util/Parse.scala#L130).

wisechengyi:
Sounds good.

wisechengyi:
@axarchambault how do feel about using something other than `;` because it is the separator for bash, so the behavior may be confusing for user who misses quoting it.

How about `#` or `$`? E.g. `org2:name2:version2#classifier=<classifier>`, `org2:name2:version2$classifier=<classifier>`

alexarchambault:
Maybe we could use `:` again… Like `org:name:version:a=b:c=d` To avoid conflicts with `org:name:version:<classifier>`, we can check for `=` in the forth part.

wisechengyi:
Using `:` is fine. User should be still able to specify config if they want along with classifier, and I do not intend to change the match cases here:
https://github.com/coursier/coursier/blob/74b2f33dd67d0a0ef8e40236fc9ebedeb09113ef/core/shared/src/main/scala/coursier/util/Parse.scala#L164-L187

so both should be valid:
org:name:version:a=b:c=d
org:name:version:config:a=b:c=d

[...]

PR #735, opened 2018-01-12
Specify classifier on module level on CLI

wisechengyi:
Currently `--classifier` is applied to all modules requested on CLI, so this change add the capability to specify classifier for each module.
```
fetch -t \
org.apache.commons:commons-compress:1.5,classifier=tests \
org.apache.commons:commons-compress:1.5
```

Fix for #717

### DIFF
--- a/cli/src/main/scala-2.11/coursier/cli/Helper.scala
+++ b/cli/src/main/scala-2.11/coursier/cli/Helper.scala
@@ -10,7 +10,6 @@ import coursier.cli.scaladex.Scaladex
 import coursier.cli.util.{JsonElem, JsonPrintRequirement, JsonReport}
 import coursier.extra.Typelevel
 import coursier.ivy.IvyRepository
-import coursier.util.Parse.ParsedModule
 import coursier.util.{Parse, Print}
 
 import scala.annotation.tailrec
@@ -84,9 +83,9 @@ class Helper(
   isolated: IsolatedLoaderOptions = IsolatedLoaderOptions(),
   warnBaseLoaderNotFound: Boolean = true
 ) {
-  import common._
   import Helper.errPrintln
   import Util._
+  import common._
 
   val ttl0 =
     if (ttl.isEmpty)
@@ -168,7 +167,7 @@ class Helper(
   val (scaladexRawDependencies, otherRawDependencies) =
     rawDependencies.partition(s => s.contains("/") || !s.contains(":"))
 
-  val scaladexModuleVersionConfigs: List[Dependency] =
+  val scaladexDeps: List[Dependency] =
     if (scaladexRawDependencies.isEmpty)
       Nil
     else {
@@ -228,11 +227,8 @@ class Helper(
       res
         .collect { case \/-(l) => l }
         .flatten
-        .map { case (mod, ver) => Dependency(mod, ver, attributes=Attributes("", "")) }
+        .map { case (mod, ver) => Dependency(mod, ver) }
     }
-
-
-
 
   val (forceVersionErrors, forceVersions0) = Parse.moduleVersions(forceVersion, scalaVersion)
 
@@ -323,14 +319,10 @@ class Helper(
       }).groupBy(_._1).mapValues(_.map(_._2).toSet).toMap
     }
 
-  val (modVerCfgErrors: Seq[String], moduleVersionConfigs: Seq[Dependency]) =
+  val (modVerCfgErrors: Seq[String], normalDeps: Seq[Dependency]) =
     Parse.moduleVersionConfigs(otherRawDependencies, globalExcludes, localExcludeMap, scalaVersion)
-  val (intransitiveModVerCfgErrors: Seq[String], intransitiveModuleVersionConfigs: Seq[Dependency]) =
+  val (intransitiveModVerCfgErrors: Seq[String], intransitiveDeps: Seq[Dependency]) =
     Parse.moduleVersionConfigs(intransitive, globalExcludes, localExcludeMap, scalaVersion)
-
-  val allModuleVersionConfigs: Seq[Dependency] =
-  // FIXME Order of the dependencies is not respected here (scaladex ones go first)
-    scaladexModuleVersionConfigs ++ moduleVersionConfigs
 
   prematureExitIf(modVerCfgErrors.nonEmpty) {
     s"Cannot parse dependencies:\n" + modVerCfgErrors.map("  "+_).mkString("\n")
@@ -341,33 +333,11 @@ class Helper(
       intransitiveModVerCfgErrors.map("  "+_).mkString("\n")
   }
 
+  val transitiveDeps: Seq[Dependency] =
+  // FIXME Order of the dependencies is not respected here (scaladex ones go first)
+    scaladexDeps ++ normalDeps
 
-  private def createDependency(parsedModule: ParsedModule, transitive: Boolean) = {
-    val attributes = parsedModule.attrs.get("classifier") match {
-      case Some(c) => Attributes("", c)
-      case None => Attributes("", "")
-    }
-
-    Dependency(
-      parsedModule.module,
-      parsedModule.version,
-      attributes = attributes,
-      configuration = parsedModule.config.getOrElse(defaultConfiguration),
-      exclusions = localExcludeMap.getOrElse(parsedModule.module.orgName, Set()) | globalExcludes
-    )
-  }
-
-//  val baseDependencies: Seq[Dependency] = allModuleVersionConfigs.map {
-//    case parsedModule =>
-//      createDependency(parsedModule, transitive = true)
-//  }
-//
-//  val intransitiveDependencies: Seq[Dependency] = intransitiveModuleVersionConfigs.map {
-//    case parsedModule =>
-//      createDependency(parsedModule, transitive = false)
-//  }
-
-  val dependencies: Seq[Dependency] = allModuleVersionConfigs ++ intransitiveModuleVersionConfigs
+  val allDependencies: Seq[Dependency] = transitiveDeps ++ intransitiveDeps
 
   val checksums = {
     val splitChecksumArgs = checksum.flatMap(_.split(',')).filter(_.nonEmpty)
@@ -383,7 +353,7 @@ class Helper(
   val userEnabledProfiles = profile.toSet
 
   val startRes = Resolution(
-    dependencies.toSet,
+    allDependencies.toSet,
     forceVersions = forceVersions,
     filter = Some(dep => keepOptional || !dep.optional),
     userActivations =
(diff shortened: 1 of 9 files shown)


alexarchambault:
@wisechengyi You can add tests for what you added in `core/shared/src/main/scala/coursier/util/Parse.scala` in [ParseTests](https://github.com/coursier/coursier/blob/70fa5f13cb36b2e30bfdddbc76566cbf878bbd79/tests/shared/src/test/scala/coursier/test/ParseTests.scala).

Also, I see you changed `Resolution` a bit… This needs a test [somewhere](https://github.com/coursier/coursier/tree/70fa5f13cb36b2e30bfdddbc76566cbf878bbd79/tests/shared/src/test/scala/coursier/test).

wisechengyi:
@alexarchambault the part changed in `Resolution` actually does not concern anything but cli, can I move that component to cli instead?

```
$ find . | grep \.scala$ | xargs grep moduleVersionConfig -irn
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:171:  val scaladexModuleVersionConfigs: List[ParsedModule] =
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:235:  val (modVerCfgErrors: Seq[String], moduleVersionConfigs: Seq[ParsedModule]) =
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:236:    Parse.moduleVersionConfigs(otherRawDependencies, scalaVersion)
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:237:  val (intransitiveModVerCfgErrors: Seq[String], intransitiveModuleVersionConfigs: Seq[ParsedModule]) =
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:238:    Parse.moduleVersionConfigs(intransitive, scalaVersion)
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:240:  val allModuleVersionConfigs: Seq[ParsedModule] =
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:242:    scaladexModuleVersionConfigs ++ moduleVersionConfigs
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:359:  val baseDependencies: Seq[Dependency] = allModuleVersionConfigs.map {
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:364:  val intransitiveDependencies: Seq[Dependency] = intransitiveModuleVersionConfigs.map {
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:885:          parsedModule: ParsedModule <- allModuleVersionConfigs.headOption
./cli/src/main/scala-2.11/coursier/cli/Helper.scala:898:          parsedModule: ParsedModule <- allModuleVersionConfigs.headOption
./core/shared/src/main/scala/coursier/util/Parse.scala:117:  def moduleVersionConfig(s: String): Either[String, ParsedModule] =
./core/shared/src/main/scala/coursier/util/Parse.scala:118:    moduleVersionConfig(s, defaultScalaVersion)
./core/shared/src/main/scala/coursier/util/Parse.scala:139:  def moduleVersionConfig(s: String, defaultScalaVersion: String): Either[String, ParsedModule] = {
./core/shared/src/main/scala/coursier/util/Parse.scala:204:  def moduleVersionConfigs(l: Seq[String]): (Seq[String], Seq[ParsedModule]) =
./core/shared/src/main/scala/coursier/util/Parse.scala:205:    moduleVersionConfigs(l, defaultScalaVersion)
./core/shared/src/main/scala/coursier/util/Parse.scala:212:  def moduleVersionConfigs(l: Seq[String], defaultScalaVersion: String): (Seq[String], Seq[ParsedModule]) =
./core/shared/src/main/scala/coursier/util/Parse.scala:213:    valuesAndErrors(moduleVersionConfig(_, defaultScalaVersion), l)
./core/shared/src/main/scala/coursier/util/Parse.scala:117:  def moduleVersionConfig(s: String): Either[String, ParsedModule] =
./core/shared/src/main/scala/coursier/util/Parse.scala:118:    moduleVersionConfig(s, defaultScalaVersion)
./core/shared/src/main/scala/coursier/util/Parse.scala:139:  def moduleVersionConfig(s: String, defaultScalaVersion: String): Either[String, ParsedModule] = {
./core/shared/src/main/scala/coursier/util/Parse.scala:204:  def moduleVersionConfigs(l: Seq[String]): (Seq[String], Seq[ParsedModule]) =
./core/shared/src/main/scala/coursier/util/Parse.scala:205:    moduleVersionConfigs(l, defaultScalaVersion)
./core/shared/src/main/scala/coursier/util/Parse.scala:212:  def moduleVersionConfigs(l: Seq[String], defaultScalaVersion: String): (Seq[String], Seq[ParsedModule]) =
./core/shared/src/main/scala/coursier/util/Parse.scala:213:    valuesAndErrors(moduleVersionConfig(_, defaultScalaVersion), l)
```

alexarchambault:
> @alexarchambault the part changed in Parse actually does not concern anything but cli, can I move > that component to cli instead?

I think I'd prefer to keep it here, along with the other methods of `Parse`. These might be useful in other contexts (for users of coursier via its API).

alexarchambault:
About `ParsedModule`, couldn't it be replaced with just `Dependency`?

wisechengyi:
Hi @alexarchambault, I spent some time to replace `ParsedModule` with `Dependency` and simplified all the logic to make construct `Dependency`. 

However, there appears to be a backward compatibility issue, not sure if we can get past that. A quick search for `moduleVersionConfigs` did not appear to be used anywhere else.
https://github.com/search?q=org%3Acoursier+moduleVersionConfigs&type=Code
```
$  sbt ++2.11.11 coreJVM/mimaReportBinaryIssues
...
[error]  * deprecated method moduleVersionConfig(java.lang.String)scala.util.Either in object coursier.util.Parse does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("coursier.util.Parse.moduleVersionConfig")
[error]  * method moduleVersionConfig(java.lang.String,java.lang.String)scala.util.Either in object coursier.util.Parse does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("coursier.util.Parse.moduleVersionConfig")
[error]  * deprecated method moduleVersionConfigs(scala.collection.Seq)scala.Tuple2 in object coursier.util.Parse does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("coursier.util.Parse.moduleVersionConfigs")
[error]  * method moduleVersionConfigs(scala.collection.Seq,java.lang.String)scala.Tuple2 in object coursier.util.Parse does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("coursier.util.Parse.moduleVersionConfigs")
[error] java.lang.RuntimeException: coursier: Binary compatibility check failed!
```

wisechengyi:
Ok. The old interfaces are added back to manage the deprecation check.

wisechengyi:
ready for review when you get a chance @alexarchambault . thank you

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -113,44 +113,115 @@ object Parse {
     }
   }
 
+  case class ModuleParseError(private val message: String = "",
```
alexarchambault: Exceptions shouldn't be case classes. Two `ModuleParseError` with the same `message` and `cause` would be considered equal (because case class), but may not actually be equal (thrown from different places, so different stack trace). But a simple class is ok.

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -113,44 +113,115 @@ object Parse {
...
+                              private val cause: Throwable = None.orNull)
+    extends Exception(message, cause)
+
+  @deprecated("use the variant accepting a default scala version", "1.0.0-M13")
+  def moduleVersionConfig(s: String, scalaVersion: String): Either[String, Dependency] =
```
alexarchambault: The return types shouldn't be changed, to keep binary compatibility. (Don't know if mima complains about that, but it's sure it would lead to `ClassCastException` later on.)

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -113,44 +113,115 @@ object Parse {
...
+    moduleVersionConfig(s, Set.empty, Map.empty, transitive=true, scalaVersion)
+
   @deprecated("use the variant accepting a default scala version", "1.0.0-M13")
-  def moduleVersionConfig(s: String): Either[String, (Module, String, Option[String])] =
-    moduleVersionConfig(s, defaultScalaVersion)
+  def moduleVersionConfig(s: String): Either[String, Dependency] =
```
alexarchambault: Same as above.

cli/src/main/scala-2.11/coursier/cli/Helper.scala
```diff
@@ -339,30 +319,26 @@ class Helper(
       }).groupBy(_._1).mapValues(_.map(_._2).toSet).toMap
     }
 
-  val baseDependencies = allModuleVersionConfigs.map {
-    case (module, version, configOpt) =>
-      Dependency(
-        module,
-        version,
-        attributes = Attributes("", ""),
-        configuration = configOpt.getOrElse(defaultConfiguration),
```
alexarchambault: It seems `defaultConfiguration` isn't taken into account in the new code.

(Yes, the original `cli` code has poor test coverage :| )

wisechengyi: good catch. will correct this and add a test

core/shared/src/main/scala/coursier/util/Parse.scala
wisechengyi: mima did not complain about that. guess we just found out what limitation it has :) will correct.

wisechengyi: Oh also, how do you think we can be clear about when this can be removed?

core/shared/src/main/scala/coursier/util/Parse.scala
wisechengyi: corrected.

core/shared/src/main/scala/coursier/util/Parse.scala
wisechengyi: reverted the interface

core/shared/src/main/scala/coursier/util/Parse.scala
wisechengyi: done.

wisechengyi:
Addressed all the comments, except CI is running into the same error as master does. Please let me know if it's good to go @alexarchambault

wisechengyi:
Planning on merging this one in the next 24 hours if no further feedback given.

alexarchambault:
So, merging.

PR #742, opened 2018-01-17
mixing urls and jars in bootstrap

bosyi:
For not standalone bootstrap in case of file:/ uri add file to pack, in case of http[s]:// add to urls, in other cases do nothing and show warning.

### DIFF
--- a/cli/src/main/scala-2.11/coursier/cli/Bootstrap.scala
+++ b/cli/src/main/scala-2.11/coursier/cli/Bootstrap.scala
@@ -126,13 +126,8 @@ final case class Bootstrap(
       ).toList.foldLeft((List.empty[String], List.empty[File])){
         case ((urls, files), (url, file)) =>
           if (options.standalone) (urls, file :: files)
-          else url match {
-            case _ if url.startsWith("http://") || url.startsWith("https://") => (url :: urls, files)
-            case _ if url.startsWith("file:/") => (urls, file :: files)
-            case _ =>
-              Console.err.println(s"Warning: non HTTP or FILE URL:\n${url}")
-              (urls, files)
-          }
+          else if (url.startsWith("file:/")) (urls, file :: files)
+          else (url :: urls, files)
       }
 
     val isolatedUrls = isolatedArtifactFiles.map { case (k, (v, _)) => k -> v }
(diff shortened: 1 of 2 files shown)


cli/src/main/scala-2.11/coursier/cli/Bootstrap.scala
```diff
@@ -119,32 +119,25 @@ final case class Bootstrap(
...
-          Seq.empty[File]
-        )
+      helper.fetchMap(
+        sources = false,
+        javadoc = false,
+        artifactTypes = artifactOptions.artifactTypes(sources = false, javadoc = false)
+      ).toList.foldLeft((List.empty[String], List.empty[File])){
+        case ((urls, files), (url, file)) =>
+          if (options.standalone) (urls, file :: files)
+          else url match {
```
alexarchambault: Nitpicking: that `match` doesn't really match anything, it should just be a bunch of `if` / `else`.

bosyi:
Refactor and change logic a bit. Now if it is 'file' add to pack otherwise add as url link.
First point is it will work as previous behavior,
Second is that maybe someone using non standard protocols to retrieve artifacts.

cli/src/main/scala-2.11/coursier/cli/Helper.scala
```diff
@@ -657,12 +657,12 @@ class Helper(
...
   }
 
-  def fetch(
+  def fetchMap(
     sources: Boolean,
     javadoc: Boolean,
     artifactTypes: Set[String],
     subset: Set[Dependency] = null
-  ): Seq[File] = {
+  ): Map[String,File] = {
```
wisechengyi: 1. Could you docstring for this interface since the return type is changed?
2. nit: space after `,`

wisechengyi: Hmm this should cause the CI to fail due to other code using this. Will investigate..

wisechengyi: Oh I see. `def fetch` still calls this. 

Looks good to me, just nits above. Thanks!

alexarchambault:
Merging, thanks @bosyi and @wisechengyi.

Issue #752, opened 2018-01-24
master ci broken


wisechengyi:
https://travis-ci.org/coursier/coursier/builds/331611470, with errors like:
```
----------------------------------- Failures -----------------------------------
X coursier.test.CentralNexus3ProxyTests.versionInterval 13195ms 
  utest.AssertionError: assert(result == expected)
  result: scala.collection.immutable.Vector[String] = Vector(org.webjars.bower:jquery:3.3.1:compile,
   org.webjars.bower:jquery-mousewheel:3.1.13:compile, org.webjars.bower:malihu-custom-scrollbar-plu
  gin:3.1.5:compile)
  expected: Seq[String] = WrappedArray(org.webjars.bower:jquery:3.2.1:compile, org.webjars.bower:jqu
  ery-mousewheel:3.1.13:compile, org.webjars.bower:malihu-custom-scrollbar-plugin:3.1.5:compile)
    utest.asserts.Asserts$.assertImpl(Asserts.scala:114)
    coursier.test.CentralTests$stateMachine$macro$28$1.apply(CentralTests.scala:122)
    coursier.test.CentralTests$stateMachine$macro$28$1.apply(CentralTests.scala:63)
    scala.concurrent.impl.CallbackRunnable.run(Promise.scala:60)
    scala.concurrent.impl.ExecutionContextImpl$AdaptedForkJoinTask.exec(ExecutionContextImpl.scala:1
    40)
    java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
    java.util.concurrent.ForkJoinPool$WorkQueue.pollAndExecAll(ForkJoinPool.java:1021)
    java.util.concurrent.ForkJoinPool$WorkQueue.execLocalTasks(ForkJoinPool.java:1046)
    java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1058)
    java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1692)
    java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)
X coursier.test.CentralNexus2ProxyTests.versionInterval 11528ms 
  utest.AssertionError: assert(result == expected)
  result: scala.collection.immutable.Vector[String] = Vector(org.webjars.bower:jquery:3.3.1:compile,
   org.webjars.bower:jquery-mousewheel:3.1.13:compile, org.webjars.bower:malihu-custom-scrollbar-plu
[...]

PR #753, opened 2018-01-24
Temporarily disable nexus test to unblock master

wisechengyi:
To get around #752


wisechengyi:
appveyor appears to be flaky as well, so I removed its cache. maybe having access to appveyor settings could be helpful in the future.

PR #738, opened 2018-01-14
Upgrade to sbt 1.1.0, update sbt-pack

leonardehrenfried:
Why is it good to upgrade sbt? Because 1.1.0 has support for LSP which in turn lets you use VSCode. (I personally have problems loading the build in IntelliJ - VSCode is better than nothing)

PR #759, opened 2018-01-29
Update build to sbt 1.1.0

cunei:
sbt 1.1.0 requires an updated sbt-pack, so update that as well


alexarchambault:
@cunei Switching to sbt 1.1.0 has been tried already in https://github.com/coursier/coursier/pull/738.

cunei:
My bad, I forgot about that.

PR #761, opened 2018-01-30
Allow to deactivate Maven profiles from the CLI

alexarchambault:
and add extra tests for profile deactivation


cli/src/main/scala-2.11/coursier/cli/Helper.scala
```diff
@@ -319,7 +319,7 @@ class Helper(
     filter = Some(dep => keepOptional || !dep.optional),
     userActivations =
       if (userEnabledProfiles.isEmpty) None
-      else Some(userEnabledProfiles.iterator.map(_ -> true).toMap),
+      else Some(userEnabledProfiles.iterator.map(p => if (p.startsWith("!")) p.drop(1) -> false else p -> true).toMap),
```
wisechengyi: Thanks @alexarchambault .lgtm. If you could add an unit test on `userActivation` that'll be even better.

alexarchambault: What do you mean? Profiles - so the `userActivations` field - are already [tested a bit](https://github.com/coursier/coursier/blob/1ddfbed8a86bdfe289999f6ab8afbfe58d438f21/tests/shared/src/test/scala/coursier/test/CentralTests.scala#L42).

PR #763, opened 2018-02-02
Use released version of sbt-compat

leonardehrenfried:
Dale was so kinda as to cut a release last night.


build.sbt
```diff
@@ -238,9 +238,9 @@ lazy val `sbt-shared` = project
   .settings(
     plugin,
     utest,
-//  addSbtPlugin("com.dwijnand" % "sbt-compat" % "1.2.0+2-f30b82f4")
+//  addSbtPlugin("com.dwijnand" % "sbt-compat" % "1.2.1")
```
wisechengyi: could you remove this line, or add a comment why this comment is needed?

leonardehrenfried: Sure.

build.sbt
```diff
@@ -238,9 +238,11 @@ lazy val `sbt-shared` = project
   .settings(
     plugin,
     utest,
-//  addSbtPlugin("com.dwijnand" % "sbt-compat" % "1.2.0+2-f30b82f4")
+// once sbt 0.13.17 is released the following line can be used
```
dwijnand: sbt 0.13.17 was released: https://github.com/sbt/sbt/releases/tag/v0.13.17

dwijnand: doesn't matter, the reason isn't because of sbt 0.13.17, it's because sbt-compat should only be added here for Scala 2.10 and 2.12

dwijnand: see https://github.com/coursier/coursier/pull/731#issuecomment-356344602

leonardehrenfried: Ok, so we will never be able to use the short syntax?

In that case, we should remove the commented out version completely.

PR #746, opened 2018-01-19
Import correct resolver from librarymanagement

leonardehrenfried:
This might be a solution for #745 by importing the correct version of the `Resolver` type from `sbt.librarymanagement` rather than `sbt` thereby sidestepping the issue described here: https://github.com/sbt/librarymanagement/pull/190#issuecomment-358897786

PR #764, opened 2018-02-02
[sbt-shared] Use sbt.util.Logger instead of sbt.Logger

leonardehrenfried:
This is the same problem as #746.

As of sbt 1.0.0 the type has moved to `sbt.util` and using the old one creates a weird compilation issue when trying to integrate with coursier with sbt.

See https://travis-ci.org/sbt/librarymanagement/jobs/336516938#L625

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -241,7 +241,7 @@ lazy val `sbt-shared` = project
     // because we don't publish for 2.11 the following declaration
     // is more wordy than usual
     libs ++= {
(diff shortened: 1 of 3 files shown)


leonardehrenfried:
Alright, seems like I have to add code to sbt-compat to make this work in sbt 0.13. *sigh*

alexarchambault:
Thanks for all that @leonardehrenfried, merging!

PR #767, opened 2018-02-05
[wip] Speed up travis builds by using more memory

leonardehrenfried:
coursier is a large build and could benefit from more memory.

This PR uses the container-based travis infrastructure (https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure) and gives sbt 4GB of memory during the build.

Lets hope this leads to actual speed ups.


leonardehrenfried:
Hmm, I don't see any significant speed up of the build time. :|

leonardehrenfried:
Travis seems to be quite busy today. I'm going to try again during a less busy time and see if it makes a difference.

wisechengyi:
It does not work because container based shard is capped at 4GB RAM. https://docs.travis-ci.com/user/reference/overview/#Virtualisation-Environment-vs-Operating-System

You can use:
```
sudo: required 
dist: trusty
```
to get more RAM with slightly higher startup cost.

leonardehrenfried:
Thanks.

I'm going to play around with these settings and see if there are speed ups
to be had.

Am 05.02.2018 7:20 nachm. schrieb "Yi Cheng" <notifications@github.com>:

wisechengyi:
let me know when you are done iterating :)

leonardehrenfried:
I tried a few combinations but so far no noticeable speed ups have been
achieved.

Am 07.02.2018 2:42 vorm. schrieb "Yi Cheng" <notifications@github.com>:

leonardehrenfried:
Hmm, it seems that the slow part is generating all that Scaladoc. I don't think it can be sped up. :|

PR #769, opened 2018-02-07
Separate out bootstrap integration test

wisechengyi:
Separate bootstrap test because Pants does not pre-build the bootstrap.jar


wisechengyi:
cc @dotordogh

Issue #710, opened 2017-12-09
[question] Does coursier support resolve by file path?


wisechengyi:
Basically resolve an artifact based on a relative or absolution file path. For example:

https://stackoverflow.com/questions/2229757/maven-add-a-dependency-to-a-jar-by-relative-path

so in some cases it can be useful for user who wants to resolve adhoc jars.

alexarchambault:
@wisechengyi One can always add file-based repositories, like `coursier fetch -r file:///path/to/repo …`

wisechengyi:
How do you feel about if I add the feature to take a single jar file path that is not in any repo (with no pom, so intransitivity implied)? Basically it allows user to consume a single jar without having to create a local maven repo.

wisechengyi:
@alexarchambault ^

alexarchambault:
@wisechengyi I'm not sure I understand what you mean… I've been willing to add something like a `download` command to the cli, like
```
$ coursier download https://foo.com/something
```
to download and cache the passed url, relying on the cache / ttl / etc. of coursier. Is it what you mean?

wisechengyi:
Hi @alexarchambault, sorry about the delay. Yes I think so, although I was thinking adding it directly to `fetch` command, so it would be something like:
```
coursier fetch --intransitive a:b:c:url=file://.../x.jar
```
and then use coursier's cache ttl to manage. Say `x.jar` changes within the TTL, then coursier should not fetch from the path again.

alexarchambault:
You mean, the supplied file would act as a dependency-free dependency, with only the supplied file as artifact? It's not the exact use case I was thinking of, but I'm not against it either!

wisechengyi:
Correct. Cool thanks.

alexarchambault:
FYI There are already [related things](https://github.com/coursier/coursier/blob/a41c07194344555f5f9ee7a6e5b3c01ff13afa7b/sbt-coursier/src/main/scala/coursier/FallbackDependenciesRepository.scala) for the sbt plugin (to handle things like `libraryDependencies += "org" % "name" % "ver" from "url"`)

wisechengyi:
Yeah I think that's the same idea, so maybe just need to plumb it for cli then.

wisechengyi:
@alexarchambault the syntax I'd like to propose is `org:name:version,classifier=<classifier>,url=<encoded url file://... or https://...>` by leveraging #735

wisechengyi:
Resolved by #774

PR #774, opened 2018-02-10
Add ability to fetch artifact with a given url

dotordogh:
## Problem
We want to be able to fetch an artifact given a url. See #710 

## Solution
Add a new field in `Attribute` and assign it as the url in `Artifact`. 

I also added a few tests to make sure fetch behaves as expected.


core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -177,9 +178,13 @@ object Parse {
 
     val parts = coords.split(":", 5)
 
-    val attributes = attrs.get("classifier") match {
-      case Some(c) => Attributes("", c)
-      case None => Attributes("", "")
+    val attributes = Attributes("", attrs.getOrElse("classifier", ""), decode(attrs.getOrElse("url", ""), "UTF-8"))
```
wisechengyi: While you are at it, could you add a check to throw an error if `attrs` contains things other than `classifier` or `url`?

tests/shared/src/test/scala/coursier/test/ParseTests.scala
```diff
@@ -86,6 +86,32 @@ object ParseTests extends TestSuite {
...
 
+    "single attr with url" - {
+      Parse.moduleVersionConfig("org.apache.avro:avro:1.7.4:runtime,url=file%3A%2F%2Fsome%2Fencoded%2Furl", ModuleRequirements(), transitive = true, "2.11.11") match {
+        case Left(err) => assert(false)
+        case Right(dep) =>
+          assert(dep.module.organization == "org.apache.avro")
+          assert(dep.module.name == "avro")
+          assert(dep.version == "1.7.4")
+          assert(dep.configuration == "runtime")
+          assert(dep.attributes == Attributes("", "", "file://some/encoded/url"))
```
wisechengyi: How about making `"file://some/encoded/url"` a variable, and then make the moduleVersionConfig to be `"org.apache.avro:avro:1.7.4:runtime,url=" +  encoded(...)`?

Same for below

cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala
```diff
@@ -380,4 +381,232 @@ class CliFetchIntegrationTest extends FlatSpec with CliTestLib {
...
+    */
+  "local dep url" should "have coursier-fetch-test.jar" in withFile() {
+    (excludeFile, _) =>
+      withFile() {
+        (jsonFile, _) => {
+          // generate temp jar with fake content
+          val filename = "coursier-fetch-test.jar"
+          val jarFileContent = "tada"
+          val file = new File(filename)
+          val bw = new BufferedWriter(new FileWriter(file))
```
wisechengyi: You should be able to levearge `withFile` above, like `withFile("tada", filename, ext)`. Slight modification to `withFile` is required, so you won't need to do all the file operations here.

wisechengyi:
@alexarchambault what would be your suggestion how to handle the incompatibility?
```
[info] coursier: found 5 potential binary incompatibilities while checking against io.get-coursier:coursier_2.10:1.0.0-RC3  (filtered 319)
[error]  * method apply(java.lang.String,java.lang.String)coursier.core.Attributes in object coursier.package#Attributes does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("coursier.package#Attributes.apply")
[error]  * the type hierarchy of object coursier.core.Attributes is different in current version. Missing types {scala.runtime.AbstractFunction2}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("coursier.core.Attributes$")
[error]  * method apply(java.lang.String,java.lang.String)coursier.core.Attributes in object coursier.core.Attributes does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("coursier.core.Attributes.apply")
[error]  * method copy(java.lang.String,java.lang.String)coursier.core.Attributes in class coursier.core.Attributes does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("coursier.core.Attributes.copy")
[error]  * method this(java.lang.String,java.lang.String)Unit in class coursier.core.Attributes does not have a correspondent in current version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("coursier.core.Attributes.this")
[error] java.lang.RuntimeException: coursier: Binary compatibility check failed!
```

alexarchambault:
Thanks @dotordogh!

As @wisechengyi noted, this still needs the POM files to be available in a repository for that to work. Is that intended?

Also, did you try to do that using the same trick as the sbt plugin, that is relying on a [custom repository](https://github.com/coursier/coursier/blob/df58a68a393400ab61b22aa1e30dfbd15b72bfed/sbt-coursier/src/main/scala/coursier/FallbackDependenciesRepository.scala)? (mentioned in https://github.com/coursier/coursier/issues/710 IIRC) I think [FallbackDependenciesRepository](https://github.com/coursier/coursier/blob/df58a68a393400ab61b22aa1e30dfbd15b72bfed/sbt-coursier/src/main/scala/coursier/FallbackDependenciesRepository.scala) could be used almost as is to handle that. It could be moved to the `extra` submodule to be available from `cli` too. (It has a few quirks - it's doing blocking things where it shouldn't, it directly calls `java.net.HttpURLConnection` instead of relying on the cache, … - but it should do the job in a first time.)

The advantage of this approach is that one has to change very little things in `core` for that to work. (`core` is the most sensitive module of coursier, so the less concerns handled in it, the better.)

dotordogh:
Thank you @wisechengyi and @alexarchambault! I overlooked the fact that the resolve still needs the POM, my apologies! 

I've been looking at FallbackDependenciesRepository and how I can use it, but I think I may be missing a piece. Are you thinking of something like 1) filter out the dependencies that have a url defined, 2) use FallbackDependenciesRepository to create artifacts for the dependencies with urls, and combine those artifacts with the artifacts that were produced as a result of the regular resolved for the other dependencies and then 3) perform the fetch?

alexarchambault:
@dotordogh Dependencies passed on the command line can be added as usual, including those with a manual uri (the uri doesn't appear in their `Dependency`, but it doesn't matter there). If some dependencies have a manual uri, they can be added to the `FallbackDependenciesRepository` with their uri. `FallbackDependenciesRepository` can be prepended to the repository list before the resolution. That way, it will be picked first, and any dependency it has will be handled by it.

I assumed the provided URIs should be used even if the dependency can be found in a repo. Should they for you? In sbt, it's only a [fallback](https://www.scala-sbt.org/1.0/docs/Library-Management.html#Explicit+URL), so the `FallbackDependencyRepository` is added [in last position](https://github.com/coursier/coursier/blob/7b12da18b68731c3599d4e2ef841506a5df42bfa/sbt-coursier/src/main/scala/coursier/Tasks.scala#L887) to the repo list.

alexarchambault:
> @alexarchambault does FallbackDependenciesRepository create a repo on disc or is it in memory?

`FallbackDependencyRepository` is entirely in memory, yes.

dotordogh:
There are some tests that are still failing, this is just a preliminary push to see if I'm on the right track. Will update when the tests pass.

dotordogh:
I still need to address @wisechengyi's comments and there are two more tests I would like to add, but feeling much better about the diff.

build.sbt
```diff
@@ -63,6 +63,7 @@ lazy val tests = crossProject
     hasITs,
     coursierPrefix,
     libs += Deps.scalaAsync.value,
+    pureJava,
```
alexarchambault: That shouldn't be necessary, it's only for Java-only projects. The scala-js errors can be addressed…

cli/src/main/scala-2.12/coursier/cli/Helper.scala
```diff
@@ -273,11 +272,37 @@ class Helper(
...
+  val allDependencies: Seq[Dependency] = allDependenciesWithExtraParams.map(dep => dep._1)
+
+  // Any dependencies with URIs should not be resolved with a pom so this is a
+  // hack to add all the deps with URIs to the FallbackDependenciesRepository
+  // which will be used during the resolve
+  val depsWithUrls: Map[(Module, String), (URL, Boolean)] = {
+    allDependenciesWithExtraParams
+      .filter(depWithExtraParams => depWithExtraParams._2.isDefinedAt("url"))
+      .map(depWithExtraParams =>
+        (depWithExtraParams._1.moduleVersion, (new URL(depWithExtraParams._2.getOrElse("url", "")), true))).toMap
```
alexarchambault: It's possible to check the presence of and get the `"url"` value in one pass, like
```scala
allDependenciesWithExtraParams
  .flatMap {
    case (dep, extraParams) =>
      extraParams.get("url").map { url =>
        dep.moduleVersion -> (new URL(url), true)
      }
  }
```

cli/src/main/scala-2.12/coursier/cli/Helper.scala
```diff
@@ -273,11 +272,37 @@ class Helper(
...
 
-  val allDependencies: Seq[Dependency] = transitiveDeps ++ intransitiveDeps
+  // Prepend FallbackDependenciesRepository to the repository list
+  // so that dependencies with URIs are resolved against this repo
+  val repositories: Seq[Repository] = depsWithUrlRepo ++ standardRepositories
+
+  depsWithUrls.foreach(dep =>
+    if (forceVersions.contains(dep._1._1) && forceVersions.getOrElse(dep._1._1, dep._1._1) != dep._1._2)
+      throw new Exception("Cannot specify a URL for a dependency for which a version is being forced")
+  )
```
alexarchambault: It should be possible to avoid the `_?._?`, like
```scala
for (((mod, ver), _) <- depsWithUrls if forceVersions.get(mod).exists(_ != ver))
  throw …
```

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -175,11 +182,19 @@ object Parse {
...
-    val attributes = attrs.get("classifier") match {
-      case Some(c) => Attributes("", c)
-      case None => Attributes("", "")
+    val attributes = Attributes("", attrs.getOrElse("classifier", ""))
+
+    val extraDependencyParams: Map[String, String] = {
+      if (attrs.isDefinedAt("url"))
+        Map("url" -> decode(attrs.getOrElse("url", ""), "UTF-8"))
+      else
+        Map()
```
alexarchambault: A bit like above, we can avoid the `isDefinedAt` / `getOrElse`, with
```scala
attrs.get("url") match {
  case Some(url) => Map("url" -> decode(url, "UTF-8"))
  case None => Map()
}
```

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -191,59 +206,81 @@ object Parse {
...
-              transitive = transitive,
-              exclusions = localExcludes.getOrElse(mod.orgName, Set()) | globalExcludes)
+            (Dependency(
+                mod,
+                version,
+                config,
+                attributes,
+                transitive = transitive,
+                exclusions = localExcludes.getOrElse(mod.orgName, Set()) | globalExcludes),
+              extraDependencyParams)
```
alexarchambault: As `extraDependencyParams` is added to all the right cases, we could do something like
```scala
val depOrError = parts match {
   …
}

depOrError.right.map(dep => (dep, extraDependencyParams))
```
(only adding extraDependencyParams at one place).

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -191,59 +206,81 @@ object Parse {
...
+   * added, they should be passed in via the second parameter
+   *
+   * @param attrs Attributes parsed
+   * @param validAttrsKeys Valid attribute keys
+   */
+  private def validateAttributes(attrs: Map[String, String], validAttrsKeys: Set[String]): Unit = {
+    val extraAttributes = attrs.keys.toSet.diff(validAttrsKeys)
+
+    if (attrs.size > validAttrsKeys.size || extraAttributes.nonEmpty)
+      throw new ModuleParseError(s"The only attributes allowed are: ${validAttrsKeys.mkString(", ")}. ${
```
alexarchambault: I'll probably deprecate `ModuleParseError` in a later PR. In the public methods of this file, errors should be returned via a `Left("error message")` rather than an exception. (I guess I underlooked it when it was added :| )

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -5,6 +5,7 @@ import coursier.core.{Module, Repository}
 import coursier.ivy.IvyRepository
 import coursier.maven.MavenRepository
 
+import java.net.URLDecoder.decode
```
alexarchambault: Couldn't we do the `decode` part only from where the `"url"` value is used, that is in the `cli` module for now? That way, we wouldn't have to do an `encode` for it in the tests. (`URLEncoder.encode` seems not to be available from scala-js, hence the current CI errors, so that would solve that to.)

cli/src/main/scala-2.12/coursier/cli/Helper.scala
```diff
@@ -800,6 +825,7 @@ class Helper(
 
         // Trying to get the main class of the first artifact
         val mainClassOpt = for {
+
```
wisechengyi: xx

extra/src/main/scala/coursier/FallbackDependenciesRepository.scala
```diff
@@ -1,13 +1,11 @@
 package coursier
 
-import java.io.{ File, FileNotFoundException, IOException }
-import java.net.{ HttpURLConnection, URL, URLConnection }
-
+import java.io.{File, FileNotFoundException, IOException}
```
wisechengyi: Let's leave this file untouched since there's no functional change. If you would like to format the file, a separate PR would be better.

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -146,12 +153,12 @@ object Parse {
     *  or
     *   org:name:version:config,attr1=val1,attr2=val2
     *
-    *  Currently only "classifier" attribute is used, and others are ignored.
+    *  Currently only the "classifier" attribute is used, and others are ignored.
```
wisechengyi: Docstring needs update since 'classifier' and 'url' are in use. The return value is also changed.

cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala
```diff
@@ -380,4 +381,355 @@ class CliFetchIntegrationTest extends FlatSpec with CliTestLib {
...
+
+          val urlInJsonFile = depNodes.head.files.head._2
+          assert(depNodes.head.files.head._1 == "")
+          assert(urlInJsonFile.contains(path))
+
+          // open jar and inspect contents
+          val fileContents = Source.fromFile(urlInJsonFile).getLines.mkString
+          assert(fileContents == "tada")
+
+
```
wisechengyi: nit: xx empty lines

core/shared/src/main/scala/coursier/util/Parse.scala
wisechengyi: my bad :)

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -164,29 +171,38 @@ object Parse {
...
+        return Left(s"Failed to parse attribute '$x' in '$s'. Keyword argument expected such as 'classifier=tests'")
       }
       (y(0), y(1))
     }
     }).toMap
 
+    // Only "classifier" and "url" attributes are allowed
+    val validAttrsKeys = Set("classifier", "url")
+
+    val maybeErrorMsg = validateAttributes(attrs, validAttrsKeys)
```
alexarchambault: Instead of the `.isDefined` / ` return Left(…)`, we can do something like
```scala
validateAttributes(attrs, validAttrsKeys) match {
  case Some(err) => Left(err)
  case None =>
    …
}
```
which uses neither `.get` nor `return` (`.get` is ~unsafe, as it can throw exceptions if ever we forgot it to ensure it's a `Some(…)`, and `return` has [many drawbacks](https://tpolecat.github.io/2014/05/09/return.html) in Scala).

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -164,29 +171,38 @@ object Parse {
...
     val attrs = strings.drop(1).map({ x => {
       if (x.mkString.contains(argSeparator)) {
-        throw new ModuleParseError(s"'$argSeparator' is not allowed in attribute '$x' in '$s'. Please follow the format " +
+        return Left(s"'$argSeparator' is not allowed in attribute '$x' in '$s'. Please follow the format " +
           s"'org${argSeparator}name[${argSeparator}version][${argSeparator}config]${attrSeparator}attr1=val1${attrSeparator}attr2=val2'")
       }
       val y = x.split("=")
       if (y.length != 2) {
-        throw new ModuleParseError(s"Failed to parse attribute '$x' in '$s'. Keyword argument expected such as 'classifier=tests'")
+        return Left(s"Failed to parse attribute '$x' in '$s'. Keyword argument expected such as 'classifier=tests'")
```
alexarchambault: The `return`s could be avoided like the one below, although is less straightforward there. (Basically, we can `map` without using return, getting a sequence of eithers, then collect all the lefts on the one hand, and all the rights on the other, and check that the former is empty - but I can look into it later if you'd like, it's still better than throwing exceptions)

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -164,29 +171,38 @@ object Parse {
...
       }
       (y(0), y(1))
     }
     }).toMap
 
+    // Only "classifier" and "url" attributes are allowed
+    val validAttrsKeys = Set("classifier", "url")
```
alexarchambault: Do we really want to restrict the attribute keys here? But we can loosen that in a later PR I guess.

wisechengyi: Yeah I would lean toward restricting it, because it provides clear error msg when seeing things it cannot handle, rather than silently ignoring it.

core/shared/src/main/scala/coursier/util/Parse.scala
dotordogh: Thank you so much for all of the tips!! It's been a while since I used scala and I forgot I could do all this stuff! I tried implementing what you suggested here but ran into some issues, so I thought I would fix the `isDefined`/`return Left(...)` issue and return to this when I have some more time if you haven't done it already. Does that sound alright to you?

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -164,29 +171,42 @@ object Parse {
...
     }
     }).toMap
 
     val parts = coords.split(":", 5)
 
+    // Only "classifier" and "url" attributes are allowed
+    val validAttrsKeys = Set("classifier", "url")
+
+    validateAttributes(attrs, validAttrsKeys) match {
+      case Some(err) => return Left(err)
```
dotordogh: I added this return statement here so that this error would behave the same way as the above errors do and stops the function from proceeding. If the return is not present, it doesn't stop the function and parsing will continue even though the attributes are invalid.

core/shared/src/main/scala/coursier/util/Parse.scala
```diff
@@ -242,6 +262,28 @@ object Parse {
...
+   * @param attrs Attributes parsed
+   * @param validAttrsKeys Valid attribute keys
+   * @return A string if there is an error, otherwise None
+   */
+  private def validateAttributes(attrs: Map[String, String], validAttrsKeys: Set[String]): Option[String] = {
+    val extraAttributes = attrs.keys.toSet.diff(validAttrsKeys)
+
+    if (attrs.size > validAttrsKeys.size || extraAttributes.nonEmpty)
+      Some(s"The only attributes allowed are: ${validAttrsKeys.mkString(", ")}. ${
+        if (extraAttributes.nonEmpty) s"The following are invalid: ${extraAttributes.mkString(", ")}"
```
wisechengyi: nit: I feel the error can be more descriptive, e.g. for `junit:junit:4.12,abc=123`:
```
The only attributes allowed are: classifier, url. The following attribute(s) are invalid: 'abc' in 'junit:junit:4.12,abc=123'
```

alexarchambault:
Not merging right now for you to see https://github.com/coursier/coursier/pull/774/commits/e0bd15beb1dc71487dfe1a90969c2d616886bf3d, but that can be merged IMO else.

dotordogh:
Ahhhhh I see what you mean now! Very cool! Thank you!!

alexarchambault:
Cool! Merging, then, thanks again!

Issue #743, opened 2018-01-18
jar dependency match with json report


wisechengyi:
### Problem

Coursier command:
```
fetch -t \
com.esotericsoftware.reflectasm:reflectasm:1.09,classifier=shaded \ # from pants jar_dep A
com.esotericsoftware.reflectasm:reflectasm:1.09 \ # from pants jar_dep B
org.ow2.asm:asm:4.0,classifier=sources \ # from pants jar_dep C
-A jar,src,doc,test-jar --json-output-file x.out

  Result:
├─ com.esotericsoftware.reflectasm:reflectasm:1.09
│  └─ org.ow2.asm:asm:4.0
├─ com.esotericsoftware.reflectasm:reflectasm:1.09
│  └─ org.ow2.asm:asm:4.0
└─ org.ow2.asm:asm:4.0
```

Since there are 3 roots to resolve, the tree is correct. One `com.esotericsoftware.reflectasm:reflectasm:1.09` wants the default jar, and the other wants the `shaded` classifier jar. (UI wise it could be better to show that which is which)

Now look at the json output:
```
{
  "conflict_resolution": {},
  "dependencies": [
    {
      "coord": "org.ow2.asm:asm:4.0",
      "files": [
        [
          "sources",
          "<cache_dir>/https/repo1.maven.org/maven2/org/ow2/asm/asm/4.0/asm-4.0-sources.jar"
        ]
      ],
      "dependencies": []
    },
    {
      "coord": "com.esotericsoftware.reflectasm:reflectasm:1.09",
      "files": [
        [
          "shaded",
          "<cache_dir>/https/repo1.maven.org/maven2/com/esotericsoftware/reflectasm/reflectasm/1.09/reflectasm-1.09-shaded.jar"
        ]
      ],
      "dependencies": [
        "org.ow2.asm:asm:4.0"
      ]
    },
    {
      "coord": "org.ow2.asm:asm:4.0",
      "files": [
        [
          "",
          "<cache_dir>/https/repo1.maven.org/maven2/org/ow2/asm/asm/4.0/asm-4.0.jar"
        ]
      ],
      "dependencies": []
    },
    {
      "coord": "com.esotericsoftware.reflectasm:reflectasm:1.09",
      "files": [
        [
          "",
          "<cache_dir>/https/repo1.maven.org/maven2/com/esotericsoftware/reflectasm/reflectasm/1.09/reflectasm-1.09.jar"
        ]
      ],
      "dependencies": [
[...]

PR #782, opened 2018-02-16
json-report: one file per dependency instead of multiple; use m2 coords

baroquebobcat:
When classifiers are used as part of dependency specifications, it's important to be able to select just the classified artifact. Unfortunately, in the current json, dependencies don't specify classifiers, so it isn't possible to just get one of the artifacts for a dependency when a classifier is required.

This patch introduces maven style artifact prefixes in order to include classifier and packaging information in the coordinates. By doing that, we can use them as keys in dependency lists more easily and it allows consumers of the json to treat those dependency keys as mostly opaque ids rather than having to parse them.

Addresses #743


alexarchambault:
@baroquebobcat The CI error seems legit… `contains` is called on an `Option[String]`, not on the contained `String` (I guess you want something like `.exists(_.contains("…"))`)

baroquebobcat:
Yeah. Thanks for looking at it.

alexarchambault:
@baroquebobcat New CI failure seems legit too.

wisechengyi:
Turns out I left something when splitting the tests https://github.com/coursier/coursier/pull/788, so CI should pass after that.

cli/src/test/scala-2.12/coursier/cli/util/JsonReportTest.scala
```diff
@@ -0,0 +1,58 @@
...
+  "empty JsonReport" should "be empty" in {
+    val report: String = JsonReport[String](IndexedSeq(), Map())(
+      children = _ => Seq(),
+      reconciledVersionStr = _ => "",
+      requestedVersionStr = _ => "",
+      getFile = _ => Option("")
+    )
+
+    assert(
+      report == "{\"conflict_resolution\":{},\"dependencies\":[],\"version\":\"0.1.0\"}")
```
wisechengyi: Any particular reason string based assertion is used rather than parsing the json report then assert on structured data? https://github.com/coursier/coursier/blob/master/cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala#L18

baroquebobcat: I think the assertion failure message is a bit clearer with the string, and the string is the actual output, so it feels like the right thing to test against. That said, it could be a bit too fragile.

wisechengyi:
This error consistently in windows CI with this change:
https://github.com/coursier/coursier/pull/782/files#diff-86042d8fbf74d324dff94d39c3760dfbR1073

```
C:\Users\appveyor\AppData\Local\Temp\1\sbt_a021fee1\scala-sources-javadoc-jars\build.sbt:35)
[00:32:20] [info] 	at $66918475b6504a4b75df$$anonfun$$sbtdef$1.apply(C:\Users\appveyor\AppData\Local\Temp\1\sbt_a021fee1\scala-sources-javadoc-jars\build.sbt:40)
[00:32:20] [info] 	at $66918475b6504a4b75df$$anonfun$$sbtdef$1.apply(C:\Users\appveyor\AppData\Local\Temp\1\sbt_a021fee1\scala-sources-javadoc-jars\build.sbt:8)
[00:32:20] [info] 	at scala.Function1$$anonfun$compose$1.apply(Function1.scala:47)
[00:32:20] [info] 	at sbt.$tilde$greater$$anonfun$$u2219$1.apply(TypeFunctions.scala:40)
[00:32:20] [info] 	at sbt.std.Transform$$anon$4.work(System.scala:63)
[00:32:20] [info] 	at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
[00:32:20] [info] 	at sbt.Execute$$anonfun$submit$1$$anonfun$apply$1.apply(Execute.scala:226)
[00:32:20] [info] 	at sbt.ErrorHandling$.wideConvert(ErrorHandling.scala:17)
[00:32:20] [info] 	at sbt.Execute.work(Execute.scala:235)
[00:32:20] [info] 	at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:226)
[00:32:20] [info] 	at sbt.Execute$$anonfun$submit$1.apply(Execute.scala:226)
[00:32:20] [info] 	at sbt.ConcurrentRestrictions$$anon$4$$anonfun$1.apply(ConcurrentRestrictions.scala:159)
[00:32:20] [info] 	at sbt.CompletionService$$anon$2.call(CompletionService.scala:28)
[00:32:20] [info] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[00:32:20] [info] 	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
[00:32:20] [info] 	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
[00:32:20] [info] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[00:32:20] [info] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[00:32:20] [info] 	at java.lang.Thread.run(Thread.java:748)
[00:32:20] [info] [error] (*:updateClassifiersCheck) java.lang.AssertionError: assertion failed: scala-library javadoc not found
[00:32:20] [info] [error] Total time: 2 s, completed Mar 3, 2018 12:38:34 AM
```

I also verified it by singling it out at https://github.com/coursier/coursier/pull/750

The test should be covered in travis as well, do you have any clue why this would be case, @alexarchambault?

alexarchambault:
It's possibly https://github.com/coursier/coursier/pull/782/commits/afe30721a00cc9790edc1289d58240d8eade6704… From what I remember, the scripted test that fails corresponds to things that are a bit hackish (but with non reg tests). I'll have a look at it.

wisechengyi:
@alexarchambault the link is invalid btw

baroquebobcat:
It looks like I've got the CI issue resolved. It seems that the change I made to Resolution broke some assumptions. I've reverted that and instead do the dep flattening in Helper.

wisechengyi:
Thanks. We are still iterating this change against our internal repo. will merge if all checks out

Issue #794, opened 2018-02-26
Edge case with invalid partial downloads


rberenguel:
Using coursier 1.0.1 under `sbt`.

If a partial download file (say, `.jar.part` somewhere in `v1/cache`) has invalid length (this can be longer than expected or exactly as expected), resolution fails with a 416 error return code: 

```
[error] coursier.ResolutionException: 1 download error
[error]     Caught java.io.IOException: Server returned HTTP response code: 416 for URL: [...]
```

To reproduce this, visit anywhere in your local coursier cache (`~/.coursier/cache/v1/path_to_some_jar`) and move any `file.jar` to `file.jar.part`. Run `sbt update` on any project requiring this `jar` to make it fail.

I suspect the problem is here: https://github.com/coursier/coursier/blob/834fbd598a38d4c8dccbecf47fcd986d213e1495/cache/src/main/scala/coursier/Cache.scala#L572

The logic is expecting for a 206 with a correct range in the header for a retry, but 416 (invalid range) is a valid response in this case, and should be handled separately.

I would like to help with this issue if possible, but don't know how I can get `sbt` to use a local build of coursier for this. Any tips?

rberenguel:
PR https://github.com/coursier/coursier/pull/795 should fix it if I understood how this area works correctly

wisechengyi:
unit tests may be a little tricky @alexarchambault may have better ideas. 
for integration tests you can put it in https://github.com/coursier/coursier/blob/master/tests/jvm/src/test/scala/coursier/test/CacheFetchTests.scala or https://github.com/coursier/coursier/blob/master/cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala. Basically you can mimic the scenario in #795 and show it works.

rberenguel:
Thanks for the pointers @wisechengyi that's what I needed. I'll get it tested ASAP

alexarchambault:
Fixed by https://github.com/coursier/coursier/pull/795.

PR #795, opened 2018-02-26
Resets the connection (and sets partial download to false) if the range is not satisfiable by the server

rberenguel:
Addresses Issue https://github.com/coursier/coursier/issues/794


wisechengyi:
Looks good in general, but could you add a test to it?

rberenguel:
@wisechengyi thanks, will love to, but I’m not sure how/where to test this behaviour. I guess in tests/jvm/src/it/[...] would be the place to do it?

alexarchambault:
About testing that, there's already a small http server running during the integration tests (used [here](https://github.com/coursier/coursier/blob/2747d1e55ab13ec998e1c1006a7546f03252fd17/tests/jvm/src/it/scala/coursier/test/HttpAuthenticationTests.scala#L56-L59) for example). It could be used, along with a cache from a temporary directory, to ensure that the cache download behaves well in that case.

The test can be added under `cache/src/it/scala/…` (which has to be created). You may need to add utest to the cache project, and enable integration tests for it (done for the tests project for example).

alexarchambault:
The test server can be manually launched via
```
$ scripts/launch-test-repo.sh --port 8080 --list-pages
```

rberenguel:
Thanks for the pointers @alexarchambault I'll get this tested ASAP

rberenguel:
@alexarchambault I think I need another pointer, can't figure out a crucial step here. I'm using https://github.com/coursier/coursier/blob/2747d1e55ab13ec998e1c1006a7546f03252fd17/tests/jvm/src/it/scala/coursier/test/HttpAuthenticationTests.scala#L56-L59 as a model (well, I'm not using `check` but a custom method because I need to mess with the cache). But I can't figure out how to get artifacts to appear in the cache temporary directory (I'm basing off `check` here: https://github.com/coursier/coursier/blob/master/tests/jvm/src/test/scala/coursier/test/CacheFetchTests.scala ) After the resolution step only `pom`s, `sha`s and similar appear in the cache folder, even if the mock server contains `jar` artifacts for the module. My impression was that once the Resolution finishes the cache should be populated, but that might not be the case?

alexarchambault:
After the resolution, only the POMs (and their checksums) are fetched. JARs are fetched in a second time (it corresponds to the `Task.gatherUnordered` in the [API example](https://github.com/coursier/coursier#api-1) of the README).

To test your changes, I think you can just create a `.part` file in the cache, corresponding to a file of the http server of the integration tests (like copy a file from `tests/jvm/src/test/resources/test-repo/http/abc.com/com/…` to the cache under `http/localhost%3A8080/com/…` and rename it to a `.part` file). Then you can make an `Artifact` fetching the corresponding URL (`http://localhost:8080/com/…`), and ask `Cache.fetch` or `Cache.file` to fetch it, and see if it succeeds.

alexarchambault:
I'd be ok with merging as is without proper tests, as the cache module itself doesn't have a test suite. We can just add a comment saying loudly what the added condtion handles, and that it can be tested once cache has some proper tests… On my side, I was planning to add some while porting the cache module to nio. Tests for this can be added after that IMO.

rberenguel:
Hi @alexarchambault I'm a bit busy atm, but I'd feel bad merging without the test in place now I have a vague idea of how to do it. Alternatively, it could be merged and I create a new issue (`795 has no tests`, or `cache module has no tests`) and then I work on that issue

wisechengyi:
How about doing so from the cli test suite side? like https://github.com/coursier/coursier/pull/797/files#diff-091ffa792cee5668b5b2ab31820ac6eeR756, but in this case we are not corrupting the jar, but just rename it to xxx.jar.part. and it does not involve setting up a http server.

rberenguel:
Hi @wisechengyi @alexarchambault thanks a lot, indeed. I worked off your PR (to have the full `CliFetchIntegrationTest`) and the following test validates this PR (I changed the junit version to help me while reading logs):

```
  "Wrong range partial artifact resolve" should "succeed with retry" in withTempDir("tmp_dir") {
    dir => {
      def runFetchJunit = {
        val fetchOpt = FetchOptions(common = CommonOptions(mode = "force", cacheOptions = CacheOptions(cache = dir.getAbsolutePath)))
        val fetch = Fetch(fetchOpt, RemainingArgs(Seq("junit:junit:4.6"), Seq()))
        assert(fetch.files0.map(_.getName).toSet
          .equals(Set("junit-4.6.jar")))
        val junitJarPath = fetch.files0.map(_.getAbsolutePath()).filter(_.contains("junit-4.6.jar"))
          .head
        val junitJarFile = new File(junitJarPath)
        junitJarFile
      }

      val originalJunitJar: File = runFetchJunit

      val originalJunitJarContent = FileUtil.readAllBytes(originalJunitJar)

      // Move the jar to partial (but complete) download
      val newJunitJar: File = new File(originalJunitJar.getAbsolutePath + ".part")
      originalJunitJar.renameTo(newJunitJar)

      // Run fetch again and it should pass because of retrying on the partial jar.
      val jar = runFetchJunit
      assert(FileUtil.readAllBytes(jar).sameElements(originalJunitJarContent))
    }
  }
```

rberenguel:
@alexarchambault if PR #799 is merged then I can add my test to the rest of the CLI suite

rberenguel:
@wisechengyi @alexarchambault test added, thanks for all the help to get it :)

alexarchambault:
Thanks @rberenguel!

Issue #780, opened 2018-02-14
bad jar in cache causes subsequent resolve to fail


wisechengyi:
### Problem

1. In some cases, e.g. connection is bad or cut off, the half downloaded jar could end up in the cache
2. If a bad jar is in the cache, it is never evicted. 

Reproducing 1 takes a bit work, but 2 is easy to repro:
```
# Fetch artifact, and it works
$ java -jar dist/coursier-cli.jar fetch junit:junit:4.12
/Users/yic/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar
/Users/yic/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.jar

# Simulate a bad artifact download
$ rm /Users/yic/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar
$ touch /Users/yic/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar

# Fetch again:
$ java -jar dist/coursier-cli.jar fetch junit:junit:4.12
  Error:
https://repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar: WrongChecksum(SHA-1,da39a3ee5e6b4b0d3255bfef95601890afd80709,2973d150c0dc1fefe998f834810d68f278ea58ec,/Users/yic/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar,/Users/yic/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar.sha1)

# And again
$ java -jar dist/coursier-cli.jar fetch junit:junit:4.12
  Error:
https://repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar: WrongChecksum(SHA-1,da39a3ee5e6b4b0d3255bfef95601890afd80709,2973d150c0dc1fefe998f834810d68f278ea58ec,/Users/yic/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar,/Users/yic/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/junit/junit/4.12/junit-4.12.jar.sha1)
```

[...]

PR #797, opened 2018-02-28
Retry on ChecksumError

wisechengyi:
Fix for #780


wisechengyi:
will add some tests

cache/src/main/scala/coursier/Cache.scala
```diff
@@ -946,50 +947,84 @@ object Cache {
...
-        validateChecksum(artifact, c, cache, pool).map(_ => f)
+      res.flatMap {
+        case (file0, None) => EitherT(Task.now[Either[FileError, File]](Right(file0)))
+        case (file0, Some(shaType)) =>
+          validateChecksum(artifact, shaType, cache, pool).map(_ => file0)
+      }.leftFlatMap {
+        case _: FileError.WrongChecksum =>
+          val badFile = localFile(artifact.url, cache, artifact.authentication.map(_.user))
+          badFile.delete()
+          Console.err.println(s"Bad file deleted: ${badFile.getAbsolutePath}")
```
alexarchambault: Use `logger` instead? (a method can be added to it)

cache/src/main/scala/coursier/Cache.scala
```diff
@@ -946,50 +947,84 @@ object Cache {
...
+            cachePolicy,
+            checksums,
+            logger,
+            pool,
+            ttl,
+            retry - 1
+          )
+        case err: FileError.NotFound =>
+          EitherT(Task.now[Either[FileError, File]](Left(err)))
+      }
```
alexarchambault: You should add an extra case `case err =>` to handle the other `FileError` cases.

cache/src/main/scala/coursier/FileError.scala
```diff
@@ -16,6 +16,11 @@ sealed abstract class FileError(
 
 object FileError {
 
+  final case class RetryExhausted(reason: String) extends FileError(
```
alexarchambault: Maybe return the last `FileError` instead of this one? Or keep the last error as a field of that one?

wisechengyi:
looks like I cannot catch `DownloadError` within the scope of `def file`, and it may be managed by `coursier.Cache#downloading`. Hence removing handling `DownloadError` for now.

I was testing with
```
fetch junit:junit:4.12 --no-default -r http://<invalid_url>
```

cache/src/main/scala/coursier/Cache.scala
wisechengyi: Added

cache/src/main/scala/coursier/Cache.scala
wisechengyi: Done

cache/src/main/scala/coursier/FileError.scala
wisechengyi: Used the former and logs "Retry exhausted..." in console.

cli/src/main/scala-2.12/coursier/cli/options/CacheOptions.scala
```diff
@@ -5,7 +5,6 @@ import coursier.Cache
 
 final case class CacheOptions(
   @Help("Cache directory (defaults to environment variable COURSIER_CACHE or ~/.coursier/cache/v1)")
-  @Short("C")
```
wisechengyi: Removing this because conflicting with `--classifier / -C`. I think the annotation processor should throw an error when there are conflicts.

cache/src/main/scala/coursier/TermDisplay.scala
```diff
@@ -509,4 +509,9 @@ class TermDisplay(
     }
   }
 
+  override def log(content: String, currentTimeOpt: Option[Long]): Unit = {
+    updateRunnable.newEntry(content, CheckUpdateInfo(currentTimeOpt, None, isDone = true), content)
+    updateRunnable.removeEntry(content, success = true, fallbackMessage = content)(x => x)
```
wisechengyi: Not entirely sure if this is the right way to log things. Please kindly advise.

wisechengyi:
@alexarchambault this is ready for another look when you get the chance. thanks

cache/src/main/scala/coursier/Cache.scala
```diff
@@ -518,7 +518,7 @@ object Cache {
                       doTouchCheckFile(file)
                       Right(false)
                     }
-                  case other =>
+                  case other =>FileError
```
alexarchambault: ?

cache/src/main/scala/coursier/Cache.scala
```diff
@@ -987,10 +994,35 @@ object Cache {
...
     res.flatMap {
-      case (f, None) => EitherT(Task.now[Either[FileError, File]](Right(f)))
-      case (f, Some(c)) =>
-        validateChecksum(artifact, c, cache, pool).map(_ => f)
+      case (file0, None) => EitherT(Task.now[Either[FileError, File]](Right(file0)))
+      case (file0, Some(shaType)) =>
+        validateChecksum(artifact, shaType, cache, pool).map(_ => file0)
+    }.leftFlatMap {
+      case err: FileError.WrongChecksum =>
+        if (retry == 0) {
```
alexarchambault: `<= 0` just in case?

cache/src/main/scala/coursier/Cache.scala
```diff
@@ -1134,6 +1166,7 @@ object Cache {
     def downloadedArtifact(url: String, success: Boolean): Unit = {}
     def checkingUpdates(url: String, currentTimeOpt: Option[Long]): Unit = {}
     def checkingUpdatesResult(url: String, currentTimeOpt: Option[Long], remoteTimeOpt: Option[Long]): Unit = {}
+    def log(content: String, currentTimeOpt: Option[Long]): Unit = {}
```
alexarchambault: I think it'd be better to have a more specific method for the error we're reporting… like
```scala
def removedCorruptFile(url: String, file: File, reason: Option[FileError]): Unit
```

cache/src/main/scala/coursier/TermDisplay.scala
alexarchambault: Let's keep an empty implem for now (with a more specific method), we'll see how we can report that (with progress bars and all) later…

cli/src/main/scala-2.12/coursier/cli/options/CacheOptions.scala
alexarchambault: I agree. A helper method could be added for that in case-app.

cache/src/main/scala/coursier/Cache.scala
wisechengyi: Reverted. This is weird...CI passed on that.

cache/src/main/scala/coursier/Cache.scala
wisechengyi: Done

cache/src/main/scala/coursier/Cache.scala
wisechengyi: Done

wisechengyi:
plan to merge this tmr

cache/src/main/scala/coursier/Cache.scala
```diff
@@ -975,6 +984,29 @@ object Cache {
...
       case (f, Some(c)) =>
         validateChecksum(artifact, c, cache, pool).map(_ => f)
+    }.leftFlatMap {
+      case err: FileError.WrongChecksum =>
+        if (retry <= 0) {
+          EitherT(S.point(Left(err)))
+        }
+        else {
+          val badFile = localFile(artifact.url, cache, artifact.authentication.map(_.user))
+          badFile.delete()
```
alexarchambault: This call should be wrapped in a `S.schedule(pool) { … }`, as it's doing blocking stuff, like
```scala
S.schedule(pool) {
  badFile.delete()
  logger.foreach(…)
}.flatMap { _ =>
  file(…)
}
```

cache/src/main/scala/coursier/Cache.scala
```diff
@@ -1123,6 +1155,8 @@ object Cache {
 
     def gettingLength(url: String): Unit = {}
     def gettingLengthResult(url: String, length: Option[Long]): Unit = {}
+
+    def removedCorruptFile(url: String, file: File, reason: Option[String]): Unit
```
alexarchambault: Let's keep the full original `FileError` in `reason` instead of a `String`.

cache/src/main/scala/coursier/Cache.scala
wisechengyi: done

cache/src/main/scala/coursier/Cache.scala
wisechengyi: done

PR #799, opened 2018-03-01
Move classifier tests under isActualCentral

wisechengyi:
Nexus tests are already skipped because
```
// Adding extra repo so it's agnostic from nexus which only has the poms		
val extraRepo = MavenRepository("https://repo1.maven.org/maven2")
```
so might as well make them central only.


alexarchambault:
Let me enable again the Nexus proxy tests before having a look at this.

wisechengyi:
np. no hurry on this

alexarchambault:
Back to that, adding `extraRepo` shouldn't be needed. It should work both with the actual Central repo and through a mirror… (no need for `isActualCentral`)

PR #800, opened 2018-03-01
Remove deprecation prior to 1.1.0-SNAPSHOT and clarify versioning and deprecation strategy

wisechengyi:
Version was bumped to 1.1.0-SNAPSHOT and backward compatiliby was removed for up to 1.0.2 at 0123a377e2d8cd19c8887a41122efad14632265c, so it should be safe to remove all the deprecations.


wisechengyi:
Thanks!

PR #806, opened 2018-03-08
Switch to latest directories

alexarchambault:
cc @soc

### DIFF
--- a/3rdparty/jvm/BUILD
+++ b/3rdparty/jvm/BUILD
@@ -138,7 +138,7 @@ jar_library(
         jar(
             org = "io.github.soc",
             name = "directories",
-            rev = "5",
+            rev = "6",
         ),
     ],
 )
--- a/paths/src/main/java/coursier/CoursierPaths.java
+++ b/paths/src/main/java/coursier/CoursierPaths.java
@@ -30,7 +30,7 @@ public final class CoursierPaths {
         if (path == null)
             path = System.getProperty("coursier.cache");
 
-        File baseXdgDir = new File(coursierDirectories.projectCacheDir);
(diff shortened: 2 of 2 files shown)


alexarchambault:
@soc A release of the latest master of directories would be welcome… The pants build here depends on it.

soc:
@alexarchambault which pants build?

I wanted to publish the new version after I have added ScalaDoc to everything, so that existing users have it easier to figure out how they can retain the directory choice they made with the old API under the new one, but it seems at least you have successfully sorted out the required handshakes. :-)

wisechengyi:
@soc There are 2 build systems for coursier, which is SBT, the other is Pants. Pants consumes soc as 3rdparty, rather than a git submodule.
https://github.com/coursier/coursier/blob/master/3rdparty/jvm/BUILD#L135-L144

wisechengyi:
and version 6 isn't on maven looks like https://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22io.github.soc%22%20AND%20a%3A%22directories%22

soc:
Thanks for explaining this!

These are the last changes: https://github.com/soc/directories/commit/2aeb0f89b0faee7968ffe3d11a98b8a9cc5c808c

I also changed the build process to create Java 6 bytecode, for maximal compatibility.

I'm publishing the jar now with version 6, it should be available on Maven shortly.

soc:
It's available on Central now: http://repo1.maven.org/maven2/io/github/soc/directories/6/

Please let me know if there is anything else that needs to be done! Thanks!

wisechengyi:
Thanks @soc !

PR #807, opened 2018-03-10
Remove scalaz dependency from core and cache

alexarchambault:


wisechengyi:
Thanks @alexarchambault. Let me know if Pants side needs any help, although I'd imagine it should be straight forward.

alexarchambault:
@wisechengyi I'll look at the pants build as soon as the other CI jobs are fine. `cache` is now cross-built on the sbt side (scala-js), which changes a few things, but there's already `core` as an example of how to handle it from pants…

alexarchambault:
Gonna go ahead and merge that one.

wisechengyi:
Is there plan to strip scalaz from the rest of the repo?

alexarchambault:
@wisechengyi The goal of having `core` and `cache` not depend on it was to allow users of coursier via its API not to require scalaz (I'm thinking of Ammonite / mill, and maybe sbt too). `cli` and the sbt plugins here still use it, via the `scalaz-interop` module that this PR adds. There's no hurry in removing the scalaz dependency from `cli` and the sbt plugins, although that should be possible now…

wisechengyi:
Removing scalaz from everywhere would be ideal, `scalaz-core_2.12-7.3.0-M20.jar` is quite heavy 8MB. I will create a ticket.

PR #808, opened 2018-03-10
Print: use roots when rendering reverse graph

shanielh:
Allows to print reverse dependencies graph ([Context](https://github.com/lihaoyi/mill/pull/213#issuecomment-372009053))

I think that it was actually a bug, but if you want, I can preserve the old API (which ignored `dependencies` and used `resolution.dependencies`) and add a new method.


wisechengyi:
normal tree:
```
resolve -t junit:junit:4.12 
  Result:
└─ junit:junit:4.12
   └─ org.hamcrest:hamcrest-core:1.3
```
so the reverse tree should be 
```
└─ org.hamcrest:hamcrest-core:1.3
   └─ junit:junit:4.12
```
but this PR has it as:
```
$ resolve -T junit:junit:4.12 
  Result:
└─ junit:junit:4.12
```
missing hamcrest

shanielh:
@wisechengyi Hi, actually you're right, but the usage should be :  

```scala
  private val junit = Dependency(Module("junit", "junit"), "4.12")
  private val hamcrest = Dependency(Module("org.hamcrest", "hamcrest-core"), "1.3")

  val start = Resolution(Set(junit))

  val repositories = Seq(
    Cache.ivy2Local,
    MavenRepository("https://repo1.maven.org/maven2")
  )

  val fetch = Fetch.from(repositories, Cache.fetch())

  val resolution = start.process.run(fetch).unsafePerformSync

  println(Print.dependencyTree(Seq(hamcrest), resolution, false, true))
```

The result is : 

```
└─ org.hamcrest:hamcrest-core:1.3
   └─ junit:junit:4.12
```

I think I'll revert the change and add another method because after the change the CLI "reverse-graph" isn't usable.

shanielh:
@wisechengyi Updated the code, now the example should be : 

```scala
  private val junit = Dependency(Module("junit", "junit"), "4.12")
  private val hamcrest = Dependency(Module("org.hamcrest", "hamcrest-core"), "1.3")

  val start = Resolution(Set(junit))

  val repositories = Seq(
    Cache.ivy2Local,
    MavenRepository("https://repo1.maven.org/maven2")
  )

  val fetch = Fetch.from(repositories, Cache.fetch())

  val resolution = start.process.run(fetch).unsafePerformSync

  println(Print.finalDependencyTree(Seq(hamcrest), resolution, false, true))
```

The result is : 

```
└─ org.hamcrest:hamcrest-core:1.3
   └─ junit:junit:4.12
```

While the `cli` results are : 

```
cli/target/pack/bin/coursier resolve -t junit:junit:4.12 

  Result:
└─ junit:junit:4.12
   └─ org.hamcrest:hamcrest-core:1.3

cli/target/pack/bin/coursier resolve -T junit:junit:4.12

  Result:
├─ junit:junit:4.12
└─ org.hamcrest:hamcrest-core:1.3
   └─ junit:junit:4.12
```

core/shared/src/main/scala/coursier/util/Print.scala
```diff
@@ -80,6 +80,17 @@ object Print {
     reverse: Boolean,
     colors: Boolean
   ): String = {
+    val backwardCompatibleDependencies = if (reverse) resolution.dependencies.toSeq else roots
+    finalDependencyTree(backwardCompatibleDependencies, resolution, printExclusions, reverse, colors)
```
wisechengyi: I am fine with changing the behavior for cli as long as the new behavior is correct, i.e. the root nodes of the printed tree should be the leaf nodes of the resolve, and then there is no need for `finalDependencyTree`.

The reason I am fine with the change is that I don't think anyone is trying to parse the cli tree output, as normally I'd lean toward the deprecation route.

I'd also like to see some tests on this.

Maybe @alexarchambault has more thoughts.

shanielh: I’m not sure but I think that if we want to test it, we should return some data structure of a tree or sequence of trees from a new method (just to keep the API backward compatible) and have another render method that will render it to the CLI. 

Now I’m not sure how to “mock” a resolution with dependencies graph..  

I think that if you want to change the behavior of the CLI then the command should be more like:

“coursier resolve some:module:version -root another:module:version”

This will render a reverse graph from another module version to some module version...

wisechengyi: Since the change is in `core/`, it will affect how cli runs as well.

In cli `resolve -T ..` means printing the reverse tree, so it should print from all leaf nodes all the way to the root nodes`, in this case
```
# no <-root another:module:version> needed here
$ resolve -T junit:junit:4.12
└─ org.hamcrest:hamcrest-core:1.3
   └─ junit:junit:4.12
```

Also I think it would be beneficial to keep the behavior consistent between cli and sbt. If you are only interested to a sub tree from a particular root from sbt, I think it's fine to add separate functionality to it without changing the existing code.

You may be able to look around and see how tests can be added (I am not entirely familiar with this portion yet, so I'd need to that do too), but in general tests would be required for changes in `core/`

shanielh: I’m sorry but I’m not sure how printing the reversed dependencies graph for a single dependency is benificial for someone. Can you give a use case for that?

wisechengyi: No, but I thought you were by proposing `-root another:module:version`

My point is that the CLI output needs to be correct, i.e. starting from all the leaf nodes, and ideally consistent with sbt output.

shanielh: @wisechengyi so can you summarize what you want to happen in every command in the CLI and API and I’ll just do it in the next 24 hours? :)

Thanks.

wisechengyi:
Ok stepping back a bit. 
> I think that it was actually a bug, but if you want, I can preserve the old API (which ignored dependencies and used resolution.dependencies) and add a new method.

Could you elaborate on what the issue is? What is the result you are seeing vs the one you expect to see? The context provided there had nothing to do with reverse graph.

shanielh:
@wisechengyi What I think is : 

 - Changing back the method so it won't be backward compatible, `resolve -T` won't show the library as a root node
 - Add Unit Tests
   - Refactor the methods so the old ones will call two methods
     - Extract Dependency Graph
     - Render
 - Add another command / add "context" parameter to the `coursierDependencyTreeTask` (sbt plugin), so users will able to find what is the root cause of having some dependency (I use that a lot when I want to exclude some dependency to eliminate some logging libraries and adding bridges instead)
 - Add the same command in [mill](https://github.com/lihaoyi/mill/)

wisechengyi:
I am fine with the behavior as is on master, which prints the reverse graph from all transitive dependencies (not perfect, but works). Since it works for CLI, it should work for sbt/mill, albeit still having the issue being a little verbose. Is it something you can work with by looking at the output from the current behavior, or is your use case trying to figure out why a particular dependency is brought in?

If the former, then no change is needed as you should be able to call this function correctly from your build tool with `reverse=true`. If the latter, maybe a separate function would be more appropriate.

Before making any changes, you can outline the behavior here to start with, and we can discuss further.

shanielh:
@wisechengyi My use case is the latter (Trying to figure out why a particular dependency is brought in).

In this case, I think that there is no reason to expose this method in the CLI because for a single dependency, it is simple to follow the graph. So we'll just continue the current branch, I'll make the `Print.finalDependencyTree` method private and add another method (`reverseGraphStartingFrom(roots, resolution, ....)`).

- There will be no changes to the CLI `resolve` operation (even with `-T` option)
- I will extract methods that will create the Tree Structure and render the Tree Structure (for better unit testing)
- Add Unit Tests
- Add SBT `coursierWhatDependsOn` task

wisechengyi:
> There will be no changes to the CLI resolve operation (even with -T option)
> I will extract methods that will create the Tree Structure and render the Tree Structure (for better unit testing)
> Add Unit Tests

Sounds good. Thanks

> Add SBT coursierWhatDependsOn task

This is not my area of expertise since I use neither sbt nor mill, so feel free to ask for additional feedbacks.

shanielh:
@alexarchambault - You brought me here :smile: Would you like me to add the SBT task as well? 
@wisechengyi - I hope to finish the non SBT related stuff tonight (GMT +2)

alexarchambault:
> Would you like me to add the SBT task as well?

@shanielh Go for it :)

I'm not against having that in the CLI too, quite the opposite actually. It could be added via another option for the `resolve` command, like `--tree-for org:name`, to print the tree for that particular dependency.

shanielh:
@wisechengyi I started to extract some methods, what do you think? Would you like me to put the `normalTree` and `reverseTree` methods somewhere else?

shanielh:
@alexarchambault I refactored a little bit of the SBT tasks but I have no knowledge in writing SBT plugins, how can I check if it works?

alexarchambault:
@shanielh You can add a scripted test under `sbt-coursier/src/sbt-test/sbt-coursier`, then run it like
```
$ sbt
> sbt-coursier/scripted sbt-coursier/new-test-name
```
(see the other tests under `sbt-coursier/src/sbt-test/sbt-coursier` as example, `simple` is the most… simple one)

If you want to test outside of scripted (in an actual project), publish the plugin and its dependencies locally
```
$ sbt
> publishLocal
> ~sbt-coursier/publishLocal
```
(first `publishLocal` publishes the dependencies of `sbt-coursier` in particular). Then setup another sbt project from where to test, with `addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.1.0-SNAPSHOT")` in its `project/plugins.sbt`.

shanielh:
@alexarchambault Thanks, I figured it out from `DEVELOPMENT.md` - good to have it :+1: 

Seems like the task works.

alexarchambault:
@shanielh About 1., maybe you can change the type of `coursierWhatDependsOn` to a `InputKey[String]` (or something more refined than `String`), and check its value from the scripted test (the `no-pom-artifact` or `clean` scripted tests illustrate adding a small task to do a few checks).

alexarchambault:
About 2., you shouldn't need to mock `Resolution`. If you add a test in the `tests` module, you should have access to `coursier.test.CentralTests.resolve`, which returns a `Future[Resolution]`. This method relies on things from [this repository](https://github.com/coursier/test-metadata) rather than hitting Central directly. If you re-use a module already fetched from `CentralTests`, it should work fine.

shanielh:
@alexarchambault thanks!

shanielh:
@alexarchambault @wisechengyi Added all the wanted tests :smile: Can you review my PR? (I can squash the commits also if you want)

core/shared/src/main/scala/coursier/util/Print.scala
```diff
@@ -4,6 +4,35 @@ import coursier.core.{ Attributes, Dependency, Module, Orders, Project, Resoluti
...
+    def reconciledVersion: String
+    def children: Seq[Elem]
+  }
+
+  trait Parent extends Renderable {
+    def module: Module
+    def version: String
+    def dependsOn: Module
+    def wantVersion: String
+    def gotVersion: String
```
wisechengyi: Thanks for refactoring these out.

Q: What is the difference between `reconciledVersion` in `Elem` and `gotVersion` in `Parent`?

core/shared/src/main/scala/coursier/util/Print.scala
```diff
@@ -80,32 +109,36 @@ object Print {
...
-    final case class Elem(dep: Dependency, excluded: Boolean) {
+  private def getElemFactory(resolution: Resolution, withExclusions: Boolean): Dependency => Elem = {
+    final case class ElemImpl(dep: Dependency, excluded: Boolean) extends Elem {
 
-      lazy val reconciledVersion = resolution.reconciledVersions
+      val reconciledVersion: String = resolution.reconciledVersions
         .getOrElse(dep.module, dep.version)
 
-      lazy val repr =
+      def repr(colors: Colors): String =
```
wisechengyi: Potentially there could be performance regression changing from `val` to `def` since the value could be access multiple times.

fine with `lazy val` -> `val` tho.

core/shared/src/main/scala/coursier/util/Print.scala
```diff
@@ -146,81 +179,85 @@ object Print {
...
+        else {
+          val assumeCompatibleVersions = compatibleVersions(wantVersion, gotVersion)
 
-      val parents: Map[Module, Seq[Parent]] = {
-        val links = for {
-          dep <- resolution.dependencies.toVector
-          elem <- Elem(dep, excluded = false).children
+          s"$module:$version " +
+            (if (assumeCompatibleVersions) colors.yellow else colors.red) +
+            s"(wants $dependsOn:$wantVersion, got $gotVersion)" +
```
wisechengyi: 1. Could you keep the message consistent with normal tree? I think the `<wanted version> -> <reconciled version>` is good and succinct.

```
$ resolve -t org.apache.avro:avro:1.8.1 org.tukaani:xz:1.6
│  ├─ org.tukaani:xz:1.5 -> 1.6 (possible incompatibility)
```

2. The message appeared at the wrong place. 
```
$ resolve -T org.apache.avro:avro:1.8.1 org.tukaani:xz:1.6
├─ org.tukaani:xz:1.6 // should be here
│  └─ org.apache.avro:avro:1.8.1 (wants org.tukaani:xz:1.5, got 1.6)
```

Would be good to add a test case for this as well.

core/shared/src/main/scala/coursier/util/Print.scala
shanielh: Actually, nothing since I don't have to maintain backward compatibility - I'll change it.

core/shared/src/main/scala/coursier/util/Tree.scala
```diff
@@ -4,7 +4,23 @@ import scala.collection.mutable.ArrayBuffer
...
 object Tree {
 
-  def apply[T](roots: IndexedSeq[T])(children: T => Seq[T], show: T => String): String = {
+  def apply[A](roots: IndexedSeq[A])(children: A => Seq[A], show: A => String): String = {
+    Tree(roots, children).render(show)
+  }
+
+}
+
+case class AppliedTree[A](root: A, children: Seq[AppliedTree[A]])
```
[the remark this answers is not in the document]
shanielh: It's actually for the tests, I can move the `AppliedTree` to the test.

core/shared/src/main/scala/coursier/util/Print.scala
shanielh: Since now the `Tree` and the `Rendering` are two different operations, I had to change it to a method. I'm not sure _how much_ of a performance regression it is, since in the same code path we're doing I/O operations :smile:

wisechengyi: It is unrelated to IO. `def repr` would be the `show` in https://github.com/coursier/coursier/blob/10f3a067fe93cf209e3dfd72f3bf834bdf73666f/core/shared/src/main/scala/coursier/util/Tree.scala#L39

so each time calling `show(elem)` will run the body of `def repr`. Since `elem` can show up in multiple parts of the tree, it would be doing redundant work.

shanielh: I'm aware it's unrelated to IO, just saying that it runs after lots of IO operations, so the performance regression is negligible...

wisechengyi: Sure. But reverting to val should be easy then? Usually we trade performance for something, but I fail to see what's gained here.

core/shared/src/main/scala/coursier/util/Print.scala
wisechengyi: Ok I realized that's the behavior on master. You can leave it as it is if you want. Thanks.

core/shared/src/main/scala/coursier/util/Print.scala
shanielh: Memory, instead of saving all the strings in the tree, we can output them in a streaming manner.

wisechengyi: Sure. I'm open to this change when streaming is implemented.

shanielh: @wisechengyi So, what would you like me to do? pass `Colors` along the tree nodes and change `repr` back to `val`? I'm open to changes, I just want to get this pull request merged so I can pass on to implement the `whatDependsOn` operation in mill :smile:

wisechengyi: Ah okay sorry I missed there's an arg to repr now. Lgtm then. The separation is nice. Thanks

shanielh: :smile: See other comments, Rest is fixed. If you want me to squash all commits I'll do it.

wisechengyi: Thanks. Should be no need to. We will squash in the end merge anyway.

shanielh: @wisechengyi :+1: 

When's the next release? :smile:

wisechengyi: that'll be question for @alexarchambault

alexarchambault:
@shanielh Why, in your example and in the scripted test, does it print `log4j:log4j` two times (once for `1.2.16` and once for `1.2.17`)? I haven't looked too closely at it, but it seems only one of those versions should land in the classpath at the end (and in the resolution result before that)

But ok to merge right now, we can always refine things later if ever it's needed…

alexarchambault:
About including that in a release, I'll probably report it manually to the 1.0 branch, and release `1.0.3` after that.

alexarchambault:
Merging, thanks @shanielh! (and @wisechengyi for reviewing)

shanielh:
@alexarchambault Thanks for merging.

About the `log4j:log4j`, I'll take a look at this today :+1:

PR #823, opened 2018-03-23
Use Java 9 in the 2.12 jobs on Travis CI

alexarchambault:


wisechengyi:
@alexarchambault Java 8 is still live and not deprecated yet, so I think that should be vetted against still. How do you feel about adding a shard with java9?

alexarchambault:
This doesn't switch all the CI jobs to java 9, only the scala 2.12 ones. This seemed like a good compromise to me, in order to:
- have java 9 jobs,
- keep some java 8 ones to ensure it's still ok, and
- not adding more overhead on the CI.

In order to check that some 2.12-only modules (`cli`…) are still fine on the CI, maybe these could also be compiled on 2.11.

alexarchambault:
https://github.com/scala/scala-dev/issues/480 makes that conflict with https://github.com/coursier/coursier/pull/824 for now anyway.

alexarchambault:
That said, errors like https://github.com/scala/bug/issues/10575 are worrying :|

wisechengyi:
Yeah looks like this might need a little more time to mature.

I'd like to be a little more clear on our perspective, cli and core modules have to test with java 8 because we heavily consume the CLI portion on java 8. Java 9 tests are nice to have, but not required at the moment.

wisechengyi:
Also generally speaking, since coursier is a fundamental tool that goes into other build tools, it would be a good idea to be a little more conservative in terms of upgrading and to offer migration path where needed.

alexarchambault:
Closing this for now, that can't work because of https://github.com/scala/scala-dev/issues/480 for now anyway.

Issue #812, opened 2018-03-14
remove scalaz from rest of the codebase


wisechengyi:
https://github.com/coursier/coursier/pull/807#issuecomment-373023671

At least for cli, we can further reduce the binary size.

alexarchambault:
If anyone wants to handle that (cli not depending anymore on scalaz), I won't oppose it. Same for the sbt plugins.

There's no reason to remove the scalaz-interop module though.

alexarchambault:
Closing this, as the cli doesn't depend anymore on scalaz since https://github.com/coursier/coursier/commit/9e5e3cfa1574d06504c2411df2326ee55dbc0555. The scalaz-interop module is still there in case it can be useful, but not used from this repo.

PR #824, opened 2018-03-23
Switch to scala 2.12.5

alexarchambault:


alexarchambault:
🤔 So on Appveyor, `java -XX:+PrintFlagsFinal -version` says
```
     intx CompilerThreadStackSize                   = 0                                   {pd product}
     intx ThreadStackSize                           = 0                                   {pd product}
     intx VMThreadStackSize                         = 0                                   {pd product}
```

alexarchambault:
@wisechengyi Is there a way to have pants get the scala version from… the environement or the command line? That way, its scala version could be bumped via what was added in https://github.com/coursier/coursier/pull/830 too.

wisechengyi:
@alexarchambault not the current version of pants, but the newer pants has an issue with scalaz/shapeless (https://github.com/pantsbuild/pants/issues/5444), hopefully that can be out of the way soon

wisechengyi:
so #812 would help

PR #830, opened 2018-03-31
Only specify scala release versions in one place

alexarchambault:


dwijnand:
I very much like what you've done here, so yes I'd be interested to see if we can generalise it to sbt-travisci.

see for instance https://github.com/dwijnand/sbt-travisci/issues/11

Issue #798, opened 2018-02-28
Local artifact isn't being copied to coursier cache


dotordogh:
Currently, if you resolve a dependency with a local artifact, the artifact will not be copied to the cache. This could be a problem if the artifact only exists temporarily, if it gets deleted, the resolve will fail if it is not in the cache.

alexarchambault:
I confirm that things from `file://` URLs aren't copied to the cache (handled by [this method](https://github.com/coursier/coursier/blob/2747d1e55ab13ec998e1c1006a7546f03252fd17/paths/src/main/java/coursier/CachePath.java#L39)).

Maybe that could be changed… Like we could have a `local://` protocol for things that wouldn't be copied, and we could copy things from `file://` to the cache from now on (a bit like [spark-submit does](https://spark.apache.org/docs/2.3.0/submitting-applications.html#advanced-dependency-management), where things from `file://` are moved around, but those from `local://` aren't).

PR #831, opened 2018-04-01
Copy local artifact to cache

dotordogh:
@alexarchambault, is this what you meant in https://github.com/coursier/coursier/issues/798? I just want to see if I'm on the right track.

### DIFF
--- a/cache/jvm/src/main/scala/coursier/Cache.scala
+++ b/cache/jvm/src/main/scala/coursier/Cache.scala
@@ -37,8 +37,8 @@ object Cache {
   // Check SHA-1 if available, else be fine with no checksum
   val defaultChecksums = Seq(Some("SHA-1"), None)
 
-  def localFile(url: String, cache: File, user: Option[String]): File =
-    CachePath.localFile(url, cache, user.orNull)
+  def localFile(url: String, cache: File, user: Option[String], localArtifactsShouldBeCached: Boolean): File =
+    CachePath.localFile(url, cache, user.orNull, localArtifactsShouldBeCached)
 
   private def readFullyTo(
     in: InputStream,
@@ -363,7 +363,8 @@ object Cache {
     cachePolicy: CachePolicy,
     pool: ExecutorService,
     logger: Option[Logger],
-    ttl: Option[Duration]
+    ttl: Option[Duration],
+    localArtifactsShouldBeCached: Boolean
   )(implicit S: Schedulable[F]): F[Seq[((File, String), Either[FileError, Unit])]] = {
 
     // Reference file - if it exists, and we get not found errors on some URLs, we assume
@@ -371,7 +372,7 @@ object Cache {
     val referenceFileOpt = artifact
       .extra
       .get("metadata")
-      .map(a => localFile(a.url, cache, a.authentication.map(_.user)))
+      .map(a => localFile(a.url, cache, a.authentication.map(_.user), localArtifactsShouldBeCached))
 
     def referenceFileExists: Boolean = referenceFileOpt.exists(_.exists())
 
@@ -779,7 +780,7 @@ object Cache {
       case Some(required) =>
         cachePolicy0 match {
           case CachePolicy.LocalOnly | CachePolicy.LocalUpdateChanging | CachePolicy.LocalUpdate =>
-            val file = localFile(required.url, cache, artifact.authentication.map(_.user))
+            val file = localFile(required.url, cache, artifact.authentication.map(_.user), localArtifactsShouldBeCached)
             localInfo(file, required.url).flatMap {
               case true =>
                 EitherT(S.point[Either[FileError, Unit]](Right(())))
@@ -793,10 +794,10 @@ object Cache {
 
     val tasks =
       for (url <- urls) yield {
-        val file = localFile(url, cache, artifact.authentication.map(_.user))
+        val file = localFile(url, cache, artifact.authentication.map(_.user), localArtifactsShouldBeCached)
 
         def res =
-          if (url.startsWith("file:/")) {
+          if (url.startsWith("file:/") && !localArtifactsShouldBeCached) {
             // for debug purposes, flaky with URL-encoded chars anyway
             // def filtered(s: String) =
             //   s.stripPrefix("file:/").stripPrefix("//").stripSuffix("/")
@@ -879,15 +880,16 @@ object Cache {
     artifact: Artifact,
     sumType: String,
     cache: File,
-    pool: ExecutorService
+    pool: ExecutorService,
+    localArtifactsShouldBeCached: Boolean = false
   )(implicit S: Schedulable[F]): EitherT[F, FileError, Unit] = {
 
-    val localFile0 = localFile(artifact.url, cache, artifact.authentication.map(_.user))
+    val localFile0 = localFile(artifact.url, cache, artifact.authentication.map(_.user), localArtifactsShouldBeCached)
 
     EitherT {
       artifact.checksumUrls.get(sumType) match {
         case Some(sumUrl) =>
-          val sumFile = localFile(sumUrl, cache, artifact.authentication.map(_.user))
+          val sumFile = localFile(sumUrl, cache, artifact.authentication.map(_.user), localArtifactsShouldBeCached)
 
           S.schedule(pool) {
             val sumOpt = parseRawChecksum(Files.readAllBytes(sumFile.toPath))
@@ -941,7 +943,8 @@ object Cache {
     logger: Option[Logger] = None,
     pool: ExecutorService = defaultPool,
     ttl: Option[Duration] = defaultTtl,
-    retry: Int = 1
+    retry: Int = 1,
+    localArtifactsShouldBeCached: Boolean = false
   )(implicit S: Schedulable[F]): EitherT[F, FileError, File] = {
 
     val checksums0 = if (checksums.isEmpty) Seq(None) else checksums
@@ -954,7 +957,8 @@ object Cache {
         cachePolicy,
         pool,
         logger = logger,
-        ttl = ttl
+        ttl = ttl,
+        localArtifactsShouldBeCached
       )) { results =>
         val checksum = checksums0.find {
           case None => true
@@ -982,7 +986,7 @@ object Cache {
     res.flatMap {
       case (f, None) => EitherT(S.point[Either[FileError, File]](Right(f)))
       case (f, Some(c)) =>
-        validateChecksum(artifact, c, cache, pool).map(_ => f)
+        validateChecksum(artifact, c, cache, pool, localArtifactsShouldBeCached).map(_ => f)
     }.leftFlatMap {
       case err: FileError.WrongChecksum =>
         if (retry <= 0) {
@@ -991,7 +995,7 @@ object Cache {
         else {
           EitherT {
             S.schedule[Either[FileError, Unit]](pool) {
-              val badFile = localFile(artifact.url, cache, artifact.authentication.map(_.user))
+              val badFile = localFile(artifact.url, cache, artifact.authentication.map(_.user), localArtifactsShouldBeCached)
               badFile.delete()
               logger.foreach(_.removedCorruptFile(artifact.url, badFile, Some(err)))
               Right(())
--- a/cli/src/main/scala-2.12/coursier/cli/Helper.scala
+++ b/cli/src/main/scala-2.12/coursier/cli/Helper.scala
@@ -295,7 +295,7 @@ class Helper(
         }
     }.toMap
 
-  val depsWithUrlRepo: FallbackDependenciesRepository = FallbackDependenciesRepository(depsWithUrls)
+  val depsWithUrlRepo: FallbackDependenciesRepository = FallbackDependenciesRepository(depsWithUrls, cacheFileArtifacts)
 
   // Prepend FallbackDependenciesRepository to the repository list
   // so that dependencies with URIs are resolved against this repo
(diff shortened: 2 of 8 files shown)


alexarchambault:
@dotordogh Yep, `CachePaths.localFile` should be changed. The code around [here](https://github.com/coursier/coursier/blob/49f4467d6f8668ecc6c3e6a316dc035a63585456/cache/jvm/src/main/scala/coursier/Cache.scala#L799) should be adjusted too (it decides whether a url should be downloaded and written to the destination file, or assumed to be already there).

Also, maybe that feature could be put behind a flag in a first time… Support for `local:/` could be enabled no matter what (as it doesn't break things). And iff the flag is enabled, things from `file:/` would be copied to the cache (else current behavior). That way, in the cli tool, users could be warned if they use `file:/` without the flag, that they should use `local:/` instead, but that wouldn't changed anything else for them for now.

There might be other stuff to change too ([`ivy2HomeUri`](https://github.com/coursier/coursier/blob/49f4467d6f8668ecc6c3e6a316dc035a63585456/cache/jvm/src/main/scala/coursier/Cache.scala#L1093), `file:/` repos from sbt should be changed to `local:/`, …)

dotordogh:
@alexarchambault got it! Thank you!!

cache/jvm/src/main/scala/coursier/Cache.scala
```diff
@@ -796,10 +796,10 @@ object Cache {
         val file = localFile(url, cache, artifact.authentication.map(_.user))
 
         def res =
-          if (url.startsWith("file:/")) {
+          if (url.startsWith("local:/")) {
```
wisechengyi: `local:/` should probably be a constant somewhere in the core, in order to avoid a lot of duplications in the rest of the review as well.

nit: like `http://` being the protocol, `local://` (double /) would follow the convention better.

cli/src/main/scala-2.12/coursier/cli/Helper.scala
```diff
@@ -291,10 +291,27 @@ class Helper(
     .flatMap {
       case (dep, extraParams) =>
         extraParams.get("url").map { url =>
-          dep.moduleVersion -> (new URL(URLDecoder.decode(url, "UTF-8")), true)
+          dep.moduleVersion -> (convertStringToUrl(url), true)
         }
     }.toMap
 
+  private def convertStringToUrl(url: String): URL = {
```
wisechengyi: might need some docstring for this. iiuc this defines the behavior to obtain URLConnection for anything with `local://`

cli/src/main/scala-2.12/coursier/cli/Helper.scala
```diff
@@ -291,10 +291,27 @@ class Helper(
...
+    if (url.startsWith("local"))
+      new URL(null, URLDecoder.decode(url, "UTF-8"), new URLStreamHandler {
+        override def openConnection(u: URL): URLConnection = {
+          val toFileUrl = new URL(url.toString.replace("local", "file"))
```
wisechengyi: bug: `replace` -> `replaceFirst`

cli/src/main/scala-2.12/coursier/cli/Helper.scala
```diff
@@ -291,10 +291,27 @@ class Helper(
...
+          toFileUrl.openConnection()
+        }
+      })
+    else
+      new URL(URLDecoder.decode(url, "UTF-8"))
+  }
+
+  depsWithUrls.foreach { case (moduleVersion: (Module, String), urlBool: (URL, Boolean)) =>
+    if (!cacheFileArtifacts && urlBool._1.toString.startsWith("file:/"))
+      errPrintln(s"Warning: Please use the 'local:/' protocol for ${urlBool._1.toString}")
```
wisechengyi: Example error msg I saw:
```
Warning: Please use the 'local:/' protocol for file:/Users/yic/.m2/repository/junit/junit/4.12/junit-4.12.jar
```
nit: 
* `local:/` -> `local://`
* `file:` -> `file://`
* to make the message a bit more clear, append `to prevent it from being copied into cache.`

cli/src/main/scala-2.12/coursier/cli/options/CommonOptions.scala
```diff
@@ -98,7 +98,13 @@ final case class CommonOptions(
     cacheOptions: CacheOptions = CacheOptions(),
 
   @Help("Retry limit for Checksum error when fetching a file")
-    retryCount: Int = 1
+    retryCount: Int = 1,
+
+  @Help("Specify if artifacts with urls with 'file://' protocol should be cached. " +
+    "Defaults to false, meaning artifacts with urls beginning with 'file://' will " +
+    "not be cached unless this is set to True. In this case, please use the 'local://' protocol.")
```
wisechengyi: > In this case, please use the 'local://' protocol.

It could be a bit more clear. `If the flag is true, please use the 'local://' protocol to prevent artifacts from getting fetched to cache.`

cli/src/main/scala-2.12/coursier/cli/options/CommonOptions.scala
```diff
@@ -98,7 +98,13 @@ final case class CommonOptions(
...
+  @Short("cla")
+    cacheFileArtifacts: Boolean = false
```
wisechengyi: By `cla`, do you mean `cacheLocalArtifacts`?

cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala
```diff
@@ -414,15 +414,15 @@ class CliFetchIntegrationTest extends FlatSpec with CliTestLib {
...
+  "local file dep url" should "have coursier-fetch-test.jar and cached for second run" in withFile() {
     (jsonFile, _) => {
       withFile("tada", "coursier-fetch-test", ".jar") {
         (testFile, _) => {
           val path = testFile.getAbsolutePath
           val encodedUrl = encode("file://" + path, "UTF-8")
 
 
-          val commonOpt = CommonOptions(jsonOutputFile = jsonFile.getPath)
+          val commonOpt = CommonOptions(jsonOutputFile = jsonFile.getPath, cacheFileArtifacts = true)
```
wisechengyi: Given this the first run should put the artifact in cache already, right? so there's no need for second run.

cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala
```diff
@@ -436,10 +436,77 @@ class CliFetchIntegrationTest extends FlatSpec with CliTestLib {
...
+        }
+      }
+    }
+  }
+
+  /**
+   * Result:
+   * |└─ e:f:g
+   */
+  "local dep url" should "have coursier-fetch-test.jar but should fail when file is removed" in withFile() {
```
wisechengyi: This is a great test case!

On the other hand, there should be another case with `a:b:c,url=file://<path>` and `--cla`, so the first run put the file into cache, then remove the file locally, re-resolve should still work because it's in the cache. However it seems to fail currently:

```
$ fetch -t junit:junit:4.12,url=file%3A%2F%2F%2FUsers%2Fme%2F.m2%2Frepository%2Fjunit%2Fjunit%2F4.12%2Fjunit-4.12.jar%20  --no-default  --cla  --json-output-file x.out
  Result:
└─ junit:junit:4.12
/Users/me/Library/Caches/Coursier/v1/file/Users/me/.m2/repository/junit/junit/4.12/junit-4.12.jar

# move the repo, so the file://... isn't valid anymore
$ mv ~/.m2/repository/ ~/.m2/repository2

# this should still succeed by looking at the cache, instead of trying to look at file://...
$ fetch -t junit:junit:4.12,url=file%3A%2F%2F%2FUsers%2Fme%2F.m2%2Frepository%2Fjunit%2Fjunit%2F4.12%2Fjunit-4.12.jar%20  --no-default  --cla  --json-output-file x.out
```

cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala
dotordogh: The second part of this test checks to see that what was fetched was actually from the cache and not the local file.

wisechengyi: Ah I see. sounds good!

cache/jvm/src/main/scala/coursier/Cache.scala
alexarchambault: Not necessarily… `file:/path/to/something` is valid (no "authority" part), and doesn't start with `file://`.

tests/jvm/src/test/scala/coursier/test/CacheFetchTests.scala
```diff
@@ -73,7 +73,13 @@ object CacheFetchTests extends TestSuite {
     * - {
       // test that everything's fine with basic file protocol
       val repoPath = new File(getClass.getResource("/test-repo/http/abc.com").getPath)
-      check(MavenRepository(repoPath.toURI.toString))
+//      check(FallbackDependenciesRepository(Map((Module("com.github.alexarchambault", "coursier_2.11"), "1.0.0-M9-test") -> (repoPath.toURI.toString, true))))
```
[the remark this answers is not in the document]
alexarchambault: @dotordogh It depends on what you want to test… If you just want to test that the warning about `file:/` is printed, I guess tests for that particular point aren't needed…

cli/src/main/scala-2.12/coursier/cli/Helper.scala
alexarchambault: Same as above, `local:/` and `file:/` are ok.

cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala
```diff
@@ -449,6 +516,20 @@ class CliFetchIntegrationTest extends FlatSpec with CliTestLib {
           // open jar and inspect contents
           val fileContents = Source.fromFile(urlInJsonFile).getLines.mkString
           assert(fileContents == "tada")
+
+//          testFile.delete()
+//
+//          assertThrows[Exception]({
```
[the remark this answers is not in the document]
alexarchambault: @dotordogh It depends on what you want to test… If you just want to test that the warning about `file:/` is printed, I guess tests for that particular point aren't needed…

cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala
dotordogh: Alright, so looked into this a bit deeper and strangely, the fetch in the test above doesn't fail, but on the command line it does. I *think* (from what I can tell) the hack of using the [FallbackDependenciesRepository](https://github.com/coursier/coursier/blob/master/extra/src/main/scala/coursier/FallbackDependenciesRepository.scala) is what is causing the problem because it doesn't have caching behavior associated with it. @alexarchambault, could you confirm this?

cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala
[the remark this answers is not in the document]
dotordogh: @alexarchambault I was hoping to test that the fetch would fail because the artifact would be missing after deleting it. Is there a way to do that with how we handle errors?

cli/src/main/scala-2.12/coursier/cli/options/CommonOptions.scala
dotordogh: Maybe `cfa` is better here? I just wanted something shorter than `cacheFileArtifacts` for when I was testing it on the command line. I'm open to suggestions!

wisechengyi: Either works as long as it's consistent :)

cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala
dotordogh: I've resolved this with my most recent commit, but I'm still stuck on trying to assert that a fetch failed without the artifact present. At the moment its causing the tests to exit, and asserting that an exception or error are thrown are not enough to catch the state. Additionally, I'd also love to get the test in `CacheFetchTests` working but I can't figure out how to import the `FallbackDependenciesRepository`. Hopefully @alexarchambault can help with those issues.

cli/src/test/scala-2.12/coursier/cli/CliFetchIntegrationTest.scala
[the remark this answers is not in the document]
alexarchambault: I though something like would work, but the assert seems false…
```
val artifactOptions = ArtifactOptions(force = true)
val commonOpt = CommonOptions(jsonOutputFile = jsonFile.getPath)
val fetchOpt = FetchOptions(common = commonOpt, artifactOptions = artifactOptions)

// fetch with encoded url set to temp jar
val fetch = Fetch(
  fetchOpt,
  RemainingArgs(
    Seq(
      "e:f:g,url=" + encodedUrl
    ),
    Seq()
  )
)

assert(fetch.helper.anyError)
```
The idea would be just to try to avoid the `sys.exit` that makes the test exit.

Overall, the cli module should be reworked so that it doesn't call `sys.exit` too deep in the code this way, but that's how it is for now… :|

alexarchambault:
@wisechengyi That would be fine this way too, yes.

alexarchambault:
@dotordogh The CI failure seems legit.

alexarchambault:
So merging, thanks!

PR #845, opened 2018-04-18
Replace @ symbol with `_at_` in cache directory path

neko-kai:
There are a couple reasons as  to why I'm requesting this change:

  * google Guava ClassPath scanner breaks when project is built with coursier due to '%40' symbols that separate username and hostname in cache path. This bug was fixed in guava 24.1, but it may impact enterprise users unable to upgrade
  * We're leaking part of the credentials outside of user's credentials file
  * `ivy` doesn't do this, so unless there's a specific use case, it's may be safe to drop this feature. Please correct me if I'm wrong.

Alternatively, an alphabetic separator could be used instead of `@`/`%40`, using i.e. Z-encoding

### DIFF
--- a/paths/src/main/java/coursier/CachePath.java
+++ b/paths/src/main/java/coursier/CachePath.java
@@ -69,7 +69,7 @@ public class CachePath {
 
         String userPart = "";
         if (user != null)
-            userPart = user + "@";
+            userPart = user + "_at_";
 
         return new File(cache, escape(protocol + "/" + userPart + remaining));
     }


alexarchambault:
Thanks @kaishh! I'm not sure I want to drop the user part though… It was added in order not to have things downloaded when authenticated, get picked while no authentication is passed to coursier. Unlike Ivy or Maven, coursier tends not to mix things from various repositories together. That way, when running resolutions or getting artifacts with coursier, all the required repositories have to be specified (including an authentication if needed). Only the user part of the credentials is put in the path, so no sensitive stuff should be leaked I think.

The guava issue is unfortunate though…

kaishh:
@alexarchambault 
How about using a wrapper folder instead of prepending username to folder name? Say:
`/v1/https/user--user.name/oss.sonatype.org/`
?

wisechengyi:
@kaishh is it an option for you to customize the cache path, rather changing the default behavior for coursier?

wisechengyi:
@kaishh you can set the cache location with `COURSIER_CACHE` https://github.com/coursier/coursier#how-to-enable-sandboxing

> if other users are bitten by guava bug, they would still have to figure out that cache path is the cause by themselves from scratch.

Assuming you are managing the developer experience, you can either `COURSIER_CACHE` to whatever script they run, or put it in `~/.bash_profile`

kaishh:
@wisenchengyi Thank you, I didn't know that! That wouldn't address the issue though, because inside the COURSIER_CACHE dir there would still be folders named 'username%40www.domain.com'. Also, by users I mean any other user of coursier, not just our company's developers - we've upgraded guava once we traced the issue.

wisechengyi:
I don't have a strong opinion on this, but as a suggestion, how about 
1. Change `"@"` to `"_at_"` to avoid the guava bug, or
2. Put your desired behavior behind a flag and leave the default behavior unchanged.

Provided that @alexarchambault is okay with any of the option.

kaishh:
@wisechengyi Solution 1 would work, except there exists a possibility of collision if there exists a site with a domain like `user_at_site.com`, but it's not a realistic worry.

alexarchambault:
What do you think of handling that via a Java property? (e.g. `coursier.cache.user.separator`) It's only a workaround for a guava issue, that was fixed in guava. Only a handful of users are going to adjust that, no need to pass a parameter around in the API for it IMO…

kaishh:
@alexarchambault I think guava bug is just a specific case of a problem with using special characters in filepaths as they can break some tools/platforms. Even if separator is set to an ascii sequence, special characters in username will exhibit the same bug. (But not special characters in domain name, because IDNs are already as encoded as punycode).

If solving the general issue for all users (by e.g. using punycode to escape '@' and unicode in username) is undesirable, then we should just go with java property and close this.

alexarchambault:
@kaishh My current preference would be to simply allow the `@` to be overridden via a Java property. If the property is unset, that has the advantage of not changing the paths in cache for these artifacts, so not requiring users to download that again. Plus there are other characters that may be translated to `%…` (see around [here](https://github.com/coursier/coursier/blob/f003314317c285f8a296c2b5879e1db0c351a3cc/paths/src/main/java/coursier/CachePath.java#L20)), so the current state of the PR wouldn't fully solve that issue.

I don't know what are other people's take on that issue… (`%` in paths in cache) I'm still unsure `%` in paths are that a problem…

wisechengyi:
> I'm still unsure % in paths are that a problem…

sort of. it needs some handling if the consumer was not careful with it. E.g. https://github.com/pantsbuild/pants/pull/5301

> allow the @ to be overridden via a Java property

Reasonable to me. I like to the idea to keep the existing behavior.

pshirshov:
Hi guys. 

I've just noticed that this breaks `classgraph` on JDK 10: https://github.com/classgraph/classgraph/issues/255

Could you fix that? Pleeeease?

pshirshov:
Guys, instead of using username you may just take a prefixed hash. Like `user-${mummur128(creds)`

lukehutch:
I added code to work around this problem in ClassGraph-4.2.6. To match a non-%-encoded path string to a path structure that contains a %-encoded directory name, whenever ClassGraph gets a FileNotFoundException, it has to break the path string into '/'-delimited segments, and then trace each segment from the filesystem root, manually matching segments against nested subdirectory names. At any stage, if a nested subdirectory doesn't exist, it has to find all subdirectory and file names within the parent directory at the current level, then iterate through them to see if any are %-encoded, then %-decode those, and match them against the current path segment, to see if there is a match, and if so, proceed to the next-deepest subdirectory.

This works but it's ugly, and it was a pain to have to work around this in ClassGraph because of this one rather obscure situation.

I wanted to point out that %-encoding is *only* intended for URLs, not for filenames. In particular, using the % character in a file or directory name in Windows is a very bad idea, because % is used for delimiting environment variables.

I think the idea to hash user ids is a good one. I'll leave the fix in ClassGraph though for users of existing and older versions of Corsier.

chadselph:
I would prefer some way to just turn off user-prefixing entirely (e.g. `coursier.cache.user.pathinclude`) so that bootstraps could work with auth'ed repositories https://github.com/coursier/coursier/issues/1774

PR #846, opened 2018-04-18
Add a Task.tailRecM for stack safe loops

johnynek:
this is useful for anyone writing loops with Task (or implementing a cats Monad for Task).

### DIFF
--- a/cache/shared/src/main/scala/coursier/util/Task.scala
+++ b/cache/shared/src/main/scala/coursier/util/Task.scala
@@ -31,14 +31,14 @@ object Task extends PlatformTask {
 
   def tailRecM[A, B](a: A)(fn: A => Task[Either[A, B]]): Task[B] =
     Task[B] { implicit ec =>
-      // this loop is safe, because
-      // Future.flatMap is stack-safe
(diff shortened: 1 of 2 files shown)


johnynek:
actually, you are right, the flatMap of Task is stack safe. In general, the Reader flatMap is not stack safe, and your Task is basically `ReaderT[Future, ExecutionContext, A]`. But in the case where the `F[_]` in reader is stack safe, it is okay, as is the case for you.

Indeed, my code would only work in that case, so I think there is no improvement.

alexarchambault:
@johnynek The naive `tailRecM` could be added though, so the test can be kept to ensure `coursier.util.Task` remains stacksafe.

The scalajs jobs on the CI are failing because of the `Await` in the test. But it's not necessary: utest is fine if the test returns a `Future` (it will just look into it, a `countTo(500000).map(n => assert(n == 500000)).future()` should work fine…)

johnynek:
okay, I updated, but it looks like one of the tests flaked.

alexarchambault:
Merging, thanks!

PR #849, opened 2018-04-23
Upgrade to Scala Native 0.3.7

densh:
This PR uses new Build API introduced in 0.3.7 to remove dependency on the Scala Native internals for `--native`.

/cc @olafurpg


alexarchambault:
Thanks @densh!

Issue #848, opened 2018-04-22
Bootstrapped executable does not accept system property changing arguments


2m:
After bootstrapping a project, the generated executable does not correctly pass in the `-Dkey=value` property changing arguments.

The problem, is that `$@` is used after the `-jar` here:

https://github.com/coursier/coursier/blob/05049a235d238020d9d5a0bbddbd6874b1cad2d7/cli/src/main/scala-2.12/coursier/cli/Bootstrap.scala#L203

and the `-D...` arguments are not picked up.

IMO the arguments from `$@` should be partinioned by whether they start with `-D` or not, and the former ones should go before `-jar`, and the latter ones after.

alexarchambault:
There's another bug along with that one: arguments like `--property foo=bar` are ignored because [this variable](https://github.com/coursier/coursier/blob/05049a235d238020d9d5a0bbddbd6874b1cad2d7/cli/src/main/scala-2.12/coursier/cli/Bootstrap.scala#L78) is trapped.

2m:
Looks like #850 fixed the `--property foo=bar` passed in to the bootstrap command, but not the `-Dkey=value` passed to the generated binary. So this ticket is still valid.

PR #850, opened 2018-04-29
Allow to generate assemblies via the bootstrap command

alexarchambault:


alexarchambault:
Fixes https://github.com/coursier/coursier/issues/848.

PR #852, opened 2018-05-03
Partition system and app args to bootstrapped executable

### LINKED ISSUE #848: Bootstrapped executable does not accept system property changing arguments

2m:
Fixes #848


2m:
I finally got all of the bashisms out which fixed all of the previous errors from the bash script when running on Travis. However now CI failed with mismatched output from the checks in `travis.sh`. It looks as if the cli is being used not from this PR. And I am not able to reproduce this on my machine:

```
➜  ./cs-echo-launcher-args -Dother=thing foo -Dfoo=baz
foo
```

alexarchambault:
It seemed to work locally for me too…

2m:
I am not too happy with the solution in my PR as well. It is a bit too magical.

An alternativce approach would be a `-J` flag which, for example, sbt launcher uses:
```
-J-X               pass option -X directly to the java runtime
```

So then the bootstrap generated script could partition the arguments if they start with `-J` and pass in the ones starting with `-J` before `-jar` with `-J` removed.

2m:
I gave this another try and this time implemented the `-J` support like sbt does.

2m:
Thanks for the https://github.com/coursier/coursier/pull/852/commits/d4c57d5432c6ee951367569689257a14b2a4d6ab

It looks like the last implementation of `-J` works even for the older `bash` and `sh` that are in Travis CI environment. So this looks ready to be merged in.

rfkm:
This breaks apps which accept quoted arguments:

Expected:
```sh
$ your-app-printing-args a b "c d"
# => List("a", "b", "c d")
```

Actual:
```sh
$ your-app-printing-args a b "c d"
# => List("a", "b", "c", "d")
```

rfkm:
Yeah, it's not an easy task to process args without arrays.
Another solution with `sh` might be iterating arguments twice...?

```sh
#!/bin/sh

nargs=$#

i=1; while [ "$i" -le $nargs ]; do
         eval arg=\${$i}
         case $arg in
             -J-*) set -- "$@" "${arg#-J}" ;;
         esac
         i=$((i + 1))
     done

set -- "$@" -jar "$0"

i=1; while [ "$i" -le $nargs ]; do
         eval arg=\${$i}
         case $arg in
             -J-*) ;;
             *) set -- "$@" "$arg" ;;
         esac
         i=$((i + 1))
     done

shift "$nargs"

java "$@"
```

2m:
Woa. That looks great! Did not know that you can change the arguments with `set` like that. :)

The only modification this will need is to incorporate the javaOpts that could be passed to the `bootstrap` command when generating the bootstrap launcher.

https://github.com/coursier/coursier/blob/13582d6ee6571f0712e7c5a9987f654b9c283fc1/cli/src/main/scala-2.12/coursier/cli/Bootstrap.scala#L88-L91

rfkm:
Thank you for checking! I'll try creating PR later.

PR #868, opened 2018-06-09
add global maven2Local cache shortcut for ease of use

mpollmeier:
When using the coursier API I was looking for a constant that links to
maven local, just like sbt's `Resolver.mavenLocal`. Didn't find it, so here it is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,6 @@ import coursier.util.Task
 
 val repositories = Seq(
   Cache.ivy2Local,
-  Cache.maven2Local,
   MavenRepository("https://repo1.maven.org/maven2")
 )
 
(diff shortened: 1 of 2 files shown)


alexarchambault:
Thanks @mpollmeier. The addition in the README makes the CI fail, maybe just don't touch the README.

Beware that repo isn't guaranteed to always work fine with coursier (it sometimes has only the metadata of some dependencies, and coursier isn't fine with that - coursier requires both the metadata and the JARs to be in the same repo). Maybe this could be added in an object instead, like
```scala
object Dangerous {
  lazy val maven2Local = …
}
```
with a small comment or some scaladoc mentioning the issue with it.

mpollmeier:
> coursier requires both the metadata and the JARs to be in the same repo

Good to know. All done and green now.

alexarchambault:
Merging, thanks!

Issue #867, opened 2018-06-05
Resolution of `parent.version` in pom.xml


lgirault:
A maven dependency containing 
```
<dependency>
    <groupId>io.swagger.parser.v3</groupId>
   <artifactId>swagger-parser-core</artifactId>
  <version>${parent.version}</version>
</dependency>
```
provoke the following error
```[error] coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
[error] io.swagger.parser.v3:swagger-parser-core:${parent.version}:
[error] not found:
[error] /home/lorilan/.ivy2/local/io.swagger.parser.v3/swagger-parser-core/${parent.version}/ivys/ivy.xml
[error] https://repo1.maven.org/maven2/io/swagger/parser/v3/swagger-parser-core/$%7Bparent.version%7D/swagger-parser-core-$%7Bparent.version%7D.pom
[error] (coursierResolutions) coursier.ResolutionException: Encountered 1 error(s) in dependency resolution:
[error] io.swagger.parser.v3:swagger-parser-core:${parent.version}:
[error] not found:
[error] /home/lorilan/.ivy2/local/io.swagger.parser.v3/swagger-parser-core/${parent.version}/ivys/ivy.xml
[error] https://repo1.maven.org/maven2/io/swagger/parser/v3/swagger-parser-core/$%7Bparent.version%7D/swagger-parser-core-$%7Bparent.version%7D.pom
```
It can be reproduced with an empty project containing a build.sbt with only :
`libraryDependencies += "io.swagger.parser.v3" % "swagger-parser-v3" % "2.0.1"`
I'm using sbt `1.1.6` and coursier `1.1.0-M4`

PR #870, opened 2018-06-13
Add parent.X substitution

lgirault:
Fixes coursier/coursier/issues/867

### DIFF
--- a/tests/shared/src/test/scala/coursier/test/CentralTests.scala
+++ b/tests/shared/src/test/scala/coursier/test/CentralTests.scala
@@ -790,6 +790,13 @@ abstract class CentralTests extends TestSuite {
       )
     }
 
+    'parentVersionInPom - {
+      * - resolutionCheck(
+          Module("io.swagger.parser.v3", "swagger-parser-v3"),
+          "2.0.1"
+        )
+    }
+
     'parentBeforeImports - {
       * - resolutionCheck(
         Module("org.kie", "kie-api"),
--- a/tests/shared/src/test/scala/coursier/test/CentralTests.scala
+++ b/tests/shared/src/test/scala/coursier/test/CentralTests.scala
@@ -791,14 +791,14 @@ abstract class CentralTests extends TestSuite {
     }
 
     'parentVersionInPom - {
-      * - resolutionCheck(
+      resolutionCheck(
           Module("io.swagger.parser.v3", "swagger-parser-v3"),
           "2.0.1"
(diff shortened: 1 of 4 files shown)


alexarchambault:
Thanks @lgirault. Maybe a non regression test could be added for that, with the failing dependency of https://github.com/coursier/coursier/issues/867. A simple test like [this one](https://github.com/coursier/coursier/blob/8a02ee8f15261880d53540f2a3dddbdbae3c2e23/tests/shared/src/test/scala/coursier/test/CentralTests.scala#L786-L791) can be added. In order to get the metadata for the test, one should set the `FILL_CHUNKS=1` in the environment while running the test, like `FILL_CHUNKS=1 sbt testsJVM/test`. This should add files under `tests/shared/src/test/resources/resolutions` and `tests/metadata`. Those of `tests/metadata` are in a git submodule, so should be pushed to the [metadata repo](https://github.com/coursier/test-metadata) (open a PR there). It's not too straightforward overall, so I can look into it if you don't feel like handling that.

lgirault:
Thank you very much for the feedback and the explanation. I think I've done what you told me but I'm not confortable with git submodule. Do I need to commit the changes performed on `tests/metadata` ? (I mean  the "symlink/ref" toward this [commit](coursier/test-metadata/pull/2/commits/d86c1e53bf6fa2fb0600d2bc949412cdf130492a) ? )

alexarchambault:
About the submodule, a `git status` in the `coursier` directory should show `tests/metadata` as modified. Now that https://github.com/coursier/test-metadata/pull/2 is merged, from the `tests/metadata` dir, you should be able to checkout the latest master. It can then be added with a `git add tests/metadata` from the `coursier` dir, and committed along with the rest.

tests/shared/src/test/scala/coursier/test/CentralTests.scala
```diff
@@ -790,6 +790,13 @@ abstract class CentralTests extends TestSuite {
       )
     }
 
+    'parentVersionInPom - {
+      * - resolutionCheck(
```
alexarchambault: (nitpicking) the `* - ` isn't necessary here, like in `groupIdVersionProperties` above. It just adds an extra level in the utest output (I see that `parentBeforeImports` does that below… but it shouldn't…).

alexarchambault:
Merging, thanks!

lgirault:
sweet ! thank you !

PR #872, opened 2018-06-16
Pluggable loggers for sbt-coursier

### LINKED ISSUE #363: Provide an option to disable the Logger in the sbt plugin
We use coursier in a build pipeline where artifacts are build from scratch. It would be awesome to be able to turn of logging of download processes for artifacts by providing either a silent Logger trait or by being able to configure the logger.

teozkr:
This change allows you to set `coursierLoggerFactory` to any `Cache.Logger` you wish. I'm planning to use this for [sbt-lockfile](https://gitlab.com/teozkr/Sbtix/issues/41), but this also fixes #363 (by letting you set it to a no-op logger).

### DIFF
--- a/cache/jvm/src/main/scala/coursier/Cache.scala
+++ b/cache/jvm/src/main/scala/coursier/Cache.scala
@@ -1165,7 +1165,7 @@ object Cache {
     def gettingLength(url: String): Unit = {}
     def gettingLengthResult(url: String, length: Option[Long]): Unit = {}
 
-    def removedCorruptFile(url: String, file: File, reason: Option[FileError]): Unit
+    def removedCorruptFile(url: String, file: File, reason: Option[FileError]): Unit = {}
 
     /***
      *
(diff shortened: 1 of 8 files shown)


alexarchambault:
Thanks @teozkr. I wonder if this could be tested via a [scripted test](https://github.com/coursier/coursier/blob/8a02ee8f15261880d53540f2a3dddbdbae3c2e23/DEVELOPMENT.md#run-a-scripted-test-of-sbt-coursier-or-sbt-shading) (those are under [this directory](https://github.com/coursier/coursier/tree/8a02ee8f15261880d53540f2a3dddbdbae3c2e23/sbt-coursier/src/sbt-test/sbt-coursier)). The test could set up its own logger, run an update, and check that a few methods on the logger are actually called (`init`, …). Is it something you would feel like adding? I can look into it else…

alexarchambault:
Merging, thanks!

teozkr:
Thanks! Any idea when the next milestone is happening?

PR #876, opened 2018-06-21
Upgrade Pants to 1.8.0rc0

wisechengyi:


alexarchambault:
I tried some time ago (see [this branch](https://github.com/coursier/coursier/tree/topic/scalafmt)), but I couldn't find ways to handle fine some details… IIRC, the issue was that things like
```scala
def foo(
  n: Int
)(implicit
  a: A,
  b: B
)
```
and
```scala
foo
  .something
  .other
  .map(_ + 1)
```
were changed to
```scala
def foo(
  n: Int
)(
  implicit a: A,
  b: B
)
```
and
```scala
foo.something.other.map(_ + 1)
```

The latter cas might be fine with the newest scalafmt. Don't know about the former.

wisechengyi:
@alexarchambault Yeah there's probably a config somewhere.

Although I tend not to engage in any holy war on which exact style to adopt, as long as there is a consistent style, but I will definitely try to accommodate the needs.

Also heads up that if we are using pants to format/enforce the style, that means for the code you care about, they need to be in a pants target.

PR #877, opened 2018-06-22
Introduce Scalafmt via Pants

wisechengyi:
This introduces scalafmt as an automated formatter and linter via Pants.

### To format:
```
./pants fmt ::
```

### To lint (which would run in CI):
```
./pants lint ::
```

Example changes are visible at: https://github.com/coursier/coursier/compare/master...wisechengyi:eg?expand=1

A separate PR would be made once this gets in to format all targets covered by Pants and to enforce it via CI.

### DIFF
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,0 +1,20 @@
+maxColumn = 100
+
+docstrings = JavaDoc
+
+assumeStandardLibraryStripMargin = false
+
+align = none
+
+continuationIndent.callSite = 2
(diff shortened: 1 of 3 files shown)


alexarchambault:
@wisechengyi I'm not sure I want to add this right now, as the sources of coursier aren't formatted by scalafmt for now… But it sure should be merged as soon as it's added.

wisechengyi:
This PR has no effect until pants lint is used in Travis. But yeah, whenever you feel comfortable.

PR #880, opened 2018-07-06
Make coursierLoggerFactory a Task rather than a Setting

teozkr:
Not sure what I was thinking there..


teozkr:
I ran `sbt scripted` locally, and none of the failures seemed to have anything to do with this change.

teozkr:
I don't like `coursierLogger` since it sounds like it just creates a `Cache.Logger` rather than a `() => Cache.Logger` (arguably it would be good to change that... but that seems like it would require bigger changes).

alexarchambault:
Merging, thanks!

alexarchambault:
Going to cut `1.1.0-M5` later today.

PR #882, opened 2018-07-16
Make bootstrap safe for arguments including spaces

rfkm:
As we discussed in #852 (thanks, @2m), bootstrap scripts generated by the latest version of  coursier can't correctly handle arguments including spaces.
This should fix the issue by using `set` command to deal with arguments, though there may be a smarter way.


alexarchambault:
Merging, thanks @rfkm and @2m!

alexarchambault:
I should release `1.1.0-M6` in a few hours…

Issue #889, opened 2018-08-04
Resolution fails with an XML parsing error on a HTML redirect page


przemek-pokrywka:
Whenever I try to fetch some artifact from my nexus using:
`fetch --checksum none --no-default --repository http://nexus.redacted.com com.redacted:redacted_2.11:1.0.0` the nexus serves a HTTP redirect with the following HTML page:
```html
<html>
<head><title>301 Moved Permanently</title></head>
<body bgcolor="white">
<center><h1>301 Moved Permanently</h1></center>
<hr><center>nginx/1.12.1</center>
</body>
</html>
```
which confuses coursier so that it fails with:
```
Error:
  com.redacted:redacted_2.11:1.0.0
    No fallback URL found
    org.xml.sax.SAXParseException; lineNumber: 6; columnNumber: 3; The element type "hr" must be terminated by the matching end-tag "</hr>". (The element type "hr" must be terminated by the matching end-tag "</hr>".)
```

What would be needed for coursier to support redirects?

przemek-pokrywka:
Update: I was able to solve the issue by specifying the correct repository address (with https prefix and with a proper path - https://nexus.redacted.com/repository/maven).

So what's left is possibility to have a more helpful error message in case one misspecifies the repo address (or when the repo breaks and serves HTML instead).

I'd propose to print out the complete URL along with the HTML received.

alexarchambault:
I think some people already run into http -> https redirect issues, that should not have been addressed already… It seems you're running into another case of that (with the redirection to https not happenning, and the xml parser failing on the redirect page).

alexarchambault:
Should be fixed by https://github.com/coursier/coursier/pull/976.

lefou:
The solution in #976 is not quite fixing the parsing error when the HTML document is not proper XML. Instead it fixed the concrete error case, in which a redirection was not possible.

To date, coursier can't handle repository pages, which don't contain proper XHTML, e.g. no closing tags.

PR #890, opened 2018-08-05
improvement: print out the content that failed to parse as XML

przemek-pokrywka:
Having coursier print out the whole content of the invalid XML can be helpful in various troubleshooting situations. For example it already allowed me to figure out, that my repository address was configured incorrectly (with http instead of https, which made the nexus send a HTML redirect page) - more details at https://github.com/coursier/coursier/issues/889

Ideally the error message would additionally include the original URL, but I don't know coursier's codebase well enough to do it quickly and the full string content of invalid XML should let the users troubleshoot more easily.


alexarchambault:
IIRC, there's an implicit assumption that these messages should be one-line… (so that the errors coursier later prints are readable). It'd be interesting to have the path to the file in cache that fails to parse though, which may require to tweak some of the APIs around.

Issue #896, opened 2018-08-14
Stop relying on directory listings


alexarchambault:
[will add details later] These were read to circumvent buggy packaging info in some POMs, but turned out not fully addressing the issue at the end. The initial issue was ultimately fixed by returning "optional" artifacts, that may throw a not found error when attempting to download them (which should be ignored).

wisechengyi:
Interested in the how this affect the result of the resolve.

On a operational perspective, this would be great, because we use a virtual repo which is an aggregate of multiple repos, so doing directory listing is super expensive due to lacking of persistent caching, which can take up to 5 sec per request without cache, and extending the caching period for directory listing would affect the update propagation. On the other hand, hitting an individual file is always cached a lot faster.

PR #899, opened 2018-09-19
Don't rely on directory listings anymore

alexarchambault:
This should save some bandwidth, and avoid round-trips with repositories.

Fixes https://github.com/coursier/coursier/issues/896.


alexarchambault:
(will probably split the second commit a bit for readibility…)

wisechengyi:
Nice! Did some initial test, it's a lot of faster without having to do directory listings!

PR #902, opened 2018-09-20
Rework attributes handling

alexarchambault:


alexarchambault:
I think some changes here change the json report (`type` and `classifier` aren't fields of `Artifact` any more)... I guess I'll add some non regression tests for it, and ensure nothing changes.

The classes that are used to decode Scaladex requests changed too (because of organization and name, that now have their dedicated case classes). I guess some basic tests for that could be added.

PR #911, opened 2018-09-27
Add website (WIP)

alexarchambault:


doc/website/siteConfig.js
```diff
@@ -0,0 +1,43 @@
...
+  highlight: {
+    theme: 'default',
+  },
+
+  scripts: ['https://buttons.github.io/buttons.js'],
+
+  onPageNav: 'separate',
+  cleanUrl: true,
+
+  enableUpdateTime: true, // doesn't seem to work
```
olafurpg: I also had problems with this option, I think it was a recent contribution

alexarchambault: They seem to [use it](https://github.com/facebook/Docusaurus/blob/7d4d9fe96127c96aa77a1d57a138174ed617db7c/v1/website/siteConfig.js#L72) successuflly on their own website though.

alexarchambault: Seems that's because they expect the `.md` files to be committed, see [here](https://github.com/facebook/Docusaurus/blob/0f8535293eefddd41dc59e6ad622f6fcac1ec7b8/v1/lib/core/DocsLayout.js#L50).

olafurpg: Makes sense, although it might be nice to fall-back to the file timestamp.

alexarchambault: Agreed...

PR #915, opened 2018-09-28
Remove sbt plugins from main repo

alexarchambault:
Checking how much faster the CI jobs would get…


alexarchambault:
Will probably go ahead and move the sbt plugins to their own repo…

alexarchambault:
Plugins should now live there: https://github.com/coursier/sbt-coursier.

PR #916, opened 2018-09-28
fix build definition to compile on JDK 11

SethTisue:
caught by https://scala-ci.typesafe.com/view/scala-2.12.x/job/scala-2.12.x-jdk11-integrate-community-build/34/


alexarchambault:
Thanks @SethTisue! I'll probably try to add a JDK 11 CI job here at some point, to catch those.

PR #931, opened 2018-10-08
 fix Potential NPE in Cache.handlerFor - this closes #898

### LINKED ISSUE #898: Potential NPE in Cache.handlerFor
The code below crashes with an NPE if the classloader is null, i.e. when we're dealing with the bootstrap classloader. This happens when the corresponding classes are loaded from the boot classpath.

https://github.com/coursier/coursier/blob/2177330c02ce5703d0c85f9f6e197fe3b4d94fce/cache/jvm/src/main/scala/coursier/Cache.scala#L208-L213

dorsev:
another attempt at #898 according to discussion at 
https://github.com/coursier/coursier/pull/929
@alexarchambault


dorsev:
closing and opening a proper PR without changing existing commits

PR #939, opened 2018-10-16
Mention `brew tap` before `brew install` in docs

vrinek:
When I tried it without `tap`, I got:
```
Error: No available formula with the name "coursier/formulas/coursier"
Please tap it and then try again: brew tap coursier/formulas
```


alexarchambault:
Thanks!

Issue #233, opened 2016-04-19
Classpath ordering


benmccann:
Coursier causes compilation to fail for me on some projects. I've created a sample project that reproduces this error: https://github.com/benmccann/coursier-test-case

alexarchambault:
The error seems to originate from the class `javax.servlet.http.HttpServletRequest` being present in both `org.apache.tomcat:tomcat-servlet-api:8.0.32` and `org.mortbay.jetty:servlet-api:2.5-20081211`. coursier hasn't paid attention to the order of the JARs returned until now, but it [may](http://stackoverflow.com/questions/793054/maven-classpath-order-issues/793193#793193) [need to](http://stackoverflow.com/questions/12624892/maven-dependency-order).

As a workaround in the mean time, adding to `build.sbt`
```scala
excludeDependencies += SbtExclusionRule("org.mortbay.jetty")
```
fixes the problem for me (`run` is fine).

alexarchambault:
Closing this. coursier is now able to sort dependencies in classpath order (dependencies before dependees, in a deterministic way). It should be enabled in the upcoming sbt-coursier version (`2.0.0-RC4`), and likely soon after that in the coursier support of sbt 1.3.

PR #946, opened 2018-10-26
Sort artifacts by the order they should be put in the classpath (WIP)

alexarchambault:
Fixes https://github.com/coursier/coursier/issues/233.

This is one of the few issues I'd like to sort out prior to coursier possibly becoming the default dependency resolver in sbt.


alexarchambault:
Now at https://github.com/coursier/coursier/pull/1296.

PR #951, opened 2018-10-30
Ensure that bootstrap task is idempotent (Fixes #950)

### LINKED ISSUE #950: Bootstrap command is not deterministic
Nixos check sha256 during it's package build.
It happens that running coursier bootstrap twice in a row yield 2 different sha256.

Tomahna:
Force time on ZipEntries to Jan 1st 1970
Remove use of java.util.Properties


alexarchambault:
Merging, thanks again @Tomahna!

PR #952, opened 2018-10-31
[WIP] [Fixes #718] Fix version parsing when merging several versions

### LINKED ISSUE #718: Semantic versions are not supported
I'm seeing these errors when I use a version with meta information (like `v1.0.0+fast`) for a dependency:

```
Ignoring unparsed versions: Vector(v1.0.0+fast)
Ignoring unparsed versions: Vector(v1.0.0+fast)
Ignoring unparsed versions: Vector(v1.0.0+fast)
```

But the dependency is resolved anyway.

I've found the line printing the message:
https://github.com/coursier/coursier/blob/b21042a42a6b2718ec1116f28f76dd769112c6aa/core/shared/src/main/scala/coursier/core/Resolution.scala#L174

But now I need to understand why the paragraph 10 of [SemVer](https://semver.org) is not supported or what is the problem.

guilgaly:
WIP - new tests not written yet

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Parse.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Parse.scala
@@ -79,13 +79,13 @@ object Parse {
       None
   }
 
-  def versionConstraint(s: String): Option[VersionConstraint] = {
+  def versionConstraint(s: String): VersionConstraint = {
     def noConstraint = if (s.isEmpty) Some(VersionConstraint.all) else None
 
     noConstraint
       .orElse(ivyLatestSubRevisionInterval(s).map(VersionConstraint.interval))
-      .orElse(version(s).map(VersionConstraint.preferred))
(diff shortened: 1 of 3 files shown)


modules/tests/shared/src/test/scala/coursier/test/ResolutionTests.scala
```diff
@@ -676,6 +697,27 @@ object ResolutionTests extends TestSuite {
         assert(res == expected)
       }
     }
+
+    'mergingTransitiveDeps{
```
guilgaly: This test did fail when tested without the fix (`my-lib-1` would apparently be resolved _without_ a version number).

alexarchambault: Ouch…

alexarchambault:
(pushed an extra commit to trigger appveyor)

alexarchambault:
Merging, thanks @guilgaly!

PR #958, opened 2018-11-15
Added configuration properties for http timeouts

aaronp:
We've experienced some long-running builds due to our nexus configuration, so it'd be a real boon to be able to configure the default http client, e.g. by setting these time-outs.

### DIFF
--- a/modules/okhttp/src/main/scala/coursier/cache/protocol/HttpHandler.scala
+++ b/modules/okhttp/src/main/scala/coursier/cache/protocol/HttpHandler.scala
@@ -1,17 +1,51 @@
 package coursier.cache.protocol
 
 import java.net.{URLStreamHandler, URLStreamHandlerFactory}
+import java.util.concurrent.TimeUnit
 
 import com.squareup.okhttp.{OkHttpClient, OkUrlFactory}
 
+import scala.concurrent.duration.FiniteDuration
+
+/**
+  * Contains the default values for the OK Http client.
+  *
+  * The default client can be configured using properties which are first checked from the system properties, then environment variables:
+  *
+  * - COURSIER_CONN_TIMEOUT_SECONDS : the http connection timeout, in seconds
+  * - COURSIER_READ_TIMEOUT_SECONDS : the http read timeout, in seconds
+  *
+  */
 object HttpHandler {
-  lazy val okHttpClient = new OkHttpClient
-  lazy val okHttpFactory = new OkUrlFactory(okHttpClient)
+
+  private def durationInSecondsFromEnv(key : String): Option[FiniteDuration] = {
+    import concurrent.duration._
+    sys.props.get(key).orElse(sys.env.get(key)).map(_.toInt.seconds)
+  }
+
+  def newClient() = {
+    val client = new OkHttpClient
+
+    durationInSecondsFromEnv("COURSIER_CONN_TIMEOUT_SECONDS").foreach { timeoutSeconds =>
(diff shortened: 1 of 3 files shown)


aaronp:
I'd like this PR to go in officially, rather than having to build/republish and use our forked version

modules/okhttp/src/main/scala/coursier/cache/protocol/HttpHandler.scala
```diff
@@ -1,17 +1,51 @@
...
+
+  lazy val okHttpClient = newClient
+
+  lazy val okHttpFactory: OkUrlFactory = new OkUrlFactory(okHttpClient)
 }
 
-class HttpHandler extends URLStreamHandlerFactory {
-  def createURLStreamHandler(protocol: String): URLStreamHandler =
-    HttpHandler.okHttpFactory.createURLStreamHandler(protocol)
+class HttpHandler(okHttpFactory: OkUrlFactory = HttpHandler.okHttpFactory) extends URLStreamHandlerFactory {
```
alexarchambault: This class is [instantiated by reflection](https://github.com/coursier/coursier/blob/dd343267654575ed11265b61de9dc7bf57d14b94/modules/cache/jvm/src/main/scala/coursier/Cache.scala#L225). Are you sure adding an argument doesn't break that?

modules/okhttp/src/main/scala/coursier/cache/protocol/HttpHandler.scala
```diff
@@ -1,17 +1,51 @@
...
+
+  private def durationInSecondsFromEnv(key : String): Option[FiniteDuration] = {
+    import concurrent.duration._
+    sys.props.get(key).orElse(sys.env.get(key)).map(_.toInt.seconds)
+  }
+
+  def newClient() = {
+    val client = new OkHttpClient
+
+    durationInSecondsFromEnv("COURSIER_CONN_TIMEOUT_SECONDS").foreach { timeoutSeconds =>
```
alexarchambault: Couldn't this be just `COURSIER_CONN_TIMEOUT` and `COURSIER_READ_TIMEOUT`, then parsed with `scala.concurrent.duration.Duration`? This would put these variables in line with `COURSIER_TTL`.

Also, to keep things line with ttl too, couldn't the Java property be read as e.g. `coursier.conn.timeout` and `coursier.read.timeout`, and only as `COURSIER_CONN_TIMEOUT` and `COURSIER_READ_TIMEOUT` from the environment?

modules/okhttp/src/main/scala/coursier/cache/protocol/HttpHandler.scala
aaronp: Thanks very much for your comments!

I wasn't aware the OkHttp wasn't the default - I'll certainly look into this and revise the PR. Much appreciated!

mkurz:
I guess there is no way yet to define timeouts when fetching dependencies?
Recently a customer's repo started te become slow and we end up with 
```
[error]   download error: Caught java.net.ConnectException: Connection timed out (Connection timed out) (Connection timed out (Connection timed out)) while downloading https://...
```
Probably this pull request needs some more work, given that by default coursier [relies on](https://github.com/coursier/coursier/blob/v2.0.0-RC5/doc/docs/other-proxy.md) `java.net.HttpURLConnection`, so there are [some other places](https://github.com/coursier/coursier/search?q=HttpURLConnection&unscoped_q=HttpURLConnection) where the timeout should be implemented.

PR #966, opened 2018-11-27
Don't clear console for cached bootstrap binaries

olafurpg:


olafurpg:
Manually validated that the fix works as expected. I generated a `target/mtags` boostrap binary with --standalone.

Before fix
```scala
@ val out = new java.io.ByteArrayOutputStream
@ System.setErr(new java.io.PrintStream(out))
@ "./target/mtags".!
res9: Int = 0
@ out.toString()
java.lang.IllegalArgumentException: Unknown ansi-escape [2K at index 16 inside string cannot be parsed into an fansi.Str
```

After fix:
```scala
@ val out = new java.io.ByteArrayOutputStream
out: java.io.ByteArrayOutputStream =

@ System.setErr(new java.io.PrintStream(out))


@ import sys.process._
import sys.process._

@ "./target/mtags".!
Hello world!
res3: Int = 0

@ out.toString()
res4: String = ""
```

alexarchambault:
Merging, thanks @olafurpg!

PR #976, opened 2018-11-29
Allow to follow HTTP -> HTTPS redirections

alexarchambault:
Disabled by default in the API, not to run into possible security issues (https://stackoverflow.com/a/1884427/3714539). Pass `followHttpToHttpsRedirections = true` to `Cache.fetch` and `Cache.file` to enable it.

Enabled by default in the command-line (less likely a custom authenticator was set up there).


alexarchambault:
Should fix https://github.com/coursier/coursier/issues/889.

PR #985, opened 2018-12-06
Use checksums from header fields if available (WIP)

alexarchambault:


alexarchambault:
Now at https://github.com/coursier/coursier/pull/1297.

PR #1015, opened 2019-01-15
Various version & URL fixes in CLI install

dwijnand:


alexarchambault:
Thanks @dwijnand! There's no need to fix the `M8` / `M9` reference. `1.1.0-M9` was published in both ways (although maybe that particular point could be made in the doc).

PR #1023, opened 2019-01-17
Don't register custom URLStreamHandlerFactory from bootstrap launcher

alexarchambault:


alexarchambault:
Seems `URLClassLoader` chokes on `URL`s whose handler isn't the default one of their protocol… Seems [spring boot](https://github.com/spring-projects/spring-boot/tree/cea6a79e4981ac372e67134f537d2090691db3d5#spring-boot-loader) has [workarounds](https://github.com/spring-projects/spring-boot/tree/cea6a79e4981ac372e67134f537d2090691db3d5/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar) for that… Maybe bits of it could be added here.

PR #1030, opened 2019-01-23
Don't automatically lookup in Scaladex anymore

alexarchambault:
Previously, things not recognized as dependencies were assumed to be Scaladex queries, and were looked into it, like
```bash
$ coursier launch ammonite
```
(which currently resolves the spurious `com.simianquant:ammonite-kernel_2.12:0.4.2`…)

Scaladex lookups now have to be passed via `--scaladex`, like
```bash
$ coursier launch --scaladex ammonite
```
(still spurious lookup though…)


2m:
Woa. TIL that coursier looks at scaladex. Neat. :)

alexarchambault:
Yeah, I wish @MasseGuillaume was still at the Scala Center to help make it work seamlessly.

PR #1040, opened 2019-01-29
Remove HEAD argument from brew install

sbrunk:
Due to the replacement of head with an explicit version in https://github.com/coursier/homebrew-formulas/commit/5c0b71234d143c8de816f662d7c16fd6d9ee6de9, the `--HEAD` option doesn't work anymore.


alexarchambault:
Thanks @sbrunk!

PR #1041, opened 2019-01-31
Modified other-proxy.md

mskimm:
To execute a jar file, Java options should be placed in front of `-jar` like:

java [-options] -jar jarfile [args...]


alexarchambault:
Indeed, thanks!

PR #1042, opened 2019-01-31
Add better ways to print conflicts from command-line

alexarchambault:


alexarchambault:
Closing (will re-open after that) to trigger Appveyor (seems its event log only sees the branch, not the PR…).

PR #981, opened 2018-12-02
Allow linking flags to be passed through LDFLAGS environment variable

jgogstad:
Allow the user to specify linking flags through `LDFLAGS` when creating native bootstraps.

This allows for instance for linking applications that uses STTP for Scala Native (requires `LDFLAGS=-L/usr/local/opt/curl/lib`, see [sttp#Building & testing the scala-native backend](https://github.com/softwaremill/sttp#building--testing-the-scala-native-backend))

PR #1054, opened 2019-02-11
Support multiple searchpaths for ld in LDFLAGS

jgogstad:
This was a stupid mistake in https://github.com/coursier/coursier/pull/981. `ld` supports multiple library paths by passing the same option multiple times (see https://linux.die.net/man/1/ld, search for `--library-path` to get to the correct part of the man page).

Makes sense to support multiple options to `ld` anyway. Users might want to pass other flags than `-L` when creating native bootstraps with coursier.

Not sure what upstream branch to target.

### DIFF
--- a/modules/extra/src/main/scala-2.12/coursier/extra/Native.scala
+++ b/modules/extra/src/main/scala-2.12/coursier/extra/Native.scala
@@ -45,7 +45,7 @@ object Native {
   }
 
   private def linkingOptions(): Seq[String] = sn.Discover.linkingOptions() ++
-    sys.env.get("LDFLAGS").toSeq
+    sys.env.get("LDFLAGS").toSeq.flatMap(_.split(","))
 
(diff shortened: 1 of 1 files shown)


alexarchambault:
@jgogstad Shouldn't this be split by `"\\s+"` rather than `","`? It seems it's more customary to separate arguments with spaces in `LDFLAGS`.

jgogstad:
You're right, I was sure I had an example with comma, but I'm unable to reproduce it. [This one](https://www2.cs.duke.edu/courses/spring04/cps108/resources/makefiles/sample.html) uses spaces.

Changed.

alexarchambault:
Thanks, merging then!

jgogstad:
Any chance to have it added to M11-2 or a M12?

alexarchambault:
I'm merging a few more fixes, and I'll do a release after that I guess.

PR #1071, opened 2019-02-20
Included Fury in the list of users

propensive:


alexarchambault:
@propensive I opened https://github.com/coursier/coursier/pull/1072 to mention both bazel-deps and fury. Is that fine for you?

alexarchambault:
Done in https://github.com/coursier/coursier/pull/1072.

PR #1157, opened 2019-05-07
Remove possibly corrupt checksum files

alexarchambault:
The file we compute a checksum for was already removed since #797. This removes the (possibly faulty) checksum file too.


alexarchambault:
cc @wisechengyi (This extends https://github.com/coursier/coursier/pull/797 somehow.)

PR #1158, opened 2019-05-07
Filter dependency tree using module regex

viqueen:
Just supporting something similar to what maven dependency tree plugin does over here https://maven.apache.org/plugins/maven-dependency-plugin/examples/filtering-the-dependency-tree.html

`coursier` already comes with a `ModuleMatcher` , I naively made use of that when resolving dependencies using the `--what-depends-on` option.

also added a test because test are always nice to have.

### DIFF
--- a/modules/cli/src/main/scala-2.12/coursier/cli/params/ResolveParams.scala
+++ b/modules/cli/src/main/scala-2.12/coursier/cli/params/ResolveParams.scala
@@ -19,7 +19,7 @@ final case class ResolveParams(
   benchmarkCache: Boolean,
   tree: Boolean,
   reverseTree: Boolean,
-  whatDependsOn: Set[ModuleMatcher],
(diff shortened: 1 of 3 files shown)


modules/cli/src/main/scala-2.12/coursier/cli/params/ResolveParams.scala
```diff
@@ -18,7 +19,7 @@ final case class ResolveParams(
   benchmarkCache: Boolean,
   tree: Boolean,
   reverseTree: Boolean,
-  whatDependsOn: Set[Module],
+  whatDependsOn: Set[ModuleMatcher],
```
alexarchambault: Nit: I guess we can make that a simple `Seq` now (it was a `Set` only to check for inclusion, which we don't do anymore now).

alexarchambault:
Merging, thanks @viqueen!

Issue #755, opened 2018-01-26
The way ivy.xml files are found is not robust


benjyw:
Ivy URL/Filesystem repos have separate patterns for finding ivy.xml files vs. finding artifacts. 

E.g., 

```
<url>
  <ivy pattern="http://foo.com/[module]/[revision]/ivy.xml" />
  <artifact pattern="http://foo.com/[module]/[revision]/[artifact]-[revision].[ext]" />
</url>
```
Coursier, however, appears to expect to find an artifact and its accompanying ivy.xml using the same pattern (with artifact="ivy" and ext="xml"). This means that in practice all patterns must end with `/[artifact].[ext]`, or the ivy.xml files must have non-standard names.

This can lead to weird behavior, such as attempting to parse a .jar file as XML (e.g., if the pattern ends with `[module]-[revision].jar` because the author doesn't care about the ivy.xml at all, see https://github.com/coursier/coursier/issues/754)

Update: I guess my statement that Coursier "appears to expect to find an artifact and its accompanying ivy.xml using the same pattern" is inaccurate, since it can find each using a separate pattern, I think. But there still remains the problem that in some cases it can match on a file that isn't an ivy.xml file, but will still attempt to parse it as XML. That said, in practice, resolving https://github.com/coursier/coursier/issues/754 would solve the current issues I'm having. The rest could be worked around with careful use of patterns.

alexarchambault:
Having different patterns for metadata and artifacts is supported when using coursier via its API or its sbt plugin, but not via the CLI. But that should easy to add, one has just to find a right syntax for that… (and handle it around [that line](https://github.com/coursier/coursier/blob/9c274c23d6626d4327e4778bdca6cd5149ca321b/core/shared/src/main/scala/coursier/util/Parse.scala#L319))

PR #1166, opened 2019-05-13
Accept metadata-specific pattern for Ivy repos on the command-line

alexarchambault:
Fixes https://github.com/coursier/coursier/issues/755.

Use like
```
$ coursier resolve -r ivy:https://repo.com/ivy/[defaultPattern]|https://repo.com/ivy/[custom]/[metadata-pattern]
```


aishfenton:
@alexarchambault the pipe character clashes with it already being used as a delimiter in the COURSIER_REPOSITORIES environment variable though doesn't it? Our company uses a custom metadata pattern, so ideally I'd be able to set this just once in a dot file. 

Actually, we'd also need to specify the metadata pattern in the install json file too.

alexarchambault:
@aishfenton I would have expected the `ivy:artifact-pattern|metadata-pattern` to work from app descriptors (the JSON files describing apps).

PR #1172, opened 2019-05-17
Add install command

alexarchambault:


d10xa:
This is amazing feature! Why it is still experimental?

PR #1180, opened 2019-05-21
Various things

alexarchambault:
These are useful for the upcoming `install` command.


alexarchambault:
@smarter FYI, the first commit (`java.class.path` thing) allows to do stuff like
```bash
$ coursier -J-Dscala.usejavacp=true launch \
    ch.epfl.lamp:dotty-compiler_0.14:0.14.0-RC1 \
    -M dotty.tools.repl.Main
scala> …
```
or
```bash
$ coursier bootstrap \
    ch.epfl.lamp:dotty-compiler_0.14:0.14.0-RC1 \
    -M dotty.tools.repl.Main \
    -o dotty-repl \
    --property scala.usejavacp=true
$ ./dotty-repl
scala> …
```

PR #1202, opened 2019-06-11
Support SHA-512 checksum

### LINKED ISSUE #1108: Support SHA-512
As dicussed on [Giiter](https://gitter.im/coursier/coursier?at=5ca3cb0893fb4a7dc2c30b0a) I am trying to download a file directly using `FileCache()`. I initialized the cache using:
 ```
 val checksums = Seq(Some("MD5"),Some("SHA-1"),Some("SHA-256"),Some("SHA-512"), None)
 val fileCache: Cache[Task] = coursier.cache.FileCache().withChecksums(checksums)
 ```
 The download works very well but SHA-512 checksum doesn't work (sha1,sha-256 do work) and fails with `checksum format error`. It seems that the parser isn't build to handle this format.
https://github.com/coursier/coursier/blob/2abcacee5f73c9890eeb7854d246ddb232433df1/modules/cache/jvm/src/main/scala/coursier/cache/CacheChecksum.scala#L9-L10

I am not sure whether SHA-512 is widely used but in my case I want to directly download the artifact from the vendor's site.
For example, Elasticsearch now supports ONLY SHA-512. An example of the 6.7.0 version is:
```
9acf1aa19ca16cb9f3e00ac90e8df71d2f4ad01f0880ee4bbf35de8048295d34131553d50421121ffff018f86c547b57e8046ee8bc02079aed2adabee0600e2d  elasticsearch-oss-6.7.0.tar.gz
```

Thanks!

lavrov:
Closes #1108


alexarchambault:
Thanks!

PR #1210, opened 2019-06-18
Support latest subrevision with literal component

avdv:
We are using versions that look like this:

* `1.2.3-release-20190218T104029`
* `1.2.3-rc-20190308T090142`
* `1.2.3-alpha-32-20190311T150837`

In our projects, we (lazily) use dynamic revisions for *internal* `libraryDependencies` (or SBT plugins) like this:
```
"de.intern" %% "project-lib" % "1.2.3-release-+"
```
This works perfectly with vanilla SBT, but fails when using coursier.

This PR is an effort to fix this. I am using the `MAX` version component which simplifies constructing a version range since one does no longer need to compute the-next-version-number.

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Parse.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Parse.scala
@@ -21,15 +21,9 @@ object Parse {
         for {
           from <- version(prefix)
           if from.rawItems.nonEmpty
-          last <- Some(from.rawItems.last).collect { case n: Version.Numeric => n }
-          // a bit loose, but should do the job
-          if from.repr.endsWith(last.repr)
-          // appending -a1 to the next version, so has not to include things like
-          // nextVersion-RC1 in the interval - nothing like nextVersion* should be included
-          to <- version(from.repr.stripSuffix(last.repr) + last.next.repr + "-a1")
+          to <- version(prefix + "-max")
           // the contrary would mean something went wrong in the loose substitution above
-          if from.rawItems.init == to.rawItems.dropRight(2).init
-          if to.rawItems.takeRight(2) == Seq(Version.Literal("a"), Version.Number(1))
+          if from.rawItems.init == to.rawItems.dropRight(1).init
         } yield VersionInterval(Some(from), Some(to), fromIncluded = true, toIncluded = false)
       case _ =>
         None
--- a/modules/core/shared/src/main/scala/coursier/core/Parse.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Parse.scala
@@ -13,18 +13,19 @@ object Parse {
   }
 
   // matches revisions with a '+' appended, e.g. "1.2.+", "1.2+" or "1.2.3-+"
-  private val latestSubRevision = "(.*[^.-])[.-]?[+]".r
+  private val latestSubRevision = "(.*[^.-])([.-]?)[+]".r
 
   def ivyLatestSubRevisionInterval(s: String): Option[VersionInterval] =
     s match {
-      case latestSubRevision(prefix) =>
+      case latestSubRevision(prefix, delim) =>
         for {
           from <- version(prefix)
           if from.rawItems.nonEmpty
-          to <- version(prefix + "-max")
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/coursier/core/Parse.scala
```diff
@@ -21,15 +21,9 @@ object Parse {
...
-          if from.repr.endsWith(last.repr)
-          // appending -a1 to the next version, so has not to include things like
-          // nextVersion-RC1 in the interval - nothing like nextVersion* should be included
-          to <- version(from.repr.stripSuffix(last.repr) + last.next.repr + "-a1")
+          to <- version(prefix + "-max")
           // the contrary would mean something went wrong in the loose substitution above
-          if from.rawItems.init == to.rawItems.dropRight(2).init
-          if to.rawItems.takeRight(2) == Seq(Version.Literal("a"), Version.Number(1))
+          if from.rawItems.init == to.rawItems.dropRight(1).init
         } yield VersionInterval(Some(from), Some(to), fromIncluded = true, toIncluded = false)
```
alexarchambault:
```suggestion
         } yield VersionInterval(Some(from), Some(to), fromIncluded = true, toIncluded = true)
```
Change `toIncluded` to `true`? So that the version ending with `-max` falls into the interval too (`1.2-max` should match `1.2+` I guess).

modules/core/shared/src/main/scala/coursier/core/Version.scala
```diff
@@ -188,7 +188,7 @@ object Version {
 
       if (isNumeric(item)) {
         val nextNonDotZero = _tokens.dropWhile{case (Tokenizer.Dot, n: Numeric) => n.isEmpty; case _ => false }
-        if (nextNonDotZero.forall(t => t._1 == Tokenizer.Hyphen || ((t._1 == Tokenizer.Dot || t._1 == Tokenizer.None) && !isNumeric(t._2)))) { // Dot && isNumeric(t._2)
+        if (nextNonDotZero.forall(t => !isMinMax(t._2) && (t._1 == Tokenizer.Hyphen || ((t._1 == Tokenizer.Dot || t._1 == Tokenizer.None) && !isNumeric(t._2))))) { // Dot && isNumeric(t._2)
```
alexarchambault: Are the changes in this file really necessary? Seems everything still works fine without those.

modules/core/shared/src/main/scala/coursier/core/Parse.scala
avdv: I thought `max` is not a legit component of a version, rather it is a placeholder for a "maximal number" in a version range / spec? So you should never actually see a version `1.2-max`...

modules/core/shared/src/main/scala/coursier/core/Version.scala
avdv: Yes, they are necessary if you have a zero in your version at some place. `1.2.0-+` should mean `1.2.0-SOMETHING`, but with "normal" versions, `1.2.0` is equivalent to `1.2` and thus the zero is redacted. If you add a `-max` to that number you have `1.2-max` which has a totally different meaning than desired.

alexarchambault: Add a test case for it then?

modules/core/shared/src/main/scala/coursier/core/Parse.scala
alexarchambault: Still, with the former behavior, these `-max` versions were in the interval. Let's keep them there just in case for now.

modules/core/shared/src/main/scala/coursier/core/Version.scala
avdv: Sure thing.

modules/core/shared/src/main/scala/coursier/core/Parse.scala
avdv: Fine, I'll change that back.

alexarchambault:
Merging, thanks!

PR #1217, opened 2019-06-19
Take root dependencies into account in strict conflict mgr

alexarchambault:


alexarchambault:
cc @timo-schmid, for https://github.com/coursier/sbt-coursier/pull/75

(currently in the midst of trying to have `evicted` work for sbt-lm-coursier, the other coursier sbt plugin)

Issue #1251, opened 2019-07-02
Perf regression in recent versions of the sbt plugin


alexarchambault:
Seen when bumping the sbt plugin from `1.1.0-M14-4` to `2.0.0-RC2`.

For example:
```bash
$ git clone https://github.com/alexarchambault/ammonite-spark.git
$ cd ammonite-spark
$ git checkout 321bc918
$ sbt update
…
[success] Total time: 37 s, completed 2 juil. 2019 11:10:54
$ git revert -n 321bc918
$ sbt update
…
[success] Total time: 3 s, completed 2 juil. 2019 11:12:21
```

PR #1252, opened 2019-07-02
Fix perf regression

alexarchambault:
Fixes https://github.com/coursier/coursier/issues/1251.


regadas:
@alexarchambault Thanks for this! I was seeing the same pref degradation.

alexarchambault:
Yeah, sbt plugin `2.0.0-RC2-1` should be fine.

PR #1273, opened 2019-07-15
Don't use okhttp by default in CLI 

alexarchambault:


alexarchambault:
Seems okhttp creates some non-daemon threads on JDK11 (only when http2 is involved?). See discussion around [here](https://gitter.im/coursier/coursier?at=5d2c8478202bb93f00a9bad2).

alexarchambault:
cc @2m

2m:
Thanks for a quick fix!

PR #1291, opened 2019-07-24
Add optional SCM info.

### LINKED ISSUE #1288: SCM info is missing in definitions
I am working on https://github.com/fthomas/scala-steward/issues/82 to include a link to the upgraded project in the Scala Steward's PR.
I am using `coursier` to extract URLs of dependencies .

IIUC, `coursier` extracts `homePage` from `pom.xml`, but does not extract [SCM infos](https://maven.apache.org/scm/maven-scm-plugin/usage.html) currently.
https://github.com/coursier/coursier/blob/39d54c266db155a55eef11e2c1bac409d4ef1c1e/modules/core/shared/src/main/scala/coursier/core/Definitions.scala#L255-L261

`homePage` may be enough, but I think SCM info (URL to repository) is more useful, since it may use `Releases` pages that contain release note.

If adding SCM infos is acceptable, I am happy to open a PR.

exoego:
Closes #1288

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
@@ -257,7 +257,8 @@ final case class Info(
   homePage: String,
   licenses: Seq[(String, Option[String])],
   developers: Seq[Info.Developer],
-  publication: Option[Versions.DateTime]
+  publication: Option[Versions.DateTime],
+  scm: Option[Info.Scm]
 )
 
 object Info {
@@ -267,7 +268,13 @@ object Info {
     url: String
   )
 
-  val empty = Info("", "", Nil, Nil, None)
+  final case class Scm(
+    url: String,
+    connection: String,
+    developerConnection: String
+  )
+
+  val empty = Info("", "", Nil, Nil, None, None)
 }
 
 // Maven-specific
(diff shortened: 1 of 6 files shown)


exoego:
Looks like tests passed but mima checker failed due to new field in `Info` case class.
What should I do to resolve?

exoego:
Added mima filter and all mima error gone.
Only webasite job faillure, which seems I could not handle 🤔 
https://travis-ci.org/coursier/coursier/jobs/563406500#L965

alexarchambault:
Merging, thanks @exoego!

fthomas:
@exoego I've noticed that `scm` is `None` when `developerConnection` is missing in the POM. But not setting `developerConnection` seems to be common:
```
    <scm>
        <url>https://github.com/typelevel/cats-effect</url>
        <connection>git@github.com:typelevel/cats-effect.git</connection>
    </scm>
```

exoego:
I will submit a PR to coursier

fthomas:
Thanks! It seems that according to the [XSD](http://maven.apache.org/xsd/maven-4.0.0.xsd) all elements in `<scm>` are optional.

exoego:
Opened a PR in https://github.com/coursier/coursier/pull/1331

PR #1293, opened 2019-07-26
Reconciliation API

### LINKED ISSUE #1284: Latest-wins conflict resolution rules
## steps

https://github.com/eed3si9n/lagom-samples/commit/6f5bd56ae34c5f35d9ec7fee6d926bbda4451930

## problem

https://travis-ci.org/eed3si9n/lagom-samples/jobs/562204770

```
[error] lmcoursier.internal.shaded.coursier.error.ResolutionError$ConflictingDependencies: Conflicting dependencies:
[error] io.grpc:grpc-core:1.16.1:default(compile)
[error] io.grpc:grpc-core:1.20.0:default(compile)
[error] io.grpc:grpc-core:[1.16.1]:default(compile)
[error] io.grpc:grpc-core:[1.20.0]:default(compile)
[error] 	at lmcoursier.internal.shaded.coursier.Resolve$.validate(Resolve.scala:394)
[error] 	at lmcoursier.internal.shaded.coursier.Resolve.validate0$1(Resolve.scala:140)
[error] 	at lmcoursier.internal.shaded.coursier.Resolve.$anonfun$ioWithConflicts0$4(Resolve.scala:184)
[error] 	at lmcoursier.internal.shaded.coursier.util.Task$.$anonfun$flatMap$2(Task.scala:14)
[error] 	at scala.concurrent.Future.$anonfun$flatMap$1(Future.scala:307)
[error] 	at scala.concurrent.impl.Promise.$anonfun$transformWith$1(Promise.scala:41)
[error] 	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
[error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
[error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
[error] 	at java.lang.Thread.run(Thread.java:748)
[error] (hello-impl / update) lmcoursier.internal.shaded.coursier.error.ResolutionError$ConflictingDependencies: Conflicting dependencies:
[error] io.grpc:grpc-core:1.16.1:default(compile)
[error] io.grpc:grpc-core:1.20.0:default(compile)
[error] io.grpc:grpc-core:[1.16.1]:default(compile)
[error] io.grpc:grpc-core:[1.20.0]:default(compile)
```

## expectation

Smoother transition.

Is there a way to configure the conflict resolution rules so that it automatically uses the latest? For example in the case of the above, use `io.grpc:grpc-core:1.20.0`?

eed3si9n:
Fixes #1284

To allow customization of the reconciliation of version conflicts, this splits `mergeVersions` out into its own API called Reconciliation.


modules/coursier/shared/src/main/scala/coursier/Resolve.scala
```diff
@@ -180,7 +180,7 @@ final class Resolve[F[_]] private[coursier] (private val params: Resolve.Params[
           }
       }
 
-    S.bind(S.bind(run(initialRes))(S.point)) { res0 =>
```
eed3si9n: I think I needed this because initialRes already contains resolution conflict, and validate0 would throw there before my rule can get to it.

eed3si9n: I think I just need to move `validate0` to somewhere so it gets run at least once.

alexarchambault:
I'm not too comfortable around the idea of passing half-baked resolutions around where `validate0` was called…

The rules mechanism was just meant to validate resolutions, and optionally tweak them a bit, not recover them from errors like this.

Rules were fine for an equivalent of a strict conflict manager, because it only amounts to validate the final resolution, not recover it from errors.

As mentioned in https://github.com/coursier/coursier/issues/1284#issuecomment-514100424, I think it'd be better to just not have to raise conflicts during resolution, by offering the possibility to rely on a different mechanism than [this method](https://github.com/coursier/coursier/blob/afb6378b48bdc2c609b7fb1c16be55324556af1d/modules/core/shared/src/main/scala/coursier/core/Resolution.scala#L217) to reconcile versions during resolution. This is what I initially intended to do: maybe add a `Map[Module, Reconciliation]` in `coursier.core.Resolution`, along with a `Seq[(ModuleMatcher, Reconciliation)]` for patterns, to allow for custom version reconciliation mechanisms, via a new `Reconciliation` ADT. All of that provided there's no perf regression in the use case described in https://github.com/coursier/coursier/pull/1292.

eed3si9n:
> As mentioned in #1284 (comment), I think it'd be better to just not have to raise conflicts during resolution, by offering the possibility to rely on a different mechanism than this method to reconcile versions during resolution.

ok. I don't think I fully appreciated the comment until I wrote [semantics of dependency resolvers](http://eed3si9n.com/dependency-resolver-semantics) blog post over the weekend. I'll try again with your approach.

alexarchambault:
@eed3si9n Don't hesitate to ping me if you have any question about it (and thanks for taking a cut at this!)

eed3si9n:
@alexarchambault Ivy calls these a "mediator". What do you think of the name as opposed to `Reconciliation`?

alexarchambault:
I think I prefer `Reconciliation`… `Mediation` sounds too formal.

eed3si9n:
ok. I'd go with that then.

eed3si9n:
@alexarchambault I implemented Reconciliation.

modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
```diff
@@ -0,0 +1,118 @@
+package coursier.core
+
+/**
+ * Represents a reconciliation strategy given a dependency conflict.
+ */
+abstract class Reconciliation {
```
alexarchambault: Maybe we could make that one `sealed` in a first time. I'm not sure we want to allow users to supply custom `Reconciliation` for now.

modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Reconcile multiple version candidate.
+   *
+   * Returns `None` in case of conflict.
+   */
+  def reconcileVersions(module: Module, versions: Seq[String]): Option[String]
```
alexarchambault: Rename to `apply`? (down the line, we have `reconciler.reconcileVersions(…)`, which feels heavyweight)

alexarchambault: Let's not pass `module` either if it's not needed.

modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  abstract class AbstractReconciliation extends Reconciliation {
+    protected def mergeVersionConstraints(constraints: Seq[VersionConstraint]): Option[VersionConstraint]
+
+    /**
+     * Reconcile multiple version candidate.
+     *
+     * Returns `None` in case of conflict.
+     */
+    override def reconcileVersions(module: Module, versions: Seq[String]): Option[String] = {
```
alexarchambault: I'm not sure this is worth sharing across implementations… `RelaxedReconciliation` actually needs a slightly different processing here: if `retainedLatestOpt` is non empty, it should be retained (and the intevals / specific versions discarded). Else it should look at other stuff. The goal of retaining both a `latest.*` and an interval in the default `Reconciliation` is to allow to check that the version that get selected by `latest.*` is indeed in the interval (it's handled from `ResolutionProcess` IIRC). We don't need that with relaxed reconciliation.

modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
```diff
@@ -0,0 +1,118 @@
...
+        val parsedConstraints = standard.map(Parse.versionConstraint)
+        mergeVersionConstraints(parsedConstraints)
+          .flatMap(_.repr)
+      }
+  }
+
+  /**
+   * Implements the basic reconciliation rule based on `VersionConstraint.merge`.
+   */
+  abstract class BasicReconciliation extends AbstractReconciliation {
```
alexarchambault: Do we really need all these `Abstract…` classes? If little needs to be shared (see my comment above), let's keep everything in the `case object`.

modules/core/shared/src/main/scala/coursier/core/VersionConstraint.scala
```diff
@@ -61,4 +61,21 @@ object VersionConstraint {
...
+      )
+      val lowerBounds = nonZeroIntervals.flatMap(c =>
+        if (c.fromIncluded) c.from.toList
+        else Nil
+      )
+      val preferreds = constraints.flatMap(_.preferred).toList
+      val vs = (lowerBounds ++ preferreds).distinct
+      if (vs.isEmpty) None
+      else Some(VersionConstraint.preferred(vs.max))
+    }
```
alexarchambault: I'd like to propose a slightly different algorithm here, more consistent with the default one:
- sort constraints in ascending order in one way or another (using the lower bound of the intervals to compare them with specific versions and other intervals)
- from the right, merge them two-by-two with the `merge` method above, and stop (right before) `merge` returns `None`
- return the last successful merge (or the last element in the sorted list if no `merge` succeeded)

That way, relaxed merge will still behave almost like the other merge, only ignoring things that needs to be ignored. For example, if passed `[1.0,2.0)`, `[3.0,4.0)`, and `2.8`, `[1.0,2.0)` will be ignored, and it'll return `[3.0,4.0)` instead of just `3.0` with the current implementation (down the line, `[3.0,4.0)` will select the highest version in the interval).

modules/coursier/shared/src/main/scala/coursier/parse/ReconciliationParser.scala
```diff
@@ -0,0 +1,34 @@
...
+      case Left(e) => Left(e)
+      case Right(elems) =>
+        (elems.validationNelTraverse { case (m: Module, v: String) =>
+          val e = reconciliation(m, v)
+          ValidationNel.fromEither(e)
+        }).either
+    }
+  }
+
+  def reconciliation(module: Module, v: String): Either[String, (ModuleMatchers, Reconciliation)] = {
```
alexarchambault: Make that method `private`?

modules/coursier/shared/src/main/scala/coursier/parse/ReconciliationParser.scala
```diff
@@ -0,0 +1,34 @@
+package coursier.parse
+
+import coursier.core.{Module, ModuleName, Organization, Reconciliation}
+import coursier.util.{ModuleMatcher, ModuleMatchers, ValidationNel}
+import coursier.util.Traverse._
+
+object ReconciliationParser {
+  def reconciliation(input: Seq[String], scalaVersionOrDefault: String): Either[List[String], Seq[(ModuleMatchers, Reconciliation)]] = {
```
alexarchambault: Return a `ValidationNel` here rather than an `Either`?

modules/coursier/shared/src/test/scala/coursier/TestHelpers.scala
```diff
@@ -56,7 +56,19 @@ object TestHelpers extends PlatformTestHelpers {
...
+          val noComma = s.replaceAllLiterally(", ", "||")
+          noComma
+            .replaceAllLiterally("|None|", "")
+            .replaceAllLiterally("|List()|", "")
+            .replaceAllLiterally("|Map()|", "")
+            .replaceAllLiterally("HashSet", "Set")
+            .replaceAllLiterally("|Set()|", "")
+        }
+        val n = normalize(params0.toString)
+        // println(s"$n: old sha1 ${sha1(params0.toString)} new sha1 ${sha1(n)}")
```
alexarchambault: remove comment?

modules/core/shared/src/main/scala/coursier/core/Resolution.scala
```diff
@@ -907,6 +853,7 @@ final class Resolution private (
     errorCache: Map[Resolution.ModuleVersion, Seq[String]] = errorCache,
     finalDependenciesCache: Map[Dependency, Seq[Dependency]] = finalDependenciesCache,
     filter: Option[Dependency => Boolean] = filter,
+    reconciliation: Option[Module => Reconciliation] = reconciliation,
```
alexarchambault: No need to add that one here (that method is only there for compatibility with the original `copy` method, it doesn't need to accept new parameters).

alexarchambault: So the related mima rules aren't needed either.

modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
eed3si9n: ok.

modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
eed3si9n: Yea. If we are making it sealed, then I won't pass more than needed.

modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
eed3si9n: Addressed in 63d43919bc3a5b940bb93297e590588d02d75819

modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
eed3si9n: Addressed in https://github.com/coursier/coursier/pull/1293/commits/63d43919bc3a5b940bb93297e590588d02d75819

modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
eed3si9n: Done.

modules/core/shared/src/main/scala/coursier/core/VersionConstraint.scala
eed3si9n: Addressed in https://github.com/coursier/coursier/pull/1293/commits/bcff24b822d54f3aaba87e627388c8b1a01c5057

modules/coursier/shared/src/main/scala/coursier/parse/ReconciliationParser.scala
eed3si9n: Addressed in https://github.com/coursier/coursier/pull/1293/commits/7d8966f262200bff55a5bef70b19af11af0f7931

modules/coursier/shared/src/main/scala/coursier/parse/ReconciliationParser.scala
eed3si9n: Addressed in https://github.com/coursier/coursier/pull/1293/commits/7d8966f262200bff55a5bef70b19af11af0f7931

modules/coursier/shared/src/test/scala/coursier/TestHelpers.scala
eed3si9n: Done

modules/core/shared/src/main/scala/coursier/core/Resolution.scala
eed3si9n: Done

alexarchambault:
Merging, thanks @eed3si9n!

eed3si9n:
Nice.

PR #1301, opened 2019-08-02
nit

alexarchambault:


modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
```diff
@@ -13,66 +13,52 @@ sealed abstract class Reconciliation {
...
-   * Implements the basic reconciliation rule based on `VersionConstraint.merge`.
-   */
-  case object Basic extends AbstractReconciliation {
-    /**
-     * Reconcile multiple version candidate.
-     *
-     * Returns `None` in case of conflict.
-     */
-    override def apply(versions: Seq[String]): Option[String] = {
+  case object Default extends Reconciliation {
```
eed3si9n: I avoided the term Default because default also means fallback, i.e. you could create a setting to use Relaxed _by default_, and Default/Basic for some modules X.

alexarchambault: That sounded ok to me… (does it really feel that confusing?)

`Default` is the default in coursier if users don't change anything. If they ask for `Relaxed` to be used instead, they aren't using the default anymore.

Issue #1276, opened 2019-07-18
OverlappingFileLockException when accessing FileCache


eatkins:
I've noticed that sometimes, usually in CI, that update fails to work in newer versions of sbt that uses coursier by default. Here is a stactkrace from a failed ci [job](https://travis-ci.org/eatkins/io/jobs/560269334):

```
[error] java.nio.channels.OverlappingFileLockException
[error] 	at sun.nio.ch.SharedFileLockTable.checkList(FileLockTable.java:255)
[error] 	at sun.nio.ch.SharedFileLockTable.add(FileLockTable.java:152)
[error] 	at sun.nio.ch.FileChannelImpl.lock(FileChannelImpl.java:1063)
[error] 	at java.nio.channels.FileChannel.lock(FileChannel.java:1053)
[error] 	at lmcoursier.internal.shaded.coursier.paths.CachePath.withStructureLock(CachePath.java:126)
[error] 	at lmcoursier.internal.shaded.coursier.cache.CacheLocks$.withStructureLock(CacheLocks.scala:21)
[error] 	at lmcoursier.internal.shaded.coursier.cache.CacheLocks$.withLockOr(CacheLocks.scala:38)
[error] 	at lmcoursier.internal.shaded.coursier.cache.FileCache.$anonfun$download$27(FileCache.scala:456)
[error] 	at scala.concurrent.Future$.$anonfun$apply$1(Future.scala:658)
[error] 	at scala.util.Success.$anonfun$map$1(Try.scala:255)
[error] 	at scala.util.Success.map(Try.scala:213)
[error] 	at scala.concurrent.Future.$anonfun$map$1(Future.scala:292)
[error] 	at scala.concurrent.impl.Promise.liftedTree1$1(Promise.scala:33)
[error] 	at scala.concurrent.impl.Promise.$anonfun$transform$1(Promise.scala:33)
[error] 	at scala.concurrent.impl.CallbackRunnable.run(Promise.scala:64)
[error] 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
[error] 	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
[error] 	at java.lang.Thread.run(Thread.java:745)
```
I've seen this failure with sbt 1.3.0-RC2 and 1.3.0-RC3.

countfloyd:
Seeing this too in a bitbucket pipelines build.

```
[...]

PR #1304, opened 2019-08-02
Better JVM-wide locks

alexarchambault:
Fixes https://github.com/coursier/coursier/issues/1276 (hopefully).

### DIFF
--- a/modules/paths/src/main/java/coursier/paths/CachePath.java
+++ b/modules/paths/src/main/java/coursier/paths/CachePath.java
@@ -93,7 +93,22 @@ public class CachePath {
         return CoursierPaths.cacheDirectory();
     }
 
-    private static ConcurrentHashMap<Path, Object> processStructureLocks = new ConcurrentHashMap<Path, Object>();
+    // Trying to limit the calls to String.intern via this map (https://shipilev.net/jvm/anatomy-quarks/10-string-intern/)
+    private static ConcurrentHashMap<String, Object> internedStrings = new ConcurrentHashMap<>();
+
+    // Even if two versions of that code end up in the same JVM (say one via a shaded coursier, the other via a
+    // non-shaded coursier), they will rely on the exact same object for locking here (via String.intern), so that the
+    // locks are actually JVM-wide.
+    private static Object lockFor(Path cachePath) {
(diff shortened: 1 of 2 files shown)


modules/paths/src/main/java/coursier/paths/CachePath.java
```diff
@@ -93,7 +93,22 @@ public static File defaultCacheDirectory() throws IOException {
...
     }
 
-    private static ConcurrentHashMap<Path, Object> processStructureLocks = new ConcurrentHashMap<Path, Object>();
+    // Trying to limit the calls to String.intern via this map (https://shipilev.net/jvm/anatomy-quarks/10-string-intern/)
+    private static ConcurrentHashMap<String, Object> internedStrings = new ConcurrentHashMap<>();
+
+    // Even if two versions of that code end up in the same JVM (say one via a shaded coursier, the other via a
+    // non-shaded coursier), they will rely on the exact same object for locking here (via String.intern), so that the
+    // locks are actually JVM-wide.
+    private static Object lockFor(Path cachePath) {
```
eatkins: It might make sense to synchronize this method to be extra safe.

alexarchambault: We're only relying on thread-safe methods in it ([`String.intern`](https://stackoverflow.com/questions/14798003/is-string-intern-thread-safe), `ConcurrentHashMap.putIfAbsent`, …), it should be ok. Plus locks aren't free.

eatkins: I was thinking about what the jdk does for the classloading lock: https://github.com/openjdk/jdk/blob/439a48bd1cdaec0337f296102752c4253e1d944c/src/java.base/share/classes/java/lang/ClassLoader.java#L663
I guess the use of interned strings is more or less equivalent since you're guaranteed to get the same object out of key.intern()? Otherwise I could see a race condition where two threads concurrently intern the string and synchronize on a different instance. I'm not saying this is wrong since I don't fully understand it, but it would be more obviously correct if it was:
```
lock0 = internedStrings.putIfAbsent(internedKey, internedKey);
lock0 = lock0 == null ? internedKey : lock0;
```

PR #1314, opened 2019-08-08
Use Relaxed reconciliation by default

### LINKED ISSUE #1312: Should we use Relaxed by default?
Ref #1293 / #1284

For the most part Coursier resolves things similarly to how Ivy-based sbt 1.2.x resolves things. The major difference it seems is the way Default treats version ranges, which is now addressed by the new "Relaxed" reconciliation. sbt will likely make Relaxed the default.

Last week on sbt-contrib Gitter chat @smarter suggested:

> If this a good idea, why isn't it this the default in coursier ? Do we really need to introduce subtle differences between vanilla coursier and sbt built-in coursier?

I've been thinking more about this...

Like many software issues, I think this issue has interesting cultural background, in this case Maven culture and npm culture. Because npm encourages SemVered version specification `^2.0.0` or `[2.0.0,3)`, we see that often in npm-translated web jars. This is no different than some Java library's version 2.0.0 getting evicted by 3.0.0. I think it would be better default if latest-wins uses "Relaxed" reconciliation, and user who wants stronger semantics to use `--strict` rules or review eviction report afterwards.

What do you think?

eed3si9n:
Fixes #1312
Ref #1293 / #1284

This switches the default reconciliation to use the relaxed one that picks the latest version interval given multiple non-mergable ones.

### DIFF
--- a/modules/cli/src/main/scala/coursier/cli/options/ResolutionOptions.scala
+++ b/modules/cli/src/main/scala/coursier/cli/options/ResolutionOptions.scala
@@ -48,7 +48,7 @@ final case class ResolutionOptions(
     rules: List[String] = Nil,
 
   @Help("Choose reconciliation strategy")
-  @Value("organization:name:(basic|relaxed)")
+  @Value("organization:name:(relaxed|check-intervals)")
     reconciliation: List[String] = Nil,
 
   strict: Option[Boolean] = None,
--- a/modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Reconciliation.scala
@@ -43,7 +43,9 @@ object Reconciliation {
       Some(retained)
     }
 
-  case object Default extends Reconciliation {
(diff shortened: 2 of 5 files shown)


alexarchambault:
@eed3si9n As discussed in https://github.com/coursier/coursier/issues/1312, I'm not sure relaxed reconciliation should ever be the default. What I mean in https://github.com/coursier/coursier/issues/1312#issuecomment-519302938, is that the default reconciliation could be evolved at some point, so that it interprets having both `1.3` and `[2.0,3.0)` as a conflict. While still [being sound](https://github.com/coursier/coursier/pull/1302), and retaining some of its current properties, like selecting `1.3` when handed `1.3` and `[1.0,2.0)` together.

It's probably too early for such a change, all the more if the only motivation for such a change is cases like those of https://github.com/sbt/sbt/issues/4720.

eed3si9n:
That's fine. The purpose of this PR partly was to explore the scope and impact of the reconciliation switch in response to

> I'd like to check whether this breaks tests in coursier…

PR #1315, opened 2019-08-09
tweak Version ordering

### LINKED ISSUE #1313: Version ordering is surprising
Ref https://github.com/coursier/coursier/issues/420

## steps

```scala
scala> import coursier.core.Version
import coursier.core.Version
 
scala> def sortVersionsCoursier(versions: String*): List[String] =
     |   versions.toList.map(Version.apply).sorted.map(_.repr)
sortVersionsCoursier: (versions: String*)List[String]
 
scala> sortVersionsCoursier("1.0", "2.0", "1.0-alpha", "1.0+alpha", "1.0-X1", "1.0a", "1.0-RC1", "1.0-MF", "2.0.2")
```

## problem

Here's how Coursier 2.0.0-RC2-6 orders versions

- 1.0-alpha
- 1.0-RC1
- 1.0
- 1.0-X1
- 1.0+alpha
- 1.0a
- 1.0-MF
- 2.0
- 2.0.2

The fact that `-X1`, `a`, `-MF` suffix comes after `1.0` (final) is unexpected for library authors who used sbt to publish libraries.

## expectation

- 1.0-MF
- 1.0-X1
- 1.0a
- 1.0-alpha
- 1.0-RC1
- 1.0
- 2.0
- 2.0.2

`1.0+alpha` might be an exceptional case because of SemVer requirements.

eed3si9n:
Fixes #1313
Ref #420

This changes the version ordering similar to that of Apache Ivy.

Specifically, milestone qualifiers are treated as unknown, and they are sorted together with other unknown qualifiers such as `-X1` and `-MF`, all lower than the final `1.0.0`.

In addition, this adds a few other means of writing `1.max`:
- `1.x`
- `1.+`
- `1.*`

### before

Here's how Coursier 2.0.0-RC2-6 orders versions

- 1.0-alpha
- 1.0-M1
- 1.0-RC1
- 1.0
- 1.+
- 1.0-X1
- 1.0-MF
- 2.0
- 2.0.2
- 2.0+20130313144700

### after 

- 1.0-M1
- 1.0-MF
- 1.0-X1
- 1.0-alpha1
- 1.0-RC1
- 1.0
- 1.+
- 2.0
- 2.0+20130313144700
- 2.0.2

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Version.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Version.scala
@@ -199,53 +199,15 @@ object Version {
     }
   }
 
-  def postProcess(item: Item, tokens0: Stream[(Tokenizer.Separator, Item)]): Stream[Item] = {
+  // def isNumeric(item: Item) = item match { case _: Numeric => true; case _ => false }
 
-    val tokens =
-      // drop some '.0' under some conditions ???
-      if (isNumeric(item)) {
-        val nextNonDotZero = tokens0.dropWhile{case (Tokenizer.Dot, n: Numeric) => n.isEmpty; case _ => false }
-        if (nextNonDotZero.headOption.forall { case (sep, t) => sep != Tokenizer.Plus && !isMinMax(t) && !isNumeric(t) })
-          nextNonDotZero
-        else
-          tokens0
-      } else
-        tokens0
-
-    def ifFollowedByNumberElse(ifFollowedByNumber: Item, default: Item) = {
-      val followedByNumber = tokens.headOption
-        .exists{ case (Tokenizer.None, num: Numeric) if !num.isEmpty => true; case _ => false }
-
-      if (followedByNumber) ifFollowedByNumber
-      else default
-    }
-
-    val nextItem = item match {
-      case Tag("min") => Min
-      case Tag("max") => Max
-      case Tag("a") => ifFollowedByNumberElse(alphaQualifier, item)
-      case Tag("b") => ifFollowedByNumberElse(betaQualifier, item)
-      case Tag("m") => ifFollowedByNumberElse(milestoneQualifier, item)
-      case _ => item
-    }
-
-    def next =
-      if (tokens.isEmpty) Stream()
-      else postProcess(tokens.head._2, tokens.tail)
-
-    nextItem #:: next
-  }
-
-  def isNumeric(item: Item) = item match { case _: Numeric => true; case _ => false }
-
-  def isMinMax(item: Item) = {
-    (item eq Min) || (item eq Max) || item == Tag("min") || item == Tag("max")
-  }
+  // def isMinMax(item: Item) = {
+  //   (item eq Min) || (item eq Max) || item == Tag("min") || item == Tag("max")
+  // }
 
   def items(repr: String): List[Item] = {
     val (first, tokens) = Tokenizer(repr)
-
-    postProcess(first, tokens).toList
+    first :: tokens.toList.map(_._2)
   }
 
   @tailrec
--- a/modules/core/shared/src/main/scala/coursier/core/Version.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Version.scala
@@ -9,7 +9,7 @@ import coursier.core.compatibility._
  *  Same kind of ordering as aether-util/src/main/java/org/eclipse/aether/util/version/GenericVersion.java
  */
 final case class Version(repr: String) extends Ordered[Version] {
-  lazy val items: List[Version.Item] = Version.items(repr)
+  lazy val items: Vector[Version.Item] = Version.items(repr)
   def compare(other: Version) = Version.listCompare(items, other.items)
   def isEmpty = items.forall(_.isEmpty)
 }
@@ -26,11 +26,6 @@ object Version {
         case (Number(a), BigNumber(b)) => -b.compare(a)
         case (BigNumber(a), Number(b)) => a.compare(b)
         case (a @ Tag(_), b @ Tag(_)) => a.compareTag(b)
-        case (BuildMetadata(_), BuildMetadata(_)) =>
-          // Semver § 10: two versions that differ only in the build metadata, have the same precedence.
-          // Might introduce some non-determinism though :-/
-          0
-
         case _ =>
           val rel0 = compareToEmpty
           val rel1 = other.compareToEmpty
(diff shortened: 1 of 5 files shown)


alexarchambault:
This is way too much changes in the way versions are handled. Both in the semantics and in the internals of it.

If you want to handle `-MF` differently, why not just add an entry [around here](https://github.com/coursier/coursier/blob/c0a93d4b303bd1c000ae0cb813edcc9f5b00590d/modules/core/shared/src/main/scala/coursier/core/Version.scala#L95)?

Overall, this PR discards or changes way too many tests.

eed3si9n:
It's not just `-MF`, but about generally how unknown qualifiers are handled. There are lots of creative version schemes people have come up with like [2.12.4-bin-typelevel-4](https://mvnrepository.com/artifact/org.typelevel/scala-library/2.12.4-bin-typelevel-4), [1.0.0-X20](https://mvnrepository.com/artifact/org.scala-sbt/zinc_2.12/1.0.0-X20), [3.1.0-SNAP12](http://repo1.maven.org/maven2/org/scalatest/scalatest_2.13.0-RC3/3.1.0-SNAP12/). Given that these libraries are published using sbt, and consumed mostly by Ivy, wouldn't it make sense to keep the version ordering of Ivy?

alexarchambault:
I'd edge towards educating them about version schemes… `SNAP12` looks definitely like an abuse of notation to me… Same for `-X10`. About `2.12.4-bin-typelevel-4`, does its order compared to `2.12.4` really matter? This version / artifact is supposed to be more-or-less forced nonetheless.

modules/tests/shared/src/test/scala/coursier/test/VersionTests.scala
```diff
@@ -45,7 +45,9 @@ object VersionTests extends TestSuite {
 
     'buildMetadata - {
       * - {
-        assert(compare("1.2", "1.2+foo") < 0)
+        assert(compare("1.2", "1.2+foo") == 0)
```
eed3si9n: https://semver.org/spec/v2.0.0.html#spec-item-10 says:

> Build metadata MUST be ignored when determining version precedence. 

so not just vs `1.2+bar`, also vs plain `1.2`, `1.2+foo` should be treated the same.

modules/tests/shared/src/test/scala/coursier/test/VersionTests.scala
```diff
@@ -142,41 +143,14 @@ object VersionTests extends TestSuite {
       assert(compare("4.1.0-173", "4.1.1-178") < 0)
     }
 
-
-    'trailingDelimitersAreSemanticallyIrrelevant - {
-      assert(compare("1", "1............." ) == 0)
```
eed3si9n: I don't think there's a point in evaluating illegal versions.

modules/tests/shared/src/test/scala/coursier/test/VersionTests.scala
alexarchambault: Yes, but that introduces some non-determinism when ordering versions. (The underlying artifacts are likely to be different, so it's better not to randomly choose one of the versions.)

eed3si9n: The same goes for any of the things that would evaluate to `== 0`, like `ga` and `final`.
Also `2.0+20130313144700` coming after `2.0.2` is a bug, I'd say.

modules/core/shared/src/main/scala/coursier/core/Version.scala
```diff
@@ -65,180 +68,108 @@ object Version {
...
-      case Literal("b") => ifFollowedByNumberElse(betaQualifier, item)
-      case Literal("m") => ifFollowedByNumberElse(milestoneQualifier, item)
-      case _ => item
+    val num1 = first.takeWhile(_.isNumber)
+    val num2 = second.takeWhile(_.isNumber)
+    (num1.size, num2.size) match {
+      case (x, y) if x == y =>
+        listCompare0(first, second)
+      case (x, y) if x > y  =>
+        listCompare0(first, padNum(second, y, x))
```
eed3si9n: Instead of dropping `.0` during parsing, I am padding lhs or rhs to have the same digits before comparison. I think this results to a more accurate comparison.

modules/tests/shared/src/test/scala/coursier/test/VersionTests.scala
```diff
@@ -214,43 +187,57 @@ object VersionTests extends TestSuite {
...
+      assert(compare("1.0m-1", "1.0" ) < 0)
+
+      assert(compare("1.0.1-MF", "1.0.0" ) > 0)
+      assert(compare("1.0.1-MF", "1.0.1" ) < 0)
+      assert(compare("1.0.1-MF", "1.0.2" ) < 0)
+      assert(compare("1.0.1-X20", "1.0.0" ) > 0)
+      assert(compare("1.0.1-X20", "1.0.1" ) < 0)
+      assert(compare("1.0.1-X20", "1.0.2" ) < 0)
+      assert(compare("1.0.1-SNAP12", "1.0.0" ) > 0)
+      assert(compare("1.0.1-SNAP12", "1.0.1" ) < 0)
```
eed3si9n: > I'd edge towards educating people about version schemes… `SNAP12` looks definitely like an abuse of notation to me… Same for `-X10`. 

I think treating unknown qualifiers as pre-release is more sane way of handling the versions. See for example https://semver.org/spec/v2.0.0.html#spec-item-9:

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated identifiers immediately following the patch version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-].

alexarchambault:
Let me ponder those changes a few days more, before discussing them further. The code this touches is quite sensitive (both in terms of semantics and of perf), I'd prefer to be cautious before revamping it.

alexarchambault:
I’d be more confident in merging something along those lines if it consisted in atomic changes of the current code…

In particular, as this code is on the hot path (versions of all known dependencies are parsed at each resolution iteration), I’m wary of the regexes added in this PR (I basically removed all of the regexes used during resolution while profiling / optimizing resolution, with significant gains sometimes).

Wouldn’t it be possible, in several commits, to merge Qualifier and Literal in a first time (without touching the parser much), then remove the particular handling of separators, etc?

alexarchambault:
@eed3si9n Yes, let’s do this then!

eed3si9n:
@alexarchambault ok. I refactored it to make smaller changes incrementally.

alexarchambault:
@eed3si9n I reworked your commits in https://github.com/coursier/coursier/pull/1348, while ensuring the tests pass at each commit (so the commit adding tests is mixed in the others, and I also had to merge the pre/post-processing commits). Don't hesitate to review / have a look.

alexarchambault:
Merged via https://github.com/coursier/coursier/pull/1348.

PR #1329, opened 2019-08-14
Increase test timeout

alexarchambault:
It tends to fail on Travis CI from time to time else.


alexarchambault:
Now in https://github.com/coursier/coursier/pull/1330.

PR #1331, opened 2019-08-15
Scm.Info's fields should be optional.

exoego:
Follow-up of #1291

Current implementation fails to parse `<scm>` if not all fields are provided.
According to http://maven.apache.org/xsd/maven-4.0.0.xsd, `<scm>`'s fields should be optional.

cc: @fthomas
https://github.com/coursier/coursier/pull/1291#issuecomment-521518104


alexarchambault:
Thanks!

Issue #1335, opened 2019-08-21
--classpath-order for launched artifact


maiflai:
Hi,

Thanks very much for the addition of --classpath-order, it's much appreciated. 

However, it seems to place the launched artifact at the end of the list, rather than the start.

Is this intentional? I'll see if I can put a PR together.

Thanks again,
Stu.

PR #1338, opened 2019-08-21
ensure that root dependencies appear first in classpath order

maiflai:
Hi,

This seemed the simplest change to fix https://github.com/coursier/coursier/issues/1335.

I can look again at the recursive helper if that helps.

Thanks,
Stu


maiflai:
This seems to have broken quite a few tests (some of which were recently updated for classpath order).

Do you think I'm on the right track? Should I work through and update the test expectations?

Thanks,
Stu

alexarchambault:
Thanks for the suggestion @maiflai! The default ordering should be adjusted to match the one followed by Maven, yes, to give a higher priority to resources from a JAR over those of its dependencies. (I didn’t have that use case in mind when I added the ordering.)

I’ll only be able to have a closer look at that on Wednesday next week, but I think this could be achieved more consistently by tweaking the current sorting method instead of some post processing. So that we get JARs in topological order rather than its the current custom order. That way, the overriding of resources should happen for transitive dependencies too: if A depends on B, and B depends on C, we should get A.jar,B.jar,C.jar. I believe your changes currently give A.jar,C.jar,B.jar (and the current master of coursier gives C.jar,B.jar,A.jar).

maiflai:
Thanks - yes, I agree it should give A.jar,B.jar,C.jar.

I'll see if I can get something done by Wednesday and let you know.

alexarchambault:
@maiflai Is it something you'd like to get past the finish line? I would love to have this in the next release of coursier.

maiflai:
I'd love to have it in too - unfortunately I've not had a chance to sit down and get any further with it this week.

I might get some time over the weekend - first thing to do is fix the merge conflict.

Have you had a chance to look at the code change? I could work my way through any test failures (this seemed to take most of the time on my first attempt).

maiflai:
Sorry for the delay. I'm having a bit of trouble with a relatively simple example. Please could you help me to work through it?

**com.github.alexarchambault:argonaut-shapeless_6.1_2.11:0.2.0:default**

declares that it has dependencies on 
1. scala-library
2. argonaut
3. shapeless

argonaut declares a dependency on scalaz-core, which then brings its own dependencies.

1. scala-parser-combinators
2. scala-xml
3. monocle-core
4. ...

Where do you think the argonaut dependencies belong in a linear classpath? I think it would be natural to place them after the shapeless dependency (essentially making this breadth-first rather than depth-first).

1. scala-libary
2. argonaut
3. shapeless
4. argonaut dependencies (and other transient items)

The trivial version I have at the moment produces:

1. scala-library
2. argonaut
3. argonaut dependencies
4. shapeless

I'm not sure it's easy to do a breadth-first search because `dependenciesOf` does not yet provide any depth information. Does the linear version seem reasonable here?

Thanks,
Stu

maiflai:
> I'm not sure it's easy to do a breadth-first search because `dependenciesOf` does not yet provide any depth information.

Sorry, I now realise that it returns the next level of dependencies, so I can use it for breadth-first traversal. 

I'm working my way through the test cases validating and updating them.

maiflai:
I think that this change introduces more than just a re-ordering of the dependencies.

>modules/tests/shared/src/test/resources/resolutions/sh.almond/scala-kernel_2.12.7/0.2.2_params41c0fcd140480114f6f156bfda2c486c71bd3493 

I think on master it has **com.lihaoyi:fansi_2.12:0.2.5:default** at lines 63 and 97.

This PR only has the dependency once at line 63, which I think is an improvement?

alexarchambault:
Nice job updating the test fixtures!

A quick check shows that only duplicates were removed, so that's fine:
```
$ find modules/tests/shared/src/test/resources/resolutions -type f | sort | while read i; do
FORMER="$(gstat -c '%s' "../coursier-current-master/$i")"
NEW="$(gstat -c '%s' "$i")"
if [[ "$FORMER" != "$NEW" ]]; then
echo "$i\nFormer: $FORMER, new: $NEW"
colordiff <(sort -u < "../coursier-current-master/$i") <(sort -u < "$i")
fi
done
modules/tests/shared/src/test/resources/resolutions/com.netflix.karyon/karyon-eureka/1.0.28_default_compile_
Former: 2907, new: 2513
modules/tests/shared/src/test/resources/resolutions/org.apache.hadoop/hadoop-client/3.2.0
Former: 10177, new: 10221
modules/tests/shared/src/test/resources/resolutions/sh.almond/scala-kernel_2.12.7/0.2.2_params41c0fcd140480114f6f156bfda2c486c71bd3493
Former: 5367, new: 4959
modules/tests/shared/src/test/resources/resolutions/sh.almond/scala-kernel_2.12.7/0.2.2_paramsb4d5cba150d8c5bd829e4364f864f7908660eb41
Former: 5367, new: 4959
modules/tests/shared/src/test/resources/resolutions/sh.almond/scala-kernel_2.12.7/0.2.2_paramsf8a82cf727c6ce408fbe4977df2fd7976e091b83
Former: 5367, new: 4959
```
(no diffs removing the if condition too)

alexarchambault:
Merging, thanks @maiflai!

PR #1348, opened 2019-09-04
Fix / rework version ordering

alexarchambault:
Based on https://github.com/coursier/coursier/pull/1315.

Fixes https://github.com/coursier/coursier/issues/1313.


alexarchambault:
Let's merge, thanks for having pushed for it @eed3si9n!

Issue #1319, opened 2019-08-12
2.0.0-RC3 global credentials not used / unauthorized


francisdb:
upgraded sbt-coursier `1.0.2` to `2.0.0-RC3-2`

otherwise no changes to our build
credentials defined in `~/.sbt/1.0/global.sbt`

```
coursier.error.ResolutionError$CantDownloadModule: Error downloading com.my.company:jvm-metrics_2.12:2.9.2
[error]   Not found
[error]   Not found
...
[error]   unauthorized: https://nexus.server.host/repository/maven-releases/.../jvm-metrics_2.12/2.9.2/jvm-metrics_2.12-2.9.2.pom
```

sbrunk:
I'm seeing the same behaviour. When upgrading from a working `1.1.0-M12` to `2.0.0-RC3-2`, I'm getting `unauthorized` errors from our private Nexus.

alexarchambault:
@francisdb @sbrunk Are you able to reproduce that error from command-line? (see the examples [around here](https://get-coursier.io/docs/other-credentials)) That would help to know if the problems originates from coursier itself, or from its wiring with sbt.

sbrunk:
@alexarchambault I'm getting `unauthorized` errors when using the coursier CLI as well. I tried with `COURSIER_CREDENTIALS` and a properties file without success.

Interestingly, it seems to work using the legacy method of specifying credentials as part of the host:
```bash
$ coursier fetch -r https://user:password@artifacts.foo.com/maven...
```

jin:
+1, I can reproduce this in rules_jvm_external from the CLI.

francisdb:
For a repo to test with see instructions here:
https://github.com/francisdb/sbt-plugin-test
(from linked sbt issue)

leonardehrenfried:
I'm seeing this issue, too.

It worked in sbt 1.3.0-RC2 but not any version after that.

leonardehrenfried:
I've been able to narrow down the versions where this problem started appearing.

The last working version of the sbt plugin was `1.1.0-M14`.

From version `2.0.0-RC2` onwards this problem appears.

If you look at the what happened [between those two tags](https://github.com/coursier/coursier/compare/v1.1.0-M14...v2.0.0-RC2) there are indeed a few commits that deal with credentials.

Could they be the problem?

francisdb:
[...]

PR #1350, opened 2019-09-06
Basic auth is not supposed to be case sensitive

jsoref:
I spent the day looking into this for sbt/sbt#4852

Our nexus server (and almost certainly everyone else's) responds with:
```
www-authenticate: BASIC realm="Sonatype Nexus Repository Manager"
```

The code w/o this change looks for:
`Basic realm="..."`

But HTTP headers are generally case-insensitive.

You can see that the Rust implementation is explicitly case insensitive here:
https://docs.rs/www-authenticate/0.3.0/src/www_authenticate/lib.rs.html#190

```rust
impl Header for WwwAuthenticate {
    fn header_name() -> &'static str {
        "WWW-Authenticate"
    }

    fn parse_header(raw: &Raw) -> Result<Self> {
...
                let (scheme, challenge) = match stream.challenge() {
...
                map.entry(UniCase(CowStr(Cow::Owned(scheme))))
```

You can see that mozilla occasionally uses `basic realm=...` as in:
https://dxr.mozilla.org/mozilla-central/source/browser/components/urlbar/tests/browser/authenticate.sjs#139
```js
        response.setHeader("WWW-Authenticate", "basic realm=\"" + realm + "\"", true);
```

Here's the Gecko (necko) implementation which is clearly case-insensitive:
https://dxr.mozilla.org/mozilla-central/source/netwerk/protocol/http/nsHttpChannelAuthProvider.cpp#1019
```cpp

nsresult nsHttpChannelAuthProvider::GetAuthenticator(
    const char* challenge, nsCString& authType, nsIHttpAuthenticator** auth) {
...
  GetAuthType(challenge, authType);
...
  // normalize to lowercase
  ToLowerCase(authType);
```

### DIFF
--- a/modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
+++ b/modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
@@ -382,6 +382,7 @@ object CacheUrl {
     }
 
   private[coursier] val BasicRealm = (
+    "(?i)" + // case-insensitive, the response might be BASIC realm=
     "^" +
       Pattern.quote("Basic realm=\"") +
       "([^" + Pattern.quote("\"") + "]*)" +
--- a/modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala
+++ b/modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala
@@ -648,6 +648,80 @@ object FileCacheTests extends TestSuite {
 
       }
 
+      'ransomCase - {
+
+        val httpRealm = "simple realm"
+        val httpsRealm = "secure realm"
+
+        val httpUserPass = ("simple", "SiMpLe")
+        val httpsUserPass = ("secure", "sEcUrE")
+
+        def withServers[T](f: (Uri, Uri) => T): T = {
+
+          var httpsBaseOpt = Option.empty[Uri]
+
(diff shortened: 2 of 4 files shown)


francisdb:
fixes #1319

francisdb:
Better add a test for this tricky case

alexarchambault:
Nice catch, yeah. I had fixed that locally some time ago, but didn't take the time to push it :|

alexarchambault:
It should be possible to add tests for it [in that file](https://github.com/coursier/coursier/blob/8115b4fefdc24b452b64a4ca13d6eb45d8201dd9/modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala). Tweaking [this header](https://github.com/coursier/coursier/blob/8115b4fefdc24b452b64a4ca13d6eb45d8201dd9/modules/cache/jvm/src/test/scala-2.12/coursier/cache/TestUtil.scala#L103) in new tests in particular.

jsoref:
I wrote the test this morning and then spent the last hours trying to figure out how to run it.

I couldn't really figure that out, so instead I just asked Travis to try.

The tests failed without this fix:
https://travis-ci.org/jsoref/coursier/jobs/581781588
And pass with this fix:
https://travis-ci.org/jsoref/coursier/builds/581809682#L4546

So while I really don't understand the build system enough to run it locally, I believe the test is reasonably close enough to ship.

PR #1373, opened 2019-09-17
Scala 2.13 for coursier-cats-interop.

takayahilton:


alexarchambault:
@takayahilton Seems the error is legit. You need to filter out the `2.0.0-RC3-4` from `mimaPreviousArtifacts`. Something like
```scala
mimaPreviousArtifacts := {
  val previous = mimaPreviousArtifacts.value
  val sv = scalaVersion.value
  if (sv.startsWith("2.13."))
    previous.filter(_.revision != "2.0.0-RC3-4")
  else
    previous
}
```
in the cats module settings should do.

alexarchambault:
Merging, thanks @takayahilton!

PR #1380, opened 2019-09-24
Switch to scala 2.13.1

alexarchambault:


alexarchambault:
Seems `2.13.1` changed the encoding of some `AnyVal` stuff…

dwijnand:
Yeah, the generic signatures were fixed.  So those APIs were never right for Java users.

PR #1392, opened 2019-10-02
Use data-class for easier bin compat

alexarchambault:


alexarchambault:
@wisechengyi Forgot to ping you about this one… I disabled the pants CI job in it, as I was not sure how the [macro-paradise plugin](https://github.com/coursier/coursier/blob/ccb59e0c6934fcdbe526cfe19354f66f6975c7df/project/Settings.scala#L63) can be enabled with pants.

If ever enabling a compiler plugin is a problem, it's not necessary anymore in scala 2.13 (we only need a compiler flag then), so we could switch the CLI over to 2.13 then.

PR #1403, opened 2019-10-09
Implement scm parsing in PomParser

exoego:
Addresses https://github.com/fthomas/scala-steward/issues/923#issuecomment-538567839
so Scala Steward will include SCM links in pull requests for most artifacts.

Related to https://github.com/coursier/coursier/pull/1291

cc: @fthomas

I could not find a right place to add tests.
I will add tests if guidance provided 🙏

### DIFF
--- a/modules/core/jvm/src/test/scala/coursier/maven/PomParserTests.scala
+++ b/modules/core/jvm/src/test/scala/coursier/maven/PomParserTests.scala
@@ -21,7 +21,7 @@ object PomParserTests extends TestSuite {
           |</project>""".stripMargin
       )
       assert(success.isRight)
-      val scm = success.getOrElse(null).info.scm
+      val scm = success.right.get.info.scm
       assert(scm.isEmpty)
     }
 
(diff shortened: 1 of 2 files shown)


alexarchambault:
@exoego A small test under `modules/core/shared/src/test/scala/coursier` would be fine I think. It can call [`parseRawPomSax`](https://github.com/coursier/coursier/blob/0e3a00c9520371a5cb304af4bea8b598f9c69bca/modules/core/shared/src/main/scala/coursier/maven/MavenRepository.scala#L69) with some inlined POM data, and check that the resulting `Project` has the right SCM fields. [utest](https://github.com/lihaoyi/utest) already is around in the tests of `core`, see [this test](https://github.com/coursier/coursier/blob/46ad7243bea71e56dbda6df133a328535ac4c61e/modules/core/jvm/src/test/scala/coursier/credentials/DirectCredentialsTests.scala) for example.

exoego:
@alexarchambault 
Added test and squahsed commits.
Tests are added only in JVM, since Scala.js does not recognize `utest` as current build settings.

alexarchambault:
Merging, thanks @exoego!

PR #1416, opened 2019-10-15
Versions are kept in properites in poms often, these though should be…

ianoc-stripe:
… trimmed as sometimes they've got new lines in them

I'm not sure if this is the place you'd rather this fix happen/where the test should live.

But the context is 

https://repo1.maven.org/maven2/org/kie/kie-parent/7.27.0.Final/kie-parent-7.27.0.Final.pom

if you see
```
<version.org.jboss.spec.javax.xml.bind.jboss-jaxb-api_2.3_spec>1.0.1.Final </version.org.jboss.spec.javax.xml.bind.jboss-jaxb-api_2.3_spec>
```
it has a new line in the version being used, which then results in invalid URLs being created

```
Err: download error: Caught java.net.MalformedURLException: Illegal character in URL (Illegal character in URL) while downloading https://repo.maven.apache.org/maven2/org/jboss/spec/javax/xml/bind/jboss-jaxb-api_2.3_spec/1.0.1.Final
    /jboss-jaxb-api_2.3_spec-1.0.1.Final
    .pom
```


alexarchambault:
Thanks @ianoc-stripe!

Do you know of a specific dependency using `kie-parent` that raises this error? It could be added to [`ResolveTests`](https://github.com/coursier/coursier/blob/aad757661b635496eeccaa21dbf5ba4db8253bd0/modules/coursier/shared/src/test/scala/coursier/ResolveTests.scala) to prevent regressions.

One then needs to run those tests like
```
$ FETCH_MOCK_DATA=1 sbt "coursierJVM/testOnly coursier.ResolveTests"
```
then commit the metadata this writes… I can take care of it if that's too cumbersome, I just need to know the exact dependency that relies on `kie-parent`.

alexarchambault:
Merging, thanks @ianoc-stripe!

ianoc-stripe:
Excellent, thank you for shepherding this through!

PR #1427, opened 2019-11-06
Sync instance for Task: override definition of Map.

diesalbla:
The default implementation of `map` in terms of `point` and `flatMap` incurs the allocation cost of a Lambda expression for any call. For `Task`, this is easily avoidable.

When profiling a compilation, this was allocating 4 MiB, which was about 1/1000th of total allocated space.


alexarchambault:
Thanks @diesalbla!

PR #1428, opened 2019-11-06
FileUtils: avoid Integer Boxing

diesalbla:
The function `withContent` was defined to take a `Function2` object, in which the second item was an integer. Sadly, this means that those Int values need to be boxed, so this is a source of allocations.


alexarchambault:
Merging this, thanks @diesalbla! That may explain some checksums-related slowdowns I've been seeing in the native CLI.

PR #1439, opened 2019-11-20
Add API to list versions

alexarchambault:
Use like
```scala
import scala.concurrent.ExecutionContext.Implicits.global
import coursier._
val availableVersions = Versions()
  .withModule(mod"org.scala-lang:scala-library")
  .versions()
  .unsafeRun()
  .available // Seq[String]
```

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Versions.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Versions.scala
@@ -85,7 +85,14 @@ object Versions {
     hour: Int,
     minute: Int,
     second: Int
-  )
+  ) extends Ordered[DateTime] {
+    def compare(other: DateTime): Int = {
+      import Ordering.Implicits._
+      if (this == other) 0
+      else if (tuple < other.tuple) -1
+      else 1
+    }
+  }
 
   val empty = Versions("", "", Nil, None)
 }
--- a/modules/coursier/shared/src/main/scala/coursier/Versions.scala
+++ b/modules/coursier/shared/src/main/scala/coursier/Versions.scala
@@ -1,0 +1,90 @@
+package coursier
+
+import coursier.cache.Cache
+import coursier.core.Version
+import coursier.util.{Sync, Task}
+import dataclass.data
+
+@data class Versions[F[_]](
+  cache: Cache[F],
+  moduleOpt: Option[Module] = None,
+  repositories: Seq[Repository] = Resolve.defaultRepositories
+)(implicit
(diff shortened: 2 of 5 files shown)


fthomas:
Wow, this looks like it could be used in Scala Steward to find new versions instead of going through sbt and sbt-updates. This has the potential to make Scala Steward a lot faster.

Thanks @alexarchambault!

alexarchambault:
@fthomas But don't you need to start sbt anyway to know about the dependencies themselves (before checking for possible updates)?

alexarchambault:
FYI, I have [a project](https://github.com/alexarchambault/deps/blob/master/WIP-README.md) in the work, that aims to list dependencies outside of sbt itself, like [this](https://github.com/alexarchambault/case-app/blob/deps/deps.json). With those files, it should be possible to e.g. have a single JVM checking for updates for as many projects as necessary. (My goal was to ping you about it as a I'll publicly announce it :)

fthomas:
You're right. Not all invocations of sbt can be eliminated only those that check for updates of all known dependencies.

deps looks cool. Finally a principled way to organize dependencies where it is totally clear where version numbers are defined! It should be easy for Scala Steward to also update `deps.json` files. Even if a project does not use sbt. This is exciting!

fthomas:
@alexarchambault I'm now using Coursier exclusively to find new updates (https://github.com/fthomas/scala-steward/pull/1209) and it is working great so far. Going through all repos now takes 30-40 minutes instead of 5-6 hours. It is so fast now that it had been blacklisted by Maven Central a few times and I had to add a rate limiter to slow it down. :-)

Thanks again for this great addition to Coursier!

alexarchambault:
Thanks for the feedback, @fthomas!

fthomas:
@alexarchambault I was wondering if `Versions` does take [mirrors](https://get-coursier.io/blog/index.html#mirrors) into account? My current understanding of the code is that it does not, but I'm not sure.

I'm still having issues with Maven Central so I tried using a mirror but that didn't help for getting versions. WDYT, should support for mirrors be added to `Versions`?

fthomas:
I added minimal support for mirrors (by copying code from `Resolve`) here and it works for me locally: https://github.com/coursier/coursier/compare/master...fthomas:topic/use-mirrors-in-versions

Should I extend that and also add `mirrorConfFiles` and `mirrors` fields to `Version` so that mirrors can also be set via the API and then open a PR with these changes?

fthomas:
PR for mirrors support: #1515

PR #1522, opened 2020-01-27
Prevent HEAD request when using LocalOnlyIfValid cache policy

### LINKED ISSUE #1520: CachePolicy.LocalOnlyIfValid causes HEAD request if artifact is older than TTL
While experimenting with different cache policies I noticed that using [`LocalOnlyIfValid`](https://github.com/coursier/coursier/blob/47f9fa15d418547abb5b8a8fdd4dda43acaf8aff/modules/cache/jvm/src/main/scala/coursier/cache/CachePolicy.scala#L10-L11) can cause a `HEAD` request if an artifact is older than the TTL. This surprised me because I thought that `LocalOnlyIfValid` would not hit the network.

I noticed this because a `FileCache` with this policy returned a result even after the requested artifact was older than the TTL. Here is my test case:
```scala
package coursier

import coursier.cache.{CachePolicy, FileCache}
import utest._
import scala.async.Async.{async, await}
import scala.concurrent.duration._
import coursier.util.Task
import TestHelpers.ec

object VersionsTests extends TestSuite {
  val tests = Tests {
    "simple" - async {
      val module = mod"com.chuusai:shapeless_2.12"

      val versions0 = Versions()
        .withCache(FileCache[Task]().withCachePolicies(List(CachePolicy.ForceDownload)))
        .withRepositories(List(Repositories.central))
      val shapelessVersions0 = await(versions0.withModule(module).versions().future).available
      val expectedShapelessVersions0 = Seq("2.3.2", "2.3.3")
      assert(shapelessVersions0 == expectedShapelessVersions0)

      // sleep so that the artifact is older than the TTL of versions1's cache
      Thread.sleep(500)

      val versions1 = Versions()
        .withCache(FileCache[Task]().withTtl(10.millis).withCachePolicies(List(CachePolicy.LocalOnlyIfValid)))
        .withRepositories(List(Repositories.central))
      val shapelessVersions1 =  await(versions1.withModule(module).versions().future).available
      val expectedShapelessVersions1 = Seq.empty[String]
      assert(shapelessVersions1 == expectedShapelessVersions1)
    }
  }
}
```

Two different `FileCache`s are used here. The first uses the `ForceDownload` policy to ensure the downloaded artifact is fresh. The second cache uses the `LocalOnlyIfValid` policy which should (according to its documentation) only return local files that are younger than the TTL.

So my expectation was that `expectedShapelessVersions1` returns an empty `Seq` (or an error) because the already downloaded `shapeless_2.12/maven-metadata.xml` was older than the TTL of 10 milliseconds at the time `versions1...versions()` is called. But instead of an empty `Seq`, `shapelessVersions1` was the same as `shapelessVersions0`.

I then debugged into the `FileCache` implementation and found that in `download(...).shouldDownload(...).check` the `HEAD` request of the artifact URL is initiated. `check` is called [here](https://github.com/coursier/coursier/blob/47f9fa15d418547abb5b8a8fdd4dda43acaf8aff/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala#L290) after `checkNeeded` returns `true` (that is the case when the artifact is older than the TTL). Since the local file is younger than the URL, `shouldDownload` returns `false` and [this case](https://github.com/coursier/coursier/blob/47f9fa15d418547abb5b8a8fdd4dda43acaf8aff/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala#L609-L610) is hit in `download(...).res` which means `Right(())` is returned instead of a `Left(FileTooOldOrNotFound)` error which I would have expected.

fthomas:
Closes #1520.

I've checked that the test in https://github.com/coursier/coursier/issues/1520#issue-555294322 passes with this change.

### DIFF
--- a/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala
+++ b/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala
@@ -233,8 +233,7 @@ import dataclass.data
       }
     }
 
-    def shouldDownload(file: File, url: String): EitherT[F, ArtifactError, Boolean] = {
-
+    def shouldDownload(file: File, url: String, checkRemote: Boolean): EitherT[F, ArtifactError, Boolean] = {
       val errFile0 = errFile(file)
 
       def checkErrFile: EitherT[F, ArtifactError, Unit] =
@@ -286,7 +285,9 @@ import dataclass.data
               S.bind(checkNeeded) {
                 case false =>
                   S.point(Right(false))
-                case true =>
+                case true if !checkRemote =>
+                  S.point(Right(true))
+                case true if checkRemote =>
                   S.bind(check.run) {
                     case Right(false) =>
                       S.schedule(pool) {
@@ -587,7 +588,7 @@ import dataclass.data
             // )
             checkFileExists(file, url)
           } else {
-            def update = shouldDownload(file, url).flatMap {
+            def update = shouldDownload(file, url, checkRemote = true).flatMap {
               case true =>
                 remoteKeepErrors(file, url, keepHeaderChecksums)
               case false =>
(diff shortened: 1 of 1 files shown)


fthomas:
The Travis build https://travis-ci.org/coursier/coursier/jobs/642527350 failed because of 403s from Maven Central. I don't think they are related to this change.

alexarchambault:
I restarted it, it's green now.

Merging, thanks a lot @fthomas!

PR #1571, opened 2020-02-12
Add uninstall command, and more tweaking of the setup / install commands

alexarchambault:


alexarchambault:
Some commit messages could have been tidied up…

PR #1576, opened 2020-02-13
Remove spark-submit command from the default command list

alexarchambault:
It can still be run like
```
$ cs launch io.get-coursier::coursier-cli:2.0.0-RC6-6 -r typesafe:ivy-releases \
    -M coursier.cli.spark.SparkSubmit -- \
      spark-submit-options…
```


alexarchambault:
cc @darkjh @YannMoisan (don't know if you're still using it as is, or if you're using it via the latest version of coursier)

PR #1582, opened 2020-02-17
Fix trailing CRLF when adding directories to PATH on Windows

alexarchambault:


alexarchambault:
CI is red because of some npm network errors…

Issue #1483, opened 2020-01-04
Unify cache directory across all OSes


eed3si9n:
Ref https://github.com/coursier/coursier/issues/1438 / https://github.com/sbt/sbt/issues/5206
Ref https://github.com/sbt/sbt/issues/5120

## steps

Where's Coursier's cache?

## problem 

"It depends..."

## expectation

Let's change the default Coursier cache once and for all to a single location: **`$HOME/.cache/coursier/v1`**. For all operating systems.

(In my original description I had `$HOME/.coursier/cache/`. I meant the new `$HOME/.cache/coursier/`)

## notes

In theory, using something like directories-jvm and following the directory standard is a good idea, but it's messy in practice. If we _really_ want, I'd say only respect `$XDG_CACHE_HOME` nothing else.

soc:
> only respect `$XDG_CACHE_HOME` nothing else

How would that work vs. `$HOME/.coursier/cache/v1`? You can't have both.

eed3si9n:
@soc here's my proposal - https://github.com/coursier/coursier/pull/1484

eed3si9n:
I just realized I typo'ed in the original proposal. I meant `$XDG_CACHE_HOME/coursier/v1` otherwise `$HOME/.cache/coursier/v1`.

yanns:
I'd also agree to pick a single place.
I was testing on mac, and have not realized that the path was depending on the OS.
As a result, the cache was wrongly configured on the CI system.

soc:
@eed3si9n I think any app ignoring operating system rules is being a really poor citizen and people are rightfully annoyed. (Just search for "XDG" on GitHub and read the hundreds of issues of users being annoyed of XDG being ignored, now add to that the people annoyed by ignoring macOS conventions and the people annoyed by ignoring Windows conventions).

All that said, dumping dotfiles in users' home folders on Windows is not nice, but doable (I actually only know of applications that stopped doing this. SBT would be the first that actively made things worse).

I would be more concerned about not following Apple's rules to be honest: I wouldn't take it for granted that applications can still indiscriminately dump their files all over the place in a few years.

To be honest, I don't really see the point in the change: the paths will still differ across operating systems even after this, and it just makes coursier a worse citizen on 2 of the 3 main platforms it runs on.

dwijnand:
> but it's messy in practice

[...]

PR #1586, opened 2020-02-18
Remove support for ~/.coursier/cache cache location

alexarchambault:
Fixes https://github.com/coursier/coursier/issues/1483.

Don't know whether I want to write some kind of migration tool to help users migrate to the new location, prior to merging that.


eed3si9n:
Awesome.

PR #1656, opened 2020-04-11
Show help on empty args

regadas:
Users might be puzzled when they just run `cs` and nothing outputs.

### DIFF
--- a/modules/cli/src/main/scala/coursier/cli/Coursier.scala
+++ b/modules/cli/src/main/scala/coursier/cli/Coursier.scala
@@ -65,14 +65,10 @@ object Coursier extends CommandAppPreA(Parser[LauncherOptions], Help[LauncherOpt
       println("Press \"ENTER\" to continue...")
       val scanner = new Scanner(System.in)
       scanner.nextLine()
-    } 
-
-    if (args.isEmpty) {
+    } else if (args.isEmpty)
       helpAsked()
-    }
-    else {
+    else
       super.main(args)
-    }
   }
 
   def beforeCommand(options: LauncherOptions, remainingArgs: Seq[String]): Unit = {
--- a/modules/cli/src/main/scala/coursier/cli/Coursier.scala
+++ b/modules/cli/src/main/scala/coursier/cli/Coursier.scala
@@ -22,7 +22,6 @@ import io.github.alexarchambault.windowsansi.WindowsAnsi
 import shapeless._
 
 import scala.util.control.NonFatal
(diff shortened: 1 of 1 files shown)


modules/cli/src/main/scala/coursier/cli/Coursier.scala
```diff
@@ -57,16 +58,23 @@ object Coursier extends CommandAppPreA(Parser[LauncherOptions], Help[LauncherOpt
...
       Setup.run(SetupOptions(banner = Some(true)), RemainingArgs(Nil, Nil))
 
       // https://stackoverflow.com/questions/26184409/java-console-prompt-for-enter-input-before-moving-on/26184535#26184535
       println("Press \"ENTER\" to continue...")
       val scanner = new Scanner(System.in)
       scanner.nextLine()
-    } else
+    } 
+
+    if (args.isEmpty) {
```
alexarchambault, line 70: I think this should be checked only in the former else branch, like
```scala
  scanner.nextLine()
} else if (args.isEmpty)
  helpAsked()
else
  super.main()
```

On Windows, for now, when no arguments are passed, we run the `setup` command. That's what the first if branch if for. That allows Windows users to run `setup` by simply double-clicking on `cs.exe`. That means your changes would only apply on Linux / macOS, until the setup logic for Windows is moved to a separate launcher.

regadas: Ahah 😄 I had that initially, but then I thought that it would™ be cool to show the help usage after the setup, so in windows after the setup it would show up the help because the args are still empty. To be honest i'm not too familiar with win at how this would look like in that platform, but if you think it's better to have that way I'll change this for sure 😄

alexarchambault: This setup thing is mostly run when Windows users double-click on `cs.exe`. That opens a terminal window, which runs `cs.exe` with no args. With the branch above, that runs the `setup` command.

I don't think it's a good idea to print the help message there.

I'm going to try to restrict even more the conditions under which this branch is run (possibly not running it when `cs` is installed by `cs  install`), so that users would see the help message in that case too, like on Linux and macOS.

regadas: Hm ok gotcha! 👍  I'll update this! ... in any case I would like to help ... I still need to poke around a little bit more to get myself acquainted with the code base.

alexarchambault:
Merging, thanks @regadas!

PR #1657, opened 2020-04-11
Add in a list command to display installed apps

### LINKED ISSUE #1651: Possiblity of a cs list command or something to list installed apps
Unless I'm missing it, there isn't a great way for a user to tell what applications they have installed with `cs install <name>`. It would be fantastic to have a `cs list` command or something of that nature which would print out a list of the applications that are installed via `cs install`.

ckipp01:
This pr introduces a small `cs list` command to the cli.

Here is the example output.
```sh
❯ modules/cli/target/pack/bin/coursier list
amm coursier cs mdoc metac metap sbt scala scalac scalafmt
```

The only flag that I've enabled is the `--dir`.
```sh
❯ modules/cli/target/pack/bin/coursier list --dir /Users/ckipp/Library/Application\ Support/Coursier/bin
amm coursier cs mdoc metac metap sbt scala scalac scalafmt
```
### Reasoning

I'm in the habit of issuing `list` commands for the various package management systems that I use. For example, brew and yarn both have these options to get a quick glance of what has been installed. I've attempted `cs list` multiple times, and today I figured it'd be fun to just try to implement it.

### Approach

I'll be honest I just cobbled this together by picking the related parts from the `update` command, and I only kept the `--dir` flag. I wasn't fully sure what some of the use cases would be for some of the others, and all of the defaults seemed to just offer a lot of stuff that wouldn't be useful for list. If I've missed some that are necessary, please let me know, and I'll add them in.

### Potential Improvements

I know some package management systems have the option to have more verbose output and include things like version, late date updated etc. For this pr, I just did the bare minimum to ensure the way I was approaching it was correct. If you think some of these additional options are a good idea, I can look into including them as well, but I figured this would be a good start.

### Tests

I didn't notice explicit tests for some of the other commands, was I correct in assuming there isn't?

Closes #1651

### DIFF
--- a/doc/docs/cli-install.md
+++ b/doc/docs/cli-install.md
@@ -30,6 +30,13 @@ $ cs update
 $ cs uninstall scala
 ```
 
+[The `list` command](#list) allows you to view all of your currently installed
+applications
+
+```bash
+$ cs list
+```
+
 ## Application descriptors
 
 The `install` command goes from an application name (`scala`, `ammonite`, `scalafmt`, etc.)
@@ -152,6 +159,18 @@ $ cs install ammonite
 $ amm
 …
 $ cs uninstall amm # ammonite is installed as "amm"
+```
+
+## List
+
+The `list` command lists all the installed applications. By default this will
+list the applications in your [default installation
+directory](#installation-directory). However, you can explicitly provide a
+different directory with the `--dir` flag if you have your install directory in
+a custom location.
+
+```bash
+$ cs list --dir /myCustomDirectory
 ```
 
 ## Default channels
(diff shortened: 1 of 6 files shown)


ckipp01:
> Thanks, that looks good! Wondering why Travis CI is red (and we can't get the logs for now…)

Not fully sure what actually happened. I added a bit more to the docs, and pushed it up again. Now we are all green!

alexarchambault:
Merging then, thanks!

alexarchambault:
Forgot to ask, are you sure about the `.mkString(" ")`? I would have went for `"\n"` as a separator, but maybe `" "` is easier to read?

ckipp01:
> Forgot to ask, are you sure about the `.mkString(" ")`? I would have went for `"\n"` as a separator, but maybe `" "` is easier to read?

Hmm, at first when I tried `brew list` I saw it was spread across the screen, so I went with the space. However, now thinking about it and I just tested it with a narrow terminal, it might break in the middle of a line, which doesn't look good. I think you're right, `"\n"` would be better. One sec and I'll change it.

Issue #1624, opened 2020-03-13
COURSIER_INSTALL_DIR does not work as advertised 


Baccata:
it seems neither `COURSIER_INSTALL_DIR` nor `COURSIER_BIN_DIR` are taken into consideration. 

* coursier : 2.0.0-RC6-10
* java  "1.8.0_162"
* macOS mojave 10.14.6

``` 
oliviermelois@OM-WXRG8WN-MBP:/tmp$ cs --help
Coursier 2.0.0-RC6-10
Usage: cs [options] [command] [command-options]

Available commands: bootstrap, complete, fetch, install, java, java-home, launch, publish, resolve, setup, uninstall, update

Type  cs command --help  for help on an individual command


oliviermelois@OM-WXRG8WN-MBP:/tmp$ COURSIER_INSTALL_DIR=/tmp/random_dir/ cs install ammonite
https://repo1.maven.org/maven2/com/lihaoyi/ammonite_2.13.1/maven-metadata.xml
  No new update since 2020-01-14 04:26:58
Warning: /Users/oliviermelois/Library/Application Support/Coursier/bin is not in your PATH
To fix that, add the following line to ~/.bashrc

export PATH="$PATH:/Users/oliviermelois/Library/Application Support/Coursier/bin"
```

zheh12:
I create a pull request can solve this problem, https://github.com/coursier/coursier/pull/1672.
@alexarchambault , if you have some time, please have a look at it.

I think the problem will be more serious. For some application such as almond on mac.
For mac the default dir is ~/Library/Application Support/Coursier/bin.
And it has a space in Application Support, when I run almond --install, it will report a exception 

```
./almond --install
Exception in thread "main" java.net.URISyntaxException: Illegal character in path at index 41: file:///Users/myuser/Library/Application Support/Coursier/bin/almond
	at java.net.URI$Parser.fail(URI.java:2848)
	at java.net.URI$Parser.checkChars(URI.java:3021)
	at java.net.URI$Parser.parseHierarchical(URI.java:3105)
	at java.net.URI$Parser.parse(URI.java:3053)
	at java.net.URI.<init>(URI.java:588)
	at almond.kernel.install.Install$.$anonfun$currentAppCommand$1(Install.scala:143)
	at scala.Option.map(Option.scala:242)
	at almond.kernel.install.Install$.currentAppCommand(Install.scala:138)
	at almond.kernel.install.Install$.install(Install.scala:231)
[...]

PR #1672, opened 2020-04-19
Fix the problem COURSIER_INSTALL_DIR does not work as issue #1624

zheh12:


modules/cli/src/main/scala/coursier/cli/install/SharedInstallParams.scala
```diff
@@ -32,10 +32,22 @@ final case class SharedInstallParams(
...
+    def fromProps = Option(System.getProperty("coursier.install.dir"))
+
+    def default = coursier.paths.CoursierPaths.dataLocalDirectory().toPath.resolve("bin")
+
+    fromEnv.orElse(fromProps).map(Paths.get(_))
+      .getOrElse(default)
   }
 
+  def defaultDir: Path =
+    defaultDir0
```
alexarchambault, line 49: Let's call `coursier.install.InstallDir.defaultDir`, instead of putting the logic computing the default directory here.

modules/install/src/main/scala/coursier/install/InstallDir.scala
```diff
@@ -386,22 +386,6 @@ import scala.util.control.NonFatal
 
 object InstallDir {
 
-  private lazy val defaultDir0: Path = {
```
alexarchambault, line 389: Let's keep this logic here, it's fine to have it along `InstallDir` (in case the `install` module is used elsewhere).

modules/install/src/test/scala/coursier/install/InstallDirTests.scala
```diff
@@ -11,7 +11,7 @@ class InstallDirTests extends AnyFlatSpec with BeforeAndAfterAll {
 
     val version = "X.Y.Z"
 
-    val installDir = InstallDir()
+    val installDir = InstallDir(Paths.get(""))
```
alexarchambault, line 14: 👍

modules/cli/src/main/scala/coursier/cli/install/SharedInstallParams.scala
zheh12: Thanks, good point.

PR #1676, opened 2020-04-20
Accept inline app descriptors, and don't always run setup on Windows when run with no arguments

alexarchambault:


alexarchambault:
(Note to self: needs to be tested again on Windows.)

PR #1677, opened 2020-04-21
DependencySet optimizations

jtjeferreira:
optimizations from this flamegraph

image

*  [x] `DependencySet.<init>`
*  [x] `AbstractMap.$plus$plus$eq`
*  [x] `AbstractTraversable.toMap`


jtjeferreira:
WDYT about removing the asserts in https://github.com/coursier/coursier/pull/1677/files#diff-c44806c4f942d69f2612e6cc72b6133eL29-L30?

modules/core/shared/src/main/scala/coursier/core/DependencySet.scala
```diff
@@ -103,11 +100,11 @@ object DependencySet {
 
 
   private object Sets {
-    def empty[T]: Sets[T] = Sets(TreeMap.empty, Map.empty, Map.empty)
+    def empty[T]: Sets[T] = Sets(IntMap.empty, Map.empty, Map.empty)
   }
 
   private final case class Sets[T] private (
-    required: TreeMap[Int, Set[T]],
+    required: IntMap[Set[T]],
```
jtjeferreira, line 107: this change broke Mima, but this is a private class...

alexarchambault:
> WDYT about removing the asserts in https://github.com/coursier/coursier/pull/1677/files#diff-c44806c4f942d69f2612e6cc72b6133eL29-L30?

I guess they can be commented out, if it helps perf-wise.

modules/core/shared/src/main/scala/coursier/core/DependencySet.scala
alexarchambault: This looks like a false positive, I just pushed a commit to ignore those.

alexarchambault:
Thanks for having opened this, @jtjeferreira!

I tried to benchmark this against the latest release of coursier, and it seems it's not faster. See my results [here](https://nbviewer.jupyter.org/gist/alexarchambault/d8f2fe2314feef65a57325cbd5009c0b) (it's a notebook run with [almond](https://almond.sh), but you can plot the data any other way).

But the `IntMap` change seems legit though, going to try to benchmark this one alone.

alexarchambault:
Actually, maybe the way I ran the benchmark without these changes is to blame… (although I wasn't expecting it).

alexarchambault:
Results with more data [here](https://nbviewer.jupyter.org/gist/alexarchambault/b3423fc3704f017bceb505939f77fd61). TL;DR is IntMap brings some perf enhancements, the full set of changes of this PR brings some more. Surprisingly, running coursier like `cs launch coursier -- resolve …` is much faster than running it via the sbt-pack directory, `sbt cli/pack && modules/cli/target/pack/bin/coursier resolve …`, even though both start coursier like `java -cp …list of JARs…`.

jtjeferreira:
> > WDYT about removing the asserts in [#1677 (files)](https://github.com/coursier/coursier/pull/1677/files#diff-c44806c4f942d69f2612e6cc72b6133eL29-L30)?
> 
> I guess they can be commented out, if it helps perf-wise.

I think it will improve performance, as seen in the flamegraph. However I was thinking if there was a way to keep the assert only for development purposes, but in a PROD build they are removed... Should we handle that here or in another PR?

Regarding the benchmarks, I saw there are some jmh ones, but was not sure which ones to run, or which options to run them with, and was hoping the travis CI would run them (but I dont think thats the case)... Do you usually run them or use the `-B` option instead?

I think I have some more perf optimizations like these, and will try to submit some PRs in the next days...

PS: do you know a good way to try sbt with these changes? since sbt includes coursier as a shaded dependency, it is harder to run it with these changes...

alexarchambault:
> I think it will improve performance, as seen in the flamegraph. However I was thinking if there was a way to keep the assert only for development purposes, but in a PROD build they are removed... Should we handle that here or in another PR?

Let's just comment them out… They're useful while debugging, I don't think they would be useful if ever they were to break on coursier or sbt users' machines. It's fine doing it in this PR.

alexarchambault:
> PS: do you know a good way to try sbt with these changes? since sbt includes coursier as a shaded dependency, it is harder to run it with these changes...

It's a bit cumbersome… That can be done by
- publishing coursier locally (like with `sbt ~publishLocal`), noting the version it publishes,
- put that version in the `build.sbt` of sbt-coursier, publish it locally too, note the version it publishes,

In the project you want to test from:
- add sbt-lm-coursier as a plugin (`addSbtPlugin("io.get-coursier" % "sbt-lm-coursier" % "…")` in `project/plugins.sbt` say),
- run sbt via the [coursier-based sbt launcher](https://github.com/coursier/sbt-launcher#how-to-use)

The last step is required so that the locally-published lm-coursier-shaded library is picked upfront, rather than the one sbt depends on.

alexarchambault:
> Regarding the benchmarks, I saw there are some jmh ones, but was not sure which ones to run, or which options to run them with, and was hoping the travis CI would run them (but I dont think thats the case)... Do you usually run them or use the -B option instead?

I mostly use the `-B` option. The jmh ones where used to validate some optimizations at some point, I think they're mostly stale now.

I wish the benchmarks were run on the CI, they just aren't now.

jtjeferreira:
> Surprisingly, running coursier like `cs launch coursier -- resolve …` is much faster than running it via the sbt-pack directory, `sbt cli/pack && modules/cli/target/pack/bin/coursier resolve …`, even though both start coursier like `java -cp …list of JARs…`.

Not sure if it is related but I forked e75ca97 (i.e tag v2.0.0-RC6-2) instead of master. I did this because this was the version used in sbt...

jtjeferreira:
> > I think it will improve performance, as seen in the flamegraph. However I was thinking if there was a way to keep the assert only for development purposes, but in a PROD build they are removed... Should we handle that here or in another PR?
> 
> Let's just comment them out. They're useful while debugging, I don't think they would be useful if ever they were to break on coursier or sbt users' machines. It's fine doing it in this PR.

Done. If you could rerun your benchmarks would be awesome

alexarchambault:
Thanks! I'm getting quite noisy results right now, will try later on with less processes running on my machine…

alexarchambault:
So I ran some more benchmarks, the results are [here](https://nbviewer.jupyter.org/gist/alexarchambault/ac7e01559daa176dca2260b6aa387190). First figure is all data from the 6 runs merged. The statistics of the time to resolve spark-repl are plotted. Second figure shows the results of the individual benchmark runs. These seem on par with the current master :|

On the other hand, these results are rather noisy, and my benchmark could be biased…

@jtjeferreira How did you assess that these changes are beneficial?

The `TreeMap` -> `IntMap` change seems fine, even though I'm not sure these maps grow that big. The `assert`s don't trigger large calculations, they should be O(1).

alexarchambault:
Some [more results](https://gist.github.com/alexarchambault/34bdfce4857674949d06182b18619199), for #1683.

jtjeferreira:
> @jtjeferreira How did you assess that these changes are beneficial?

Only from the flamegraph that I posted in the PR description...

> The asserts don't trigger large calculations, they should be O(1).

They iterate over `grouped` and they account to 1/3 of the flamegraph (`DependencySet.<init>`)

jtjeferreira:
Closing because you already picked the most of the changes in #1694

PR #1691, opened 2020-04-29
Warn if a legacy cache is being used

alexarchambault:


alexarchambault:
@eed3si9n @dwijnand As a (long overdue) first step towards https://github.com/coursier/coursier/issues/1483, if the coursier cache is still at `~/.coursier/cache`, I'm planning on printing a warning like
```text
Warning: a legacy coursier cache was found at /Users/alexandre/.coursier/cache/v1 and
is currently being used.

Support for that cache location will be removed in coursier 2.0.0 final, whose release
is imminent.

Follow the instructions at
  https://github.com/coursier/cache-migration#cache-migration
in order to migrate your cache to the newer location.
```

The CLI of coursier should print it in the upcoming release, and I'm thinking about printing it in sbt, via `lm-coursier`. Do you think that will be fine? I'm thinking about just running [the same check](https://github.com/coursier/coursier/blob/8ecc8684a97c95ae516cafca8f07e7ab7f63c35e/modules/cache/jvm/src/main/scala/coursier/cache/CacheDefaults.scala#L18) when some lm-coursier get initialized.

dwijnand:
SGTM.  Good work, Alex.

PR #1683, opened 2020-04-22
Resolution perf improvments

jtjeferreira:
optimization from this flamegraph

image

PR #1694, opened 2020-04-30
Dummy optimizations

alexarchambault:


alexarchambault:
cc @jtjeferreira

This includes the less risky changes of #1677 and #1683.

jtjeferreira:
Dont you think the change from https://github.com/coursier/coursier/pull/1677/files#diff-c44806c4f942d69f2612e6cc72b6133eR56-R62 is safe?

From the flamegraph in #1677 we can see 1/3 of the traces are from `m ++= grouped` and another 1/3 are from `m.toMap`; another 1/3 is from the removed asserts...

jtjeferreira:
> I'm not sure. IIRC, I actually went from immutable maps to mutable ones while trying to optimize things, using async-profiler to find hot spots, and validating gains with the same kind of benchmarks as in #1677, using the -B option of the coursier CLI.

This is what the old code does: 
1) initializing an empty mutable map `m`
2) add all elements of `grouped` to the map (1/3 of the flamegraph in `AbstractMap.$plus$plus$eq`)
3) update elements of map `m`
4) convert `m` to an immutable Map (1/3 of the flamegraph in `AbstractTraversable.toMap`)

My code does the following:
1) use existing immutable map `grouped`
2) update elements of immutable map `m` and update var `m` to the result of that copy 
3) return m

My intuition would say that my code would be better, but I have to admit I have no evidence/benchmarks that is faster. I will try get some benchmarks for that...

alexarchambault:
> 2. update elements of immutable map m and update var m to the result of that copy

"updating" the elements of an immutable map isn't free. It can re-use bits of the former map, and create some new internal stuff. Atomic updates such as those are usually cheaper in mutable collections.

But I'd be ok adding changing that code, if some benchmarks show that it's a gain.

alexarchambault:
(Merging the PR here, in the mean time.)

PR #1704, opened 2020-05-06
Delete lockfile explicitly

jodersky:
This partially reverts the changes introduced in 344f6e73c10cf161c25a5878d6db9cce470e02b0 (between 2.0.0-RC6-1 and 2.0.0-RC6-2). Specifically, it reverts back to explicitly deleting lockfiles from caches, rather than relying on the DELETE_ON_CLOSE channel option.

It solves a concurrent download bug in mill and unblocks this PR https://github.com/lihaoyi/mill/pull/858. I'm not exactly sure why this change fixes the issue, but this is what I found after many hours of bisecting. Note that it is not sufficient to explicitly delete the file in the finally block, but that the DELETE_ON_CLOSE must not be used as it somehow interferes.

### DIFF
--- a/modules/cache/jvm/src/main/scala/coursier/cache/CacheLocks.scala
+++ b/modules/cache/jvm/src/main/scala/coursier/cache/CacheLocks.scala
@@ -41,8 +41,7 @@ object CacheLocks {
       channel = FileChannel.open(
         lockFile,
         StandardOpenOption.CREATE,
-        StandardOpenOption.WRITE,
-        StandardOpenOption.DELETE_ON_CLOSE
+        StandardOpenOption.WRITE
       )
     }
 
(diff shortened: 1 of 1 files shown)


jodersky:
[Here](https://github.com/coursier/coursier/commit/344f6e73c10cf161c25a5878d6db9cce470e02b0#diff-4c452019718e1f3f2a82708ab60d939e) is the direct diff of the file that somehow causes the bug in mill.

jodersky:
Thanks for merging! I'd be curious to find out what exactly was causing this, but I just lost the motivation after chasing this. It was quite the adventure to find :)

alexarchambault:
The `DELETE_ON_CLOSE` javadoc says "This option is not recommended for use when opening files that are open concurrently by other entities." Maybe I should have paid a closer attention to that before using it :|

Issue #1618, opened 2020-03-11
Scala.js runtime error: require is not a function


MaximeKjaer:
## Problem

In Scala.js, Coursier's NPM dependencies are loaded through `js.Dynamic.global.require("...")`. This is problematic for two reasons:

- Incompatibility with ES modules. 
  
  Scala.js projects depending on Coursier should have the choice between generating ES module `import`s or CommonJS `require`s, but the current implementation of Coursier contains hard-coded calls to `require`. This means that Scala.js  projects targeting ES modules will also generate `require` calls, which won't work in the browser (unless the bundle is passed through a bundler like Webpack).
  
- `global.require` may be `undefined` depending on how the JS code is loaded.
  
  For instance, running `console.log(global.require)` prints `undefined` in a JS file (loaded using `node test.js`) but prints the `require` function in a Node.js REPL. See [this StackOverflow question](https://stackoverflow.com/q/34566343/918389).

These two reasons both lead to the same runtime error: `TypeError: $g.require is not a function`. 

## Solution

There are two possible solutions for this:

1. Use `@JSImport` over `js.Dynamic.global.require`.
2. Update to Scala.js 1.x and use `@JSGlobal` to access `require` instead of `global.require`.

I think solution 1 is probably preferable to solution 2, as Scala.js projects can then target the module system they need. While I think Coursier should eventually move to Scala.js 1.x, solution 2 requires getting all of Coursier's dependencies to switch to 1.x, which may take time.

## Workaround until the solution is implemented

If anyone else needs to work around the issue for now, the hack I've found is to use Webpack to bundle my Scala.js output with a JS file containing `eval("global.require = require")`. Note that you should then very carefully determine if the Coursier dependencies (xhr2, cheerio, xmldom and sax) should be bundled too (for the Web) or just be available in `node_modules` (for Node.js).

alexarchambault:
Thanks for opening this! I agree solution 1. is fine.

About Scala.JS 1.0, we're only waiting for [scalajs-react](https://github.com/japgolly/scalajs-react) to support it to switch to it.

About bundling dependencies or not, `xhr2`, `cheerio`, `xmldom`, and `sax` are all only needed / used from Node.

alexarchambault:
If you feel like fixing this @MaximeKjaer, go ahead!

MaximeKjaer:
I'm up for it! I'll give it a try.

MaximeKjaer:
[...]

PR #1715, opened 2020-05-12
Update scalajs-react / react / Scala.JS

alexarchambault:


modules/core/js/src/main/scala/coursier/core/compatibility/package.scala
```diff
@@ -34,10 +44,10 @@ package object compatibility {
   lazy val DOMParser = newFromXmlDomOrGlobal("DOMParser")
   lazy val XMLSerializer = newFromXmlDomOrGlobal("XMLSerializer")
   lazy val sax =
-    if (js.isUndefined(g.sax))
-      g.require("sax")
+    if (js.isUndefined(globalObject.sax))
+      globalObject.require("sax")
     else
-      g.sax
+      globalObject.sax
```
sjrd, line 50: If you do things this way, you'll *still* have the problem that "`require` is not defined". You're going out of your way to reproduce the behavior of Scala.js 0.6.x, which is the cause of #1618.

You should take full advantage of the behavior of Scala.js 1.x instead, as follows:
```scala
lazy val sax =
  if (js.typeOf(g.sax) == "undefined")
    g.require("sax")
  else
    g.sax
```
Likewise, you should inline `newFromXmlDomOrGlobal` at the two places where it's used to be able to use the same pattern:
```scala
lazy val DOMParser = {
  val defn =
    if (js.typeOf(g.DOMParser) == "undefined") g.require("xmldom").DOMParser
    else g.DOMParser
  js.Dynamic.newInstance(defn)()
}
```
and copy-paste for `XMLSerializer`.

If you can't stand the duplication, you can abstract away *some* of it as follows:
```scala
def newFromXmlDomOrGlobal(typeOf: String, globalDefn: => js.Dynamic, xmldomName: String) = {
  val defn =
    if (typeOf == "undefined") g.require("xmldom").selectDynamic(xmldomName)
    else globalDefn
  js.Dynamic.newInstance(defn)()
}
```
which you can use as
```scala
lazy val DOMParser = newFromXmlDomOrGlobal(js.typeOf(g.DOMParser), g.DOMParser, "DOMParser")
lazy val XMLSerializer = newFromXmlDomOrGlobal(js.typeOf(g.XMLSerializer), g.XMLSerializer, "XMLSerializer")
```

modules/tests/js/src/test/scala/coursier/Platform.scala
```diff
@@ -15,12 +15,21 @@ object Platform {
...
+
   /** Available if we're running on node, and package xhr2 is installed */
   lazy val xhr =
-    if (js.isUndefined(g.XMLHttpRequest))
-      g.require("xhr2")
+    if (js.isUndefined(globalObject.XMLHttpRequest))
+      globalObject.require("xhr2")
     else
-      g.XMLHttpRequest
+      globalObject.XMLHttpRequest
```
sjrd, line 32: Of course, the same story applies here, and everywhere else where this pattern comes up.

modules/util/js/src/main/scala/coursier/util/WebPageCompatibility.scala
```diff
@@ -6,10 +6,19 @@ import scala.scalajs.js.Dynamic.{global => g}
...
+      g.global
+    else
+      // In all other well-known environment, we can use the global `this`
+      js.special.fileLevelThis.asInstanceOf[js.Dynamic]
+
   // FIXME Won't work in the browser
   lazy val cheerio = g.require("cheerio")
 
-  lazy val jqueryAvailable = !js.isUndefined(g.$)
+  lazy val jqueryAvailable = !js.isUndefined(globalObject.$)
```
sjrd, line 21:
```suggestion
  lazy val jqueryAvailable = js.typeOf(g.$) != "undefined"
```

modules/util/js/src/main/scala/coursier/util/WebPageCompatibility.scala
```diff
@@ -18,7 +27,7 @@ private[coursier] abstract class WebPageCompatibility {
     // getting weird "maybe a wrong Dynamic method signature" errors when trying to factor that more
 
     if (jqueryAvailable)
-        g.$("<div></div>").html(page).find("a").each({ self: js.Dynamic =>
+        globalObject.$("<div></div>").html(page).find("a").each({ self: js.Dynamic =>
```
sjrd, line 30:
```suggestion
        g.$("<div></div>").html(page).find("a").each({ self: js.Dynamic =>
```

alexarchambault:
Thanks for the review, @sjrd!

PR #1759, opened 2020-06-03
Use new coursier-version library

alexarchambault:


alexarchambault:
(Will add back removed classes to preserve binary compatibility)

alexarchambault:
Way to out-of-date.

PR #1760, opened 2020-06-04
fix installing zsh completion

darklore:
Hello.
There is a typo bug so that we can't install zsh completion through homebrew.
(zsh-completion => zsh-completions)

### DIFF
--- a/scripts/coursier.rb.template
+++ b/scripts/coursier.rb.template
@@ -21,14 +21,14 @@ class Coursier < Formula
   depends_on :java => "1.8+"
 
   def install
-    unless build.without? "zsh-completion"
-      FileUtils.mkdir_p "completions/zsh"
-      system "bash", "-c", "bash ./coursier --completions zsh > completions/zsh/_coursier"
-      zsh_completion.install "completions/zsh/_coursier"
-    end
-
     bin.install 'cs-x86_64-apple-darwin' => "cs"
     resource("jar-launcher").stage { bin.install "coursier" }
+
+    unless build.without? "zsh-completions"
+      chmod 0555, bin/"coursier"
+      output = Utils.popen_read("#{bin}/coursier --completions zsh")
+      (zsh_completion/"_coursier").write output
+    end
   end
 
   test do


scripts/coursier.rb.template
```diff
@@ -21,14 +21,14 @@ class Coursier < Formula
...
-      system "bash", "-c", "bash ./coursier --completions zsh > completions/zsh/_coursier"
-      zsh_completion.install "completions/zsh/_coursier"
-    end
-
     bin.install 'cs-x86_64-apple-darwin' => "cs"
     resource("jar-launcher").stage { bin.install "coursier" }
+
+    unless build.without? "zsh-completions"
+      chmod 0555, bin/"coursier"
+      output = Utils.popen_read("#{bin}/coursier --completions zsh")
```
alexarchambault, line 29: Using `cs` instead of `coursier` ought to work here (`cs` is a native application, so it should be faster).

darklore: Thank you for reviewing.
I tried using `cs` but it fails with exception.
```
$ cs --completions zsh
Exception in thread "main" java.lang.NullPointerException
        at coursier.cli.Coursier$.zshCompletions(Coursier.scala:51)
        at coursier.cli.Coursier$.$anonfun$beforeCommand$4(Coursier.scala:104)
        at scala.Option.foreach(Option.scala:407)
        at coursier.cli.Coursier$.beforeCommand(Coursier.scala:102)
        at coursier.cli.Coursier$.beforeCommand(Coursier.scala:27)
        at caseapp.core.app.CommandAppWithPreCommand.$anonfun$main$2(CommandAppWithPreCommand.scala:82)
        at caseapp.core.app.CommandAppWithPreCommand.$anonfun$main$2$adapted(CommandAppWithPreCommand.scala:82)
        at scala.util.Either.fold(Either.scala:191)
        at caseapp.core.app.CommandAppWithPreCommand.main(CommandAppWithPreCommand.scala:82)
        at coursier.cli.Coursier$.main(Coursier.scala:78)
        at coursier.cli.Coursier.main(Coursier.scala)
```

alexarchambault: Mmmh, I thought I had fixed that (but yeah, I confirm it still fails…)

alexarchambault: https://github.com/coursier/coursier/pull/1765 should fix it in the next release.

alexarchambault: Let's merge this in the mean time.

alexarchambault:
Merging, thanks @darklore!

PR #1769, opened 2020-06-12
Remove unused pants-related files

alexarchambault:
The pants build has been disabled on the CI for some time now, and I'm not sure these files are necessary anymore.


alexarchambault:
cc @wisechengyi @olafurpg @cosmicexplorer

You're not building coursier launchers with these files anymore, right?

wisechengyi:
Ah thanks! Right, historically we needed the fatjar for the CLI, but now they are published on github already.

Issue #1675, opened 2020-04-20
Coursier could cache jar validation more agressively


jtjeferreira:
Related to https://github.com/sbt/sbt/issues/5508 where I described a problem with sbt+coursier in a multi module scenario. Basically running `sbt update` in a multi module project takes a lot of time because coursier is validating the same dependencies several times (once per module AFAICT)

One of the approaches to fix this problem would be to cache this more aggressively in courier and skip validation of already validated files...

alexarchambault:
We should cache checksum validations somehow, yes. I've been wanting to find a way not to move a file to its final destination in cache if it has invalid checksums for some time, but couldn't find a convenient way to do it.

Writing somewhere that we validated a checksum could work, or better, persist our calculated checksums, and only compare those to the downloaded ones. So that we only do the comparison later on, not recomputing the checksum itself.

oyvindberg:
So if I want to try my hand at this, where does it make sense to put the files? side by side with the downloaded files? or in a parallel hierarchy within `$COURSIER_CACHE`?

alexarchambault:
@oyvindberg There's already a concept of ["auxiliary files"](https://github.com/coursier/coursier/blob/84d20de788ffc09b3cac2f491e213c48c97bea9a/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala#L953) in `FileCache`, which correspond to files kept alongside another one. For example, if we get a SHA-1 via an HTTP header when downloading `foo/name`, it's kept in `foo/.name.sha1`. This particular auxiliary file is written around [here](https://github.com/coursier/coursier/blob/84d20de788ffc09b3cac2f491e213c48c97bea9a/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala#L399). When a changing artifact (snapshots typically) is re-downloaded, its auxiliary files are [wiped](https://github.com/coursier/coursier/blob/84d20de788ffc09b3cac2f491e213c48c97bea9a/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala#L943) .

Checksums are currently computed around [here](https://github.com/coursier/coursier/blob/84d20de788ffc09b3cac2f491e213c48c97bea9a/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala#L685-L693). You may want to write an auxiliary file here (with an extension along the lines of `.sha1.computed` for example), or read an existing one.

[...]

PR #1801, opened 2020-07-14
Cache computed digests in auxiliary files along with downloaded files #1675

oyvindberg:
So this is a first take. I did try it locally when developing it, and it did shave a few seconds off resolve time for the benchmark project.

I have a few doubts about this implementation:
- It currently writes binary files, which is unnoticeably faster and somewhat easier to write, but they cant be `cat`ed on the command line. I'll change to a textual format if you think it's a problem
- I wanted to look closer at performance before submitting, but like you said it shouldn't make things slower. I went with an exception-based approach instead of something like `if (exists(cachedFile)) readFile()`. Performing IO twice in the common path (cached files exists) seemed slower in the profiler. There is quite a bit of overhead also from reading all the small files, an improvement down the line might be to save all computed digests in a central place.
- You'll be a better judge of this than me, but I tried a lockless approach with atomic files moves, where last thread to compute the digests "wins". As far as I can see this should be fast and safe, but I'll add back locking if you think it's necessary.

Take this from here however you like, you can brush up the code after merge or I'll do it if you point out what you don't like.

### DIFF
--- a/modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala
+++ b/modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala
@@ -3,7 +3,8 @@ package coursier.cache
 import java.io.File
 import java.net.URI
 import java.nio.charset.StandardCharsets
-import java.nio.file.Files
+import java.nio.file.{Files, Paths}
+import java.util
 
 import cats.effect.IO
 import coursier.cache.TestUtil._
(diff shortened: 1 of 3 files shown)


alexarchambault:
Nice, thanks! Latest win could work yes, given that it's not much of a problem if ever this computation is triggered more than once for a (short) period of time.

I'm not sure things should be written in a file provided by `File.createTempFile` though. It may be in a different partition than the cache, so that an atomic move would not be possible. Better write it in the same directory as the destination.

I think the case where the move fails (because the destination exists now) needs to be handled too. I'll look into it. Thanks for having opened that!

oyvindberg:
> Latest win could work yes, given that it's not much of a problem if ever this computation is triggered more than once for a (short) period of time.

I think it'll be fine, this is a very short window where worst case we revert to current performance

> I'm not sure things should be written in a file provided by File.createTempFile though

Good catch. if you keep it without locking better generate unique temporary names then, that's why I went with that API in the first place

> I'll look into it. 
Tell me if I should do something more here :)

alexarchambault:
> Tell me if I should do something more here :)

Oh, sorry, I had misread the end of your first comment above… Go ahead if you want to address the `File.createTempFile` use and the handling of the `Files.move` exception (if the target already exists)!

This could also be unit tested. Some tests could be added around [here](https://github.com/coursier/coursier/blob/00c80111f17d8930da7793e4c95d69ff7ece6667/modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala#L889), using the same fixtures ([`data/foo.xml` and its checksums](https://github.com/coursier/coursier/tree/00c80111f17d8930da7793e4c95d69ff7ece6667/modules/cache/jvm/src/test/resources/data)). We could look in the directory of the `java.io.File` provided by `FileCache`, manually edit the `.sha1.computed` say, and check that new attempts fail ([`FileCache.retry`](https://github.com/coursier/coursier/blob/00c80111f17d8930da7793e4c95d69ff7ece6667/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala#L40) may need to be set to 0 for that, with a `.withRetry(0)`).

oyvindberg:
Sorry, vacation time means I didn't have time to follow up. I'll try to work through your comments within the next few days. I thought I had checked that `Files.move` would just silently overwrite (that's why I wrote last thread wins), but I'll verify again.

oyvindberg:
Ok, so I think I addressed your comments now.

- I kept using `File.createTempFile`, but an overload which puts files with a unique name in the specified folder. the temp file is adjacent to all the other files
- You were right about `Files.move` throwing an exception. I changed it so first thread now wins, and just delete the temp file in any other case.
- the implementation ended up using more `try` than I've written in several years worth of scala code. If you're uncomfortable with that I can try to do one more pass and change it around a bit.
- the test case for wrong stored digest should probably be refactored (I left a todo) to get a fresh directory for each run. The actual test is currently drowning in setup to make it all work
- `Files.move` can throw `AtomicMoveNotSupportedException` - should we try to recover from that with a non-atomic move?

I'll do some proper testing of it tomorrow, otherwise it should all be in place now

modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala
```diff
@@ -1022,6 +1023,133 @@ object FileCacheTests extends TestSuite {
...
+      * - async {
+        val Right(file: File) = await {
+          FileCache()
+            .withChecksums(Seq(Some("SHA-1")))
+            .file(artifact)
+            .run
+            .future()
+        }
+
+        Files.write(computedPath, Array(1: Byte, 2: Byte, 3: Byte))
```
alexarchambault, line 1125: NIT:
```suggestion
        Files.write(computedPath, Array[Byte](1, 2, 3))
```

modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala
```diff
@@ -1022,6 +1023,133 @@ object FileCacheTests extends TestSuite {
...
+            val computedPath = Paths.get(s"${file.getParent}/.${file.getName}__md5.computed")
+            val expected = stringToByteArray("001717e73bca14e4fb2df3cabd6eac98")
+            val actual = Files.readAllBytes(computedPath)
+            assert(util.Arrays.equals(actual, expected))
+          case Left(e) => throw e
+        }
+      }
+    }
+
+    // todo: should probably refactor so we get a fresh folder for each test run
```
alexarchambault, line 1084: Indeed, same for the additional tests above.

oyvindberg:
Ok @alexarchambault , this PR is updated now. Rebased on master and I updated the tests as discussed some months back.
Cursory testing seems that everything works, and that we shave off a few seconds of `update` time.

oyvindberg:
@alexarchambault can we move forward with this?

alexarchambault:
@oyvindberg It seems the CI error is legit… I should have time to investigate later this week if needed.

oyvindberg:
I should have fixed those tests now, also rebased on master again.

The test failures were real, and showcased exactly the effect of the change in this PR. If a file is corrupted after we have computed its checksum, coursier will no longer notice. I updated the tests (see last commit) to instead corrupt the computed checksums.

alexarchambault (approved):
Looks good, thanks for having pushed this through, @oyvindberg!

PR #1827, opened 2020-08-18
Add in license parsing to `PomParser`

ckipp01:
I've had an idea I've been playing around with lately about using Coursier for license reporting. I noticed the Ivy parsing seemed to be there, but not the Pom. This adds in license parsing to the PomParser.

~One thing to note is that currently in `Info` licenses are modeled as `(String, Option[String])` since in Ivy there is just a name and url. In Maven it seems that there can also be a distribution and comments. Do you think those should be added in right away and licenses changed in `Info` to a case class like:~

☝️ All added now.

Let me know your thoughts!

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
@@ -257,13 +257,60 @@ object Attributes {
 @data class Info(
   description: String,
   homePage: String,
-  licenses: Seq[(String, Option[String])],
   developers: Seq[Info.Developer],
   publication: Option[Versions.DateTime],
-  scm: Option[Info.Scm]
-)
+  scm: Option[Info.Scm],
+  licenseInfo: Seq[Info.License]
+) {
+  def licenses: Seq[(String, Option[String])] = licenseInfo.map(li => li.name -> li.url)
+
+  def withLicenses(license: Seq[(String, Option[String])]) = {
+    new Info(
+      description,
+      homePage,
+      developers,
+      publication,
+      scm,
+      license.map(l => Info.License(l._1, l._2, None, None))
+    )
+  }
+
+  def this(
+    description: String,
+    homePage: String,
+    licenses: Seq[(String, Option[String])],
+    developers: Seq[Info.Developer],
+    publication: Option[Versions.DateTime],
+    scm: Option[Info.Scm]
+  ) = {
+    this(
+      description,
+      homePage,
+      developers,
+      publication,
+      scm,
+      licenses.map(l => Info.License(l._1, l._2, None, None))
+      )
+  }
+}
 
 object Info {
+  def apply(
+    description: String,
+    homePage: String,
+    licenses: Seq[(String, Option[String])],
+    developers: Seq[Info.Developer],
+    publication: Option[Versions.DateTime],
+    scm: Option[Info.Scm]
+  ): Info = new Info(
+    description,
+    homePage,
+    developers,
+    publication,
+    scm,
+    licenseInfo = licenses.map(l => License(l._1, l._2, None, None))
+  )
+
   @data class Developer(
     id: String,
     name: String,
@@ -276,7 +323,14 @@ object Info {
     developerConnection: Option[String]
   )
 
-  val empty = Info("", "", Nil, Nil, None, None)
+  @data class License(
+    name: String,
+    url: Option[String],
+    distribution: Option[String], // Maven-specific
+    comments: Option[String] // Maven-specific
+  )
+
+  val empty = Info("", "", Nil, None, None, Nil)
 }
 
 // Maven-specific
--- a/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
@@ -290,7 +290,7 @@ object Attributes {
       publication,
       scm,
       licenses.map(l => Info.License(l._1, l._2, None, None))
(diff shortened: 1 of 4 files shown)


alexarchambault:
> Do you think those should be added in right away and licenses changed in Info to a case class like:

These can definitely be added. (This would be a `@data class License` rather than a `case class License`.)

I think the only tricky point would be maintaining binary compatibility for [`Info.licenses`](https://github.com/coursier/coursier/blob/5cb7be37f7fd29c3177c9fc22498b9ca064b0061/modules/core/shared/src/main/scala/coursier/core/Definitions.scala#L260). It should be possible to
- remove `licenses`
- add a new `licenseDetails` say (or any other name), and 
- add
  - a compatibility constructor,
  - an `apply` method, still accepting licenses as `Seq[(String, Option[String])]`, and
  - a `def licenses: Seq[(String, Option[String])] = licenseDetails.map(…)`.

By adding `licenseDetails` in last position, the compatibility constructor and `apply` method shouldn't conflict with the generated ones.

alexarchambault:
I can take care of adding the compatibility stuff at the end, if you'd like.

ckipp01:
> I can take care of adding the compatibility stuff at the end, if you'd like.

I'll take a stab

> ~remove licenses~
> ~add a new licenseDetails say (or any other name), and~
>    add
>     a compatibility constructor,
>     ~an apply method, still accepting licenses as Seq[(String, Option[String])], and~
>     ~a def licenses: Seq[(String, Option[String])] = licenseDetails.map(…).~

So I think the only think I may be missing is compatibility constructor, but I'm a bit unsure what you mean by this. Could you elaborate? Thanks

alexarchambault:
@ckipp01 The CI [points at the missing methods / constructors](https://travis-ci.org/github/coursier/coursier/jobs/719186365#L2664-L2669): it wants a `withLicenses` method accepting licenses with the former type (this method was generated from the `licenses` field by data-class), and a constructor (a `def this(…)`) with the same signature as the former one.

modules/core/shared/src/main/scala/coursier/maven/PomParser.scala
```diff
@@ -95,7 +95,12 @@ object PomParser {
 
     var description = ""
     var url = ""
-    val licenses = Nil // TODO
+    var licenses = Nil // Left in for binary compat
```
ckipp01, line 98: So it was complaining about this as well :/ Is it true that since this was inferred previously as a `scala.collection.immutable.Nil.type` it now breaks binary compatibility if `PomParser` doesn't have a licenses filed that returns a `Nil.type`?

ckipp01:
Alright, thanks for the pointers. I haven't worked with Mima before, so it was cool to dive into that some. Everything looks happy now.

modules/core/shared/src/main/scala/coursier/core/Definitions.scala
```diff
@@ -257,13 +257,60 @@ object Attributes {
...
+  }
+
+  def this(
+    description: String,
+    homePage: String,
+    licenses: Seq[(String, Option[String])],
+    developers: Seq[Info.Developer],
+    publication: Option[Versions.DateTime],
+    scm: Option[Info.Scm]
+  ) {
```
regadas, line 285: 🎨 procedure syntax

```suggestion
  ) = {
```

ckipp01: 😬

modules/core/shared/src/main/scala/coursier/maven/PomParser.scala
alexarchambault: Oh, I wanted to mention it too: this class is private, so you can just add the filter printed by `mimaReportBinaryIssues` in [`Mima.scala`](https://github.com/coursier/coursier/blob/85df5e041156cb89105ab3ac8d89f098adce1daa/project/Mima.scala#L74).

modules/core/shared/src/main/scala/coursier/core/Definitions.scala
```diff
@@ -257,13 +257,60 @@ object Attributes {
...
+    scm: Option[Info.Scm]
+  ) = {
+    this(
+      description,
+      homePage,
+      developers,
+      publication,
+      scm,
+      licenses.map(l => Info.License(l._1, l._2, None, None))
+      )
```
alexarchambault, line 293: nit: indentation
```suggestion
    )
```

modules/core/shared/src/main/scala/coursier/maven/PomParser.scala
```diff
@@ -95,7 +95,12 @@ object PomParser {
 
     var description = ""
     var url = ""
-    val licenses = Nil // TODO
+    val licenses = Nil // Left in for binary compat
```
alexarchambault, line 98: See my comment above, this can be removed, and corresponding MiMa filters can be added in `project/Mima.scala`.
```suggestion
```

modules/core/shared/src/main/scala/coursier/maven/PomParser.scala
ckipp01: Ah, ok cool. Added.

modules/core/shared/src/main/scala/coursier/core/Definitions.scala
ckipp01: Fixed.

alexarchambault (approved):

alexarchambault:
Merging, thanks @ckipp01!

PR #1834, opened 2020-08-26
Handle -J-D* options from native launcher too

alexarchambault:


dwijnand:
> Handle -J-D* options ourselves from native launcher
> So that users can set Java properties manually.
> 
> native images accept -D* options out-of-the-box, but we disabled that so
> that -D* options are processed the same way on the JVM and from the
> native launchers.

I don't understand what this means because I can't get -J-D (or just -D?) to work:

```
$ cs launch scala -Dscala.color -Dfoo=bar
Unrecognized argument: -Dscala.color
Unrecognized argument: -Dfoo=bar
$ cs launch scala -- -Dscala.color -Dfoo=bar
Welcome to Scala 2.13.3 (OpenJDK 64-Bit Server VM, Java 11.0.8).
Type in expressions for evaluation. Or try :help.

scala> sys.props.contains("foo")
val res0: Boolean = false

scala> sys.props.contains("scala.color")
val res1: Boolean = false
```

This change broke scala-runners, btw, as I'd suspected https://github.com/dwijnand/scala-runners/issues/26

dwijnand:
Sorry I didn't test out v2.0.0-RC6-27 before it became 2.0.0 😞

PR #1849, opened 2020-09-13
Add aarch64 and 32bit Arm to supported architectures

matsluni:
This adds `aarch64` as supported architectures in the JvmIndex.

Are there any tests I should add?

### DIFF
--- a/modules/jvm/src/main/scala/coursier/jvm/JvmIndex.scala
+++ b/modules/jvm/src/main/scala/coursier/jvm/JvmIndex.scala
@@ -82,6 +82,7 @@ object JvmIndex {
   lazy val currentArchitecture: Either[String, String] =
     Option(System.getProperty("os.arch")).map(_.toLowerCase(Locale.ROOT)) match {
       case Some("x86_64" | "amd64") => Right("amd64")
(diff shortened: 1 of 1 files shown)


matsluni:
Some additional thoughts on this:
Currently, this only supports 64bit architectures of ARM devices returning `aarch64` for `System.getProperty("os.arch")`. What do you think of also supporting ARM devices on 32bit hardware, like older Raspberry Pis?

Also, the `aarch64` is returned by a Radxa Rock Pi 4 (Raspberry Pi clone) on Armbian buster. I am not 100% sure if a Raspberry Pi 4 would return the same value.

alexarchambault:
> Are there any tests I should add?

It's not needed here, I think.

> What do you think of also supporting ARM devices on 32bit hardware, like older Raspberry Pis?

It could be added, yes.

matsluni:
I could confirm that a RaspberryPi 4 also returns `aarch64` for `System.getProperty("os.arch")`. 
Also, I checked my old RaspberryPi 3 with 32 bit and it returned just `arm`.

alexarchambault (approved):
Thanks!

PR #1875, opened 2020-10-02
Fix some tests on Windows

martijnhoekstra:
get started with making tests pass on Windows


alexarchambault:
Thanks @martijnhoekstra! Don't hesitate to rebase to actually run the CI, the GitHub actions config was invalid on master to fix issues during the latest release (I just fixed it).

modules/jvm/src/test/scala/coursier/jvm/JvmCacheTests.scala
```diff
@@ -66,6 +74,14 @@ object JvmCacheTests extends TestSuite {
           |        "1.2": "tgz+https://foo.com/download/the-jdk-1.2.tar.gz"
           |      }
           |    }
+          |  },
+          |  "windows": {
+          |    "the-arch": {
+          |      "jdk@the-jdk": {
```
martijnhoekstra, line 80: I added some java dummy bat files in these archives, but maybe it's better to do these tests in terms of just reading the files content so it won't be platform dependent whether it'll run or not?

alexarchambault: It's fine this way I think. It's a good thing to actually run things on Windows.

alexarchambault (approved):
Thanks @martijnhoekstra! Would you mind rebasing your commits on the latest master, so that we get a clean history upon merge?

martijnhoekstra:
yeah sure. I'll rebase shortly! There is still a lot of tests failing, and I intend to clean up a bit more. I don't mind doing it in separate PRs or collecting more along the way.

alexarchambault:
We can do as you wish. The earlier we merge things, the less chance there is of conflicts in the future.

modules/jvm/src/test/scala/coursier/jvm/JvmCacheTests.scala
martijnhoekstra: Agreed on that it's good to run things on windows. In the glorious future, that might be in separate tests, but at least we have something that can make noise when it stops running now.

martijnhoekstra:
closed in favor of #1890

PR #1885, opened 2020-10-07
Fix doc of Launch command as per changes in `-J` option

shubhamJay:
With changes in launch command options, `-J` is no longer used to provide java option to forked JVM.


alexarchambault (approved):
Indeed, thanks!

PR #1892, opened 2020-10-09
Update docs for passing java options to launchers installed via cs in…

Poorva17:
…stall cli command


Poorva17:
@alexarchambault I have added documentation for passing java arguments to launchers installed using `cs install`. We are using this for our applications. I could not found documentation for this. So added in this PR

alexarchambault:
That's useful, thanks @Poorva17!

I pushed an extra commit to move the section up on the page, and also mention `JAVA_OPTS`.

PR #1897, opened 2020-10-13
Fix zsh completions

note:
Change zsh completions: "coursier" -> "cs"

Add new commands to zsh completions

### DIFF
--- a/doc/docs/cli-installation.md
+++ b/doc/docs/cli-installation.md
@@ -97,14 +97,14 @@ Install it via `pkg` from the [Ports Collection](https://www.freshports.org/deve
 $ pkg install coursier
 ```
 
-### zsh completions
+## zsh completions
 
 If you use ZSH, simple tab-completions are available by writing the
 coursier completion data into your completions directory.
-You can install the completions with,
+You can install the completions with:
 ```bash
 mkdir -p ~/.zsh/completion
-cs --completions zsh > ~/.zsh/completion/
+cs --completions zsh > ~/.zsh/completion/cs
 echo 'fpath=(~/.zsh/completion $fpath)' >> ~/.zshrc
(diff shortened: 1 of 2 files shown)


modules/cli/src/main/resources/completions/zsh
```diff
@@ -54,39 +59,35 @@ _coursier() {
...
 
-    launch)
-      _arguments -C $mainClass $extraJars $isolated $isolateTarget $keepOptional $mode $ttl $quiet $verbose $progress $maxIterations $repository $sourcesR $noDefault $sbtPluginHack $dropInfoAttr $forceVersion $exclude $scalaVersion $intransitive $classifier $defaultConfiguration $parallel $checksum $benchmark $tree $reverseTree $profile $typelevel $cache
-      ;;
-    
     fetch)
       _arguments -C $sourcesS $javadoc $classpath $artifactType $force $keepOptional $mode $ttl $quiet $verbose $progress $maxIterations $repository $sources $noDefault $sbtPluginHack $dropInfoAttr $forceVersion $exclude $scalaVersion $intransitive $classifier $defaultConfiguration $parallel $checksum $benchmark $tree $reverseTree $profile $typelevel $cache
       ;;
+
+    install)
```
note, line 79: I could've added parameters completions also for other commands. Since it's a tedious work I decided to go with partial change

doc/docs/cli-installation.md
```diff
@@ -97,14 +97,14 @@ Install it via `pkg` from the [Ports Collection](https://www.freshports.org/deve
 $ pkg install coursier
 ```
 
-### zsh completions
+## zsh completions
```
note, line 100: Previously it was a subsection of `JAR-based launcher` which was a bit misleading. That way it's IMHO clearer

alexarchambault (approved):
Thanks @note! (And sorry this slipped my attention.)

Issue #1786, opened 2020-07-02
Command to deal with channels


shubhamJay:
Adding and using channels feels a bit confusing now as some commands support `—add-channel`  option, and some don’t. Even though the once which does not allow adding channels, use the channels which are previously added by others commands. For example. `launch` command does not support `—add-channel` option, but it uses channels added by others commands such as `install` command.

  May be having a dedicated command like `channel` which will allow operations such as add channel or remove channel (Not sure that remove channel exists)
Something like: 
```
 cs channel —add <channel> 
cs channel —remove <channel>
```

alexarchambault:
@shubhamJay I agree a `channel` command would be a nice addition, and a better substitute to the `--add-channel` option of `install`.

PRs welcome! https://github.com/coursier/coursier/pull/1780 adds an extra command, as an example.

kubukoz:
Looks like `channel --add` already exists, so this issue is now limited to a missing `--remove`, right?

PR #1899, opened 2020-10-18
Channel command to deal with channels

shubhamJay:
Alternative for `cs install --add-channel` as discussed in issue #1786 .

### DIFF
--- a/modules/cli/src/main/scala/coursier/cli/install/Install.scala
+++ b/modules/cli/src/main/scala/coursier/cli/install/Install.scala
@@ -7,6 +7,7 @@ import java.time.Instant
 
 import caseapp.core.app.CaseApp
 import caseapp.core.RemainingArgs
+import coursier.cli.channel.ChannelCommand
 import coursier.cli.setup.MaybeSetupPath
 import coursier.cli.Util.ValidatedExitOnError
 import coursier.install.{Channels, InstallDir, RawSource}
@@ -45,10 +46,11 @@ object Install extends CaseApp[InstallOptions] {
       .withNativeImageJavaHome(Some(graalvmHome))
 
     if (params.installChannels.nonEmpty) {
+      ChannelCommand.addChannel(params.installChannels.toList,params.output)
+
       val progName = coursier.cli.Coursier.progName
       val options = params.installChannels.flatMap(c => Seq("--add", c)).mkString(" ")
       System.err.println(s"Warning: the --add-channel option is deprecated. Use '$progName channel $options' instead.")
-
     } else if (params.env.env)
       println(installDir.envUpdate.script)
     else if (params.env.disableEnv) {
--- a/modules/cli/src/main/scala/coursier/cli/install/Install.scala
+++ b/modules/cli/src/main/scala/coursier/cli/install/Install.scala
@@ -7,7 +7,7 @@ import java.time.Instant
 
 import caseapp.core.app.CaseApp
 import caseapp.core.RemainingArgs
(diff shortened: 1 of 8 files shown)


modules/cli/src/main/scala/coursier/cli/install/Install.scala
```diff
@@ -44,26 +44,7 @@ object Install extends CaseApp[InstallOptions] {
       .withVerbosity(params.output.verbosity)
       .withNativeImageJavaHome(Some(graalvmHome))
 
-    if (params.installChannels.nonEmpty) {
```
alexarchambault, line 47: Let's keep that just-in-case, and print a deprecation warning if it's used.

modules/cli/src/main/scala/coursier/cli/install/InstallOptions.scala
```diff
@@ -24,8 +24,6 @@ final case class InstallOptions(
   @Recurse
     envOptions: EnvOptions = EnvOptions(),
 
-  addChannel: List[String] = Nil,
```
alexarchambault, line 27: See the comment above. Let's not remove options for now, but rather print a deprecation warning.

alexarchambault (commented):
Thanks @shubhamJay! I left a comment - about keeping the old options, along with the new command.

modules/cli/src/main/scala/coursier/cli/install/Install.scala
shubhamJay: any place which I can refer, to understand the convention for marking a command deprecated

shubhamJay: Thank you. I have added the same and pushed.

modules/cli/src/main/scala/coursier/cli/install/Install.scala
```diff
@@ -45,24 +45,10 @@ object Install extends CaseApp[InstallOptions] {
       .withNativeImageJavaHome(Some(graalvmHome))
 
     if (params.installChannels.nonEmpty) {
+      val progName = coursier.cli.Coursier.progName
+      val options = params.installChannels.flatMap(c => Seq("--add", c)).mkString(" ")
+      System.err.println(s"Warning: the --add-channel option is deprecated. Use '$progName channel $options' instead.")
 
-      // TODO Move to install module
```
alexarchambault, line 49: Let's keep the code that was removed too, so that a deprecation warning is printed, but the channels are actually added. The goal would be to nudge people to migrate to the new command, but not disrupt anyone's scripts using it.

modules/cli/src/main/scala/coursier/cli/install/InstallOptions.scala
```diff
@@ -24,6 +24,7 @@ final case class InstallOptions(
   @Recurse
     envOptions: EnvOptions = EnvOptions(),
 
+  @Help("(deprecated)")
```
alexarchambault, line 27: 👍

modules/cli/src/main/scala/coursier/cli/install/Install.scala
alexarchambault: Or rather, call the equivalent logic in `coursier.cli.channel.ChannelCommand` from here.

shubhamJay: makes sense. Called the logic to add channel from here

alexarchambault:
Merging, thanks @shubhamJay!

PR #1900, opened 2020-10-20
Improve heap occupancy

jtjeferreira:
Hi! I was looking into some memory dumps from sbt and noticed that a non negligilbe percentage comes from sbt/coursier. I then ran [yourkit duplicate objects inspection](https://www.yourkit.com/docs/java/help/inspections_mem.jsp#duplicate_objects) and got the following

image

So I think there is opportunity to introduce some deduplication/memoization. First thing I did was to to write a test to measure any improvements (see 2 `ObjectSizeTests.scala`). These tests use JOL (https://www.baeldung.com/jvm-measuring-object-sizes) to measure the object graph size. 

This is what I got for the "resolution" of `dep"org.apache.spark:spark-sql_2.12:2.4.0"` for `develop`:

```
coursier.core.Resolution@3dd762add footprint:
     COUNT       AVG       SUM   DESCRIPTION
     16821        48    823904   [C
       342       144     49248   [Ljava.lang.Object;
       581        38     22176   [Lscala.collection.immutable.HashMap;
      3224        36    119200   [Lscala.collection.immutable.HashSet;
         1        16        16   coursier.Resolve$$$Lambda$137/125185118
       415        24      9960   coursier.core.Activation
       416        32     13312   coursier.core.Activation$Os
      9480        16    151680   coursier.core.Configuration
     11458        48    549984   coursier.core.Dependency
         1        32        32   coursier.core.DependencySet
       157        24      3768   coursier.core.DependencySet$Sets
       251        40     10040   coursier.core.Info
       112        32      3584   coursier.core.Info$License
       141        24      3384   coursier.core.Info$Scm
      9939        32    318048   coursier.core.Module
      4227        16     67632   coursier.core.ModuleName
      3512        16     56192   coursier.core.Organization
       415        40     16600   coursier.core.Profile
       251        80     20080   coursier.core.Project
      9512        32    304384   coursier.core.Publication
         2       128       256   coursier.core.Resolution
         1        16        16   coursier.core.Resolution$$Lambda$434/1676871737
       191        16      3056   coursier.core.Type
        26        24       624   coursier.core.Version
        38        24       912   coursier.core.Version$Number
        16        24       384   coursier.core.VersionInterval
         1        32        32   coursier.maven.MavenRepository
         1        80        80   coursier.params.ResolutionParams
         2        16        32   java.lang.Boolean
     16964        24    407136   java.lang.String
         2        64       128   java.util.concurrent.ConcurrentHashMap
         1        16        16   scala.None$
      1453        16     23248   scala.Some
     18598        24    446352   scala.Tuple2
         1        24        24   scala.Tuple3
         1        16        16   scala.collection.Iterator$$Lambda$454/1715693094
         1        24        24   scala.collection.Iterator$$anon$3
      4282        24    102768   scala.collection.immutable.$colon$colon
      2556        32     81792   scala.collection.immutable.HashMap$HashMap1
       581        24     13944   scala.collection.immutable.HashMap$HashTrieMap
         1        16        16   scala.collection.immutable.HashSet$EmptyHashSet$
     13511        24    324264   scala.collection.immutable.HashSet$HashSet1
      3224        24     77376   scala.collection.immutable.HashSet$HashTrieSet
        53        32      1696   scala.collection.immutable.IntMap$Bin
       210        24      5040   scala.collection.immutable.IntMap$Tip
         1        16        16   scala.collection.immutable.Map$EmptyMap$
       200        24      4800   scala.collection.immutable.Map$Map1
        39        32      1248   scala.collection.immutable.Map$Map2
        13        40       520   scala.collection.immutable.Map$Map3
        61        48      2928   scala.collection.immutable.Map$Map4
         1        16        16   scala.collection.immutable.Nil$
         1        16        16   scala.collection.immutable.Set$EmptySet$
       557        16      8912   scala.collection.immutable.Set$Set1
       115        24      2760   scala.collection.immutable.Set$Set2
        52        24      1248   scala.collection.immutable.Set$Set3
        48        32      1536   scala.collection.immutable.Set$Set4
         1        24        24   scala.collection.immutable.Stream$$Lambda$455/1134769044
      7895        24    189480   scala.collection.immutable.Stream$Cons
         1        16        16   scala.collection.immutable.Stream$Empty$
       313        56     17528   scala.collection.immutable.Vector
        16        16       256   scala.util.Left
         8        16       128   scala.util.Right
    142295             4263888   (total)
```

and this what I got after my changes
```
coursier.core.Resolution@409924b3d footprint:
     COUNT       AVG       SUM   DESCRIPTION
     10566        52    558224   [C
       344       144     49536   [Ljava.lang.Object;
       576        38     22168   [Lscala.collection.immutable.HashMap;
      2173        36     79944   [Lscala.collection.immutable.HashSet;
         1        16        16   coursier.Resolve$$$Lambda$138/1861585601
       415        24      9960   coursier.core.Activation
       416        32     13312   coursier.core.Activation$Os
        25        16       400   coursier.core.Classifier
     12203        16    195248   coursier.core.Configuration
      2723        56    152488   coursier.core.Dependency
         1        32        32   coursier.core.DependencySet
       157        24      3768   coursier.core.DependencySet$Sets
        25        16       400   coursier.core.Extension
       251        40     10040   coursier.core.Info
       112        32      3584   coursier.core.Info$License
       141        24      3384   coursier.core.Info$Scm
       864        40     34560   coursier.core.Module
      2045        16     32720   coursier.core.ModuleName
      1505        16     24080   coursier.core.Organization
       415        40     16600   coursier.core.Profile
       251        80     20080   coursier.core.Project
        25        32       800   coursier.core.Publication
         2       128       256   coursier.core.Resolution
         1        16        16   coursier.core.Resolution$$Lambda$434/1409425781
       216        16      3456   coursier.core.Type
        26        24       624   coursier.core.Version
        38        24       912   coursier.core.Version$Number
        16        24       384   coursier.core.VersionInterval
         1        32        32   coursier.maven.MavenRepository
         1        80        80   coursier.params.ResolutionParams
         2        16        32   java.lang.Boolean
     10708        24    256992   java.lang.String
         2        64       128   java.util.concurrent.ConcurrentHashMap
         1        16        16   scala.None$
      1453        16     23248   scala.Some
     14740        24    353760   scala.Tuple2
       865        24     20760   scala.Tuple3
        25        32       800   scala.Tuple4
      2723        40    108920   scala.Tuple7
         1        16        16   scala.collection.Iterator$$Lambda$454/190658073
         1        24        24   scala.collection.Iterator$$anon$3
      4310        24    103440   scala.collection.immutable.$colon$colon
      2564        32     82048   scala.collection.immutable.HashMap$HashMap1
       576        24     13824   scala.collection.immutable.HashMap$HashTrieMap
         1        16        16   scala.collection.immutable.HashSet$EmptyHashSet$
      8959        24    215016   scala.collection.immutable.HashSet$HashSet1
      2173        24     52152   scala.collection.immutable.HashSet$HashTrieSet
        57        32      1824   scala.collection.immutable.IntMap$Bin
       214        24      5136   scala.collection.immutable.IntMap$Tip
         1        16        16   scala.collection.immutable.Map$EmptyMap$
       195        24      4680   scala.collection.immutable.Map$Map1
        43        32      1376   scala.collection.immutable.Map$Map2
        14        40       560   scala.collection.immutable.Map$Map3
        59        48      2832   scala.collection.immutable.Map$Map4
         1        16        16   scala.collection.immutable.Nil$
         1        16        16   scala.collection.immutable.Set$EmptySet$
       301        16      4816   scala.collection.immutable.Set$Set1
        28        24       672   scala.collection.immutable.Set$Set2
        12        24       288   scala.collection.immutable.Set$Set3
         8        32       256   scala.collection.immutable.Set$Set4
         1        24        24   scala.collection.immutable.Stream$$Lambda$455/249701190
      7895        24    189480   scala.collection.immutable.Stream$Cons
         1        16        16   scala.collection.immutable.Stream$Empty$
       315        56     17640   scala.collection.immutable.Vector
        16        16       256   scala.util.Left
         8        16       128   scala.util.Right
     93809             2698328   (total)
```

We went from `4263888` bytes to `2698328` (almost half) just by memoizing `coursier.core.Dependency`, `coursier.core.Publication` and `coursier.core.Module` and I think there is room for improvement in other classes.

To implement the memoization I got inspired `stalagmite` in https://gitlab.com/fommil/attic/-/blob/master/stalagmite/src/test/scala/testing/weakmemoised/weakmemoised.scala#L94, but eventually these changes would have to be ported to https://github.com/alexarchambault/data-class/.

Please note that besides the memoization the equals implementation is now reference based, which might also have performance improvements.

The biggest downside still to be solved is mima compatibility. Besides a few easy ones like `the type hierarchy of class coursier.core.Module is different in current version. Missing types {scala.Product,scala.Serializable,scala.Equals,java.io.Serializable}`, the `method this(java.lang.String,java.lang.String,scala.collection.immutable.Map)Unit in class coursier.core.Module does not have a correspondent in current version` is a problem because I would need to make the constructor private.

WDYT?


jtjeferreira:
Fixed the binary incompatibility of the constructor...

alexarchambault:
Thanks @jtjeferreira! I definitely agree that lowering the allocation rate would be a good thing.

Things could be moved to data-class once things are settled here, yes. An extra field like `memoize: Boolean` can be added to the [`@data`](https://github.com/alexarchambault/data-class/blob/7ab3f1802538943dc4302141047463da0d250ed9/src/main/scala/dataclass/data.scala) annotation to enable that.

alexarchambault:
> Fixed the binary incompatibility of the constructor...

Yes, binary compatibility shouldn't be a problem at the end I think. Some generated methods must just be missing…

alexarchambault:
> Please note that besides the memoization the equals implementation is now reference based, which might also have performance improvements.

That's a problem though. `coursier.core.Dependency` equality must compare individual fields during resolution. [`Resolution.dependencySet`](https://github.com/coursier/coursier/blob/6557ea9d1696707aa91e5c1a1575bd6bcc321298/modules/core/shared/src/main/scala/coursier/core/Resolution.scala#L667) needs it in particular.

jtjeferreira:
> > Please note that besides the memoization the equals implementation is now reference based, which might also have performance improvements.
> 
> That's a problem though. `coursier.core.Dependency` equality must compare individual fields during resolution. [`Resolution.dependencySet`](https://github.com/coursier/coursier/blob/6557ea9d1696707aa91e5c1a1575bd6bcc321298/modules/core/shared/src/main/scala/coursier/core/Resolution.scala#L667) needs it in particular.

But since all the `coursier.core.Dependency` are memoized (and we make sure all the instantiations happen through `Dependency.apply`), then reference based equality works, otherwise the tests would be failing.

(Speaking of tests, could you have a look at CI? I am having weird `java.lang.VerifyError: Bad type on operand stack` locally on ìnstall tests`, but that also happens in master)

modules/core/shared/src/main/scala/coursier/core/Dependency.scala
```diff
@@ -43,15 +47,73 @@ import dataclass.data
...
+    b.append(", ")
+    b.append(String.valueOf(transitive))
+    b.append(")")
+    b.toString
+  }
 }
 
 object Dependency {
 
+  val memoised_cache: java.util.WeakHashMap[(Module, String, Configuration, Set[(Organization, ModuleName)], Publication, Boolean, Boolean), java.lang.ref.WeakReference[Dependency]] =
```
alexarchambault, line 82: Having both `WeakHashMap` and `WeakReference` seems fishy… Isn't only one of them enough?

jtjeferreira:
> (Speaking of tests, could you have a look at CI? I am having weird `java.lang.VerifyError: Bad type on operand stack` locally on `install` tests, but that also happens in master)

Nevermind the tests are now okay. The failure is in mima (which will be fixed when we revert back to data-class

modules/core/shared/src/main/scala/coursier/core/Dependency.scala
```diff
@@ -43,15 +47,73 @@ import dataclass.data
...
+  def apply(module: Module, version: String, configuration: Configuration, exclusions: Set[(Organization, ModuleName)], publication: Publication, optional: Boolean, transitive: Boolean): Dependency = {
+    val key = (module, version, configuration, exclusions, publication, optional, transitive)
+    val first = {
+      val weak = memoised_cache.get(key)
+      if (weak == null) null else weak.get
+    }
+    if (first != null) {
+      first
+    } else {
+      memoised_cache.synchronized {
```
alexarchambault, line 94: I'd be wary of using `synchronized`… I would have just used a `java.util.concurrent.ConcurrentHashMap[(…), WeakReference[…]]` here, and handled concurrent insertions at the same key with `putIfAbsent`.

alexarchambault:
> But since all the coursier.core.Dependency are memoized (and we make sure all the instantiations happen through Dependency.apply), then reference based equality works, otherwise the tests would be failing.

Right. This works if we're extra careful about concurrency, and if the constructors are private.

But about making the constructors private, I'd have to check again, but I think it prevents `shapeless.Generic` to be derived for these classes, which can be handy sometimes. I also remember having had issues with generating actually private constructors from data-class - it seemed the private modifier wasn't always taken into account… Also, the constructors are currently public IIRC, so making them private would break binary compatibility.

alexarchambault:
As a middle ground, maybe we could have a slightly more optimized `equals` implementation, comparing `hashCode`s, then comparing the fields if the hash codes are equal.

Hash codes are currently marked as `lazy`, but they are computed for `memoised_cache`. So maybe they could simply be passed upfront as argument to the constructor.

modules/core/shared/src/main/scala/coursier/core/Dependency.scala
jtjeferreira: I can try, but I used I tried to do the same as stalagmite which has some nice tests...

modules/core/shared/src/main/scala/coursier/core/Dependency.scala
jtjeferreira: Again used what stalagmite is using, but they have a warning about this: https://gitlab.com/fommil/attic/-/blob/master/stalagmite/src/test/scala/testing/weakmemoised/weakmemoised.scala#L69

jtjeferreira:
> As a middle ground, maybe we could have a slightly more optimized `equals` implementation, comparing `hashCode`s, then comparing the fields if the hash codes are equal.

Well, but that wont improve heap occupancy due to duplicate objects...

alexarchambault:
> Well, but that wont improve heap occupancy due to duplicate objects...

Yes, of course! I meant using this particular `equals` only as a substitute to reference equality. Memoization should stay.

From the re-collections I have of past attempts in data-class, I'm not 100% sure we'll succeed in making all constructors private with it. So theoretically,  some users might still create non-memoized instances. So this `equals` would be fine with that, while still being more optimal than the current one. These constructors, except the main one, could (hopefully) be marked as `@deprecated`, so that we would catch where they're used in the coursier code base, and stop doing so. And users of coursier could do the same.

jtjeferreira:
> > Well, but that wont improve heap occupancy due to duplicate objects...
> 
> Yes, of course! I meant using this particular `equals` only as a substitute to reference equality. Memoization should stay.

clear. I will try to introduce these changes to data-class. 

Meanwhile I pushed a small commit that removes some elusive `scala.Stream` from the object graph that was being reported by jol. The heap size stays more or less the same, but I think removing the Stream is still a good move.

jtjeferreira:
Hi @alexarchambault! 

I created 2 PRs on data-class
* alexarchambault/data-class#48 (reference equality)
* alexarchambault/data-class#49 (setters need to call apply)

which are requirements for this PR. I published `data-class` locally and got things working with much less changes, and Mima is happy.

Regarding public constructors, I only now realized the cleverness of your suggestion about better equals implementation. So I agree we can keep constructor public, but if one wants to make use of `memoised_cache`, then it should use `apply`, but otherwise things will still work correctly if one mixes both approaches.

jtjeferreira:
@alexarchambault ready for review :)

modules/core/shared/src/main/scala/coursier/core/Dependency.scala
alexarchambault: Re-reading this, I think I was wrong about it 😅

`WeakHashMap` has weak _keys_, but not weak values. Using a `WeakReference[_]` as value ensures the values are weak too.

alexarchambault:
Thanks @jtjeferreira, it's a nice job overall! There are two things I'm not too happy with the current state of things though:
- the use of `synchronized` (because of `WeakHashMap` not being thread-safe)
- memoization not being handled from data-class (with just some `@data(memoization = true)` here).

If you're ok with it, let me experiment a bit on top of what you did in the coming days. In particular, I'd like to see if the following is do-able:
- using [`ConcurrentReferenceHashMap`](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/ConcurrentReferenceHashMap.java) from Spring (copy-pasting it here),
- adding a `jvmMemoizationWith: String` field in the `data` annotation, that users can pass a `ConcurrentHashMap`-like class name (here, we would pass our `ConcurrentReferenceHashMap`), that would enable memoization on the JVM.

alexarchambault:
Unless you want to give that kind of things a try?

jtjeferreira:
> Unless you want to give that kind of things a try yourself?

please go ahead.

jtjeferreira:
> These constructors, except the main one, could (hopefully) be marked as @deprecated, so that we would catch where they're used in the coursier code base, and stop doing so. And users of coursier could do the same.

Should we do this?

Also, do you think we should apply the cache to other classes?

modules/core/shared/src/main/scala/coursier/core/Resolution.scala
```diff
@@ -1193,10 +1193,10 @@ object Resolution {
           .flatMap(projectCache(_)._2.dependencies)
       )
       .withDependencyManagement(
-        depMgmt.values.toSeq
```
jtjeferreira, line 1196: btw this was to remove a `Stream` as explained in 1cb8f3d15358fcb856f86e28261a0c63fec0c6c9

modules/core/jvm/src/test/scala/coursier/core/ObjectSizeTests.scala
```diff
@@ -57,30 +57,34 @@ object ObjectSizeTests extends TestSuite {
...
     }
 
-    test("Dependency memoised_cache should hold objects until they can be GCd") {
+    test("Dependency instanceCache should hold objects until they can be GCd") {
       test("should be the different for different dependency") {
         implicit val retryMax = RetryMax(5.seconds)
         implicit val retryInterval = RetryInterval(200.millis)
-        Dependency.memoised_cache.clear();
+        def cacheSize(): Int = {
+          Dependency.instanceCache.purgeStaleEntries()
```
jtjeferreira: Why you need to call purgeStaleEntries?

alexarchambault: Because unlike `get` for example, `size` still includes stale entries (whose keys and / or values where garbage-collected).

jtjeferreira:
@alexarchambault anything left to merge this? The license issue you mentioned in commit message is still a problem?

alexarchambault:
Let's merge, thanks for having pushed for this! This is probably going to evolve a bit again in the future, if memoization support is managed by data-class in particular.

jtjeferreira:
> Let's merge, thanks for having pushed for this, @jtjeferreira! This is probably going to evolve a bit again in the future, if memoization support is managed by data-class in particular.

Thx for you support. Any plan for a release soon? Would like to do a PR in sbt

modules/core/shared/src/main/scala/coursier/core/Resolution.scala
jtjeferreira: any reason you reverted this? Can I include it in another PR?

alexarchambault: > Can I include it in another PR?

Sure, go ahead! It was somewhat unrelated to the other changes of this PR (memo-izing some data class instances), I preferred not to include it here.

PR #1903, opened 2020-10-22
Attempt to fix #1884

### LINKED ISSUE #1884: jdk:1.6.65 is returned as available... but it isn't
```
$ cs java --available | rg 1\.6
119:jdk:1.6.65
$ cs java --jvm jdk:1.6.65 -version
JVM jdk:1.6.65 not found in index: JVM jdk not found
```

note:
I wrote a unit test reproducing https://github.com/coursier/coursier/issues/1884 and managed to fix the direct reason for it.

However it still doesn't work. For `jdk1.6.65`:

```
[info] running coursier.cli.Coursier java --jvm jdk:1.6.65 -version
JVM jdk:1.6.65 not found in index: Unrecognized archive type 'dmg'
```

For `1.15.0`:

```
Extraction failed: org.codehaus.plexus.archiver.ArchiverException: Error while expanding /Users/michal/Library/Caches/Coursier/v1/https/download.oracle.com/otn-pub/java/jdk/15.0.1%252B9/51f4f36ad4ef43e39d0dfdbaf6549e32/jdk-15.0.1_osx-x64_bin.tar.gz
```

Which makes sense given that following link from jabba (https://github.com/shyiko/jabba/blob/master/index.json#L1358) you are redirected to https://download.oracle.com/errors/download-fail-1505220.html

Therefore I don't know if there's a proper fix on the coursier side. I wonder why jabba maintains links that do not work though

### DIFF
--- a/modules/jvm/src/main/scala/coursier/jvm/JvmIndex.scala
+++ b/modules/jvm/src/main/scala/coursier/jvm/JvmIndex.scala
@@ -195,7 +195,12 @@ object JvmIndex {
       arch <- arch.map(Right(_)).getOrElse(JvmIndex.currentArchitecture)
       osIndex <- content.get(os).toRight(s"No JVM found for OS $os")
       archIndex <- osIndex.get(arch).toRight(s"No JVM found for OS $os and CPU architecture $arch")
-      versionIndex <- archIndex.get(jdkNamePrefix.getOrElse("") + name).toRight(s"JVM $name not found")
+      versionIndex <- 
+        archIndex.get(jdkNamePrefix.getOrElse("") + name)
+          // While almost all entries in https://github.com/shyiko/jabba/blob/master/index.json follows `jdk@*` pattern
+          // there's also one `jdk` entry. Fixes https://github.com/coursier/coursier/issues/1884
+          .orElse(archIndex.get(name))
+          .toRight(s"JVM $name not found")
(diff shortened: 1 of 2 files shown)


note:
@alexarchambault Of course feel free to just close it. I just wanted to share what I did on my way but since the issue lies deeper it probably does not bring any value.

alexarchambault:
@note About JDK 1.6 and DMG files, see the [comment](https://github.com/coursier/coursier/issues/1884#issuecomment-714399341) I just added in https://github.com/coursier/coursier/issues/1884. The listing logic needs to be made aware of the packaging types that are supported.

But support for DMG files could also be added. jabba handles them [this way](https://github.com/shyiko/jabba/blob/31090e95cd248ebae0f412a1f9add8cc92f1a306/command/install.go#L332-L375). Last time I checked, I didn't stumble upon libraries allowing to deal with DMG files from Java / Scala, so we may have to resort to using the same commands as jabba.

alexarchambault:
About JDK 15, I had missed those in the jabba index. Currently, I'm using the [JVM index from coursier](https://github.com/coursier/jvm-index), which is not too widely known yet, to use openjdk 15:
```text
$ cs java --jvm-index cs --jvm 15 -version
openjdk version "15" 2020-09-15
OpenJDK Runtime Environment AdoptOpenJDK (build 15+36)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 15+36, mixed mode, sharing)
```

PR #1907, opened 2020-10-29
changing=false urls are not changing

martijnhoekstra:


martijnhoekstra:
Maybe it needs tests, but I couldn't figure out how or where

alexarchambault:
Thanks @martijnhoekstra!

There are some tests [here](https://github.com/coursier/coursier/blob/adf62e28488d4ff6efa2db8ed654b052f22f3ace/modules/cli-tests/src/main/scala/coursier/clitests/GetTests.scala). I'm not sure how I would test this without resorting to too heavyweight stuff (like spawning an http4s server during the test, so that we can change the changing artifact - it's done elsewhere in the coursier codebase, grep for http4s)… Feel free to add a test there if you have other ideas or if you feel like it. The code of `cs get` is minimal, this can be merged as is else…

alexarchambault:
Merging as is, so that it's included in the upcoming release.

PR #1920, opened 2020-11-21
Dependency micro optimizations

jtjeferreira:
since Dependency is now cached, "allocations" are more expensive. So I introduced a copy method to avoid chaining `with`s

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
@@ -64,7 +64,7 @@ object ModuleName {
 
   override final lazy val hashCode = tuple.hashCode()
 
-  def copy(
+  private[core] def copy(
     organization: Organization = this.organization,
     name: ModuleName = this.name,
     attributes: Map[String, String] = this.attributes
--- a/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
@@ -40,9 +40,10 @@ object ModuleName {
   attributes: Map[String, String]
 ) {
 
-  def trim: Module =
-    withOrganization(organization.map(_.trim))
-      .withName(name.map(_.trim))
+  def trim: Module = copy(
+    organization.map(_.trim),
+    name.map(_.trim)
+  )
 
   private def attributesStr = attributes.toSeq
     .sortBy { case (k, _) => k }
@@ -62,6 +63,12 @@ object ModuleName {
     s"${organization.value}:${name.value}"
 
   override final lazy val hashCode = tuple.hashCode()
+
+  def copy(
+    organization: Organization = this.organization,
+    name: ModuleName = this.name,
+    attributes: Map[String, String] = this.attributes
+  ) = Module(organization, name, attributes)
 }
 
 object Module {
--- a/modules/core/shared/src/main/scala/coursier/core/Dependency.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Dependency.scala
@@ -45,6 +45,14 @@ import dataclass.data
   def withPublication(name: String, `type`: Type, ext: Extension, classifier: Classifier): Dependency =
     withPublication(Publication(name, `type`, ext, classifier))
 
+  def copy(
+    module: Module,
+    version: String,
+    configuration: Configuration,
+    exclusions: Set[(Organization, ModuleName)],
+    attributes: Attributes
+  ) = Dependency(module, version, configuration, exclusions, attributes, this.optional, this.transitive)
+
   lazy val clearExclusions: Dependency =
     withExclusions(Set.empty)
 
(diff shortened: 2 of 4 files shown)


jtjeferreira:
I ran `coursier.benchmark.ResolutionTests` with 
`benchmark/jmh:run coursier.benchmark.ResolutionTests -i 1 -wi 0 -f1 -t1 -prof "async:libPath=/home/joao/async-profiler-1.8.2-linux-x64/build/libasyncProfiler.so;output=flamegraph"`

```
[info] Benchmark                           Mode  Cnt    Score   Error  Units
[info] ResolutionTests.coursierCli         avgt         3.993          ms/op
[info] ResolutionTests.coursierCli:·async  avgt           NaN            ---
[info] ResolutionTests.sparkSql            avgt       176.868          ms/op
[info] ResolutionTests.sparkSql:·async     avgt           NaN            ---
```
vs 2.0.7
```
[info] Benchmark                           Mode  Cnt    Score   Error  Units
[info] ResolutionTests.coursierCli         avgt         4.002          ms/op
[info] ResolutionTests.coursierCli:·async  avgt           NaN            ---
[info] ResolutionTests.sparkSql            avgt       198.346          ms/op
[info] ResolutionTests.sparkSql:·async     avgt           NaN            ---
```

Also in the flamegraph the references to "refhashmap" were reduced from 12.7% to 8.4%

jtjeferreira:
More recent number for `benchmark/jmh:run coursier.benchmark.ResolutionTests -i 5 -wi 1 -f1 -t1 -prof "async:libPath=/home/joao/async-profiler-1.8.2-linux-x64/build/libasyncProfiler.so;output=flamegraph"`


```
[info] Benchmark                           Mode  Cnt    Score   Error  Units
[info] ResolutionTests.coursierCli         avgt    5    3.149 ± 0.068  ms/op
[info] ResolutionTests.coursierCli:·async  avgt           NaN            ---
[info] ResolutionTests.sparkSql            avgt    5  158.446 ± 9.968  ms/op
[info] ResolutionTests.sparkSql:·async     avgt           NaN            ---
```

vs 2.0.7

```
[info] Benchmark                           Mode  Cnt    Score   Error  Units
[info] ResolutionTests.coursierCli         avgt    5    3.135 ± 0.071  ms/op
[info] ResolutionTests.coursierCli:·async  avgt           NaN            ---
[info] ResolutionTests.sparkSql            avgt    5  176.969 ± 6.622  ms/op
[info] ResolutionTests.sparkSql:·async     avgt           NaN            ---
```

modules/core/shared/src/main/scala/coursier/core/Definitions.scala
```diff
@@ -62,6 +63,12 @@ object ModuleName {
     s"${organization.value}:${name.value}"
 
   override final lazy val hashCode = tuple.hashCode()
+
+  def copy(
```
alexarchambault, line 67: I'm wary of the binary compatibility story of `copy` methods… Couldn't we add more fine grained methods, like
```scala
def update(
  org: Organization,
  name: Name
): Module = …
```
with no default values in particular.

It may also be interesting to generate a full `copy` method from data-class, but marked as `private`. That way, we would call that private `copy` from additional methods like `update`.

modules/core/shared/src/main/scala/coursier/core/Dependency.scala
```diff
@@ -45,6 +45,16 @@ import dataclass.data
   def withPublication(name: String, `type`: Type, ext: Extension, classifier: Classifier): Dependency =
     withPublication(Publication(name, `type`, ext, classifier))
 
+  def copy(
```
alexarchambault, line 48: Like above, let's just add a more fine grained `update` method, with no default values.

modules/core/shared/src/main/scala/coursier/core/Definitions.scala
jtjeferreira: as long this is not an autogenerated copy method then it is a binary compatible one, right? 

I can try to make a PR to data-class, but that might take while...

alexarchambault: I was thinking, we could just mark this `private[core]`, and be done with it (no need to consider binary compatibility then):
```suggestion
  private[core] def copy(
```

modules/core/shared/src/main/scala/coursier/core/Dependency.scala
alexarchambault: Just like above, we can mark this `private[core]`, so that we don't have to maintain binary compatibility for it:
```suggestion
  private[core] def copy(
```

modules/core/shared/src/main/scala/coursier/core/Definitions.scala
jtjeferreira: Sorry I forgot about this PR...

jtjeferreira: @alexarchambault ready for review/merge...

alexarchambault (approved):
Thanks!

PR #1937, opened 2020-12-22
Update coursier native launcher installation documentation

eloots:
- MacOS and Linux instructions to install the native launcher
  are now unified

### DIFF
--- a/doc/docs/cli-installation.md
+++ b/doc/docs/cli-installation.md
@@ -4,26 +4,16 @@ title: Installation
 
 ## Native launcher
 
-### Linux
+### Linux & macOS
 
-On Linux, download and run the coursier launcher with
+On Linux and macOS, download and run the coursier launcher with
 ```bash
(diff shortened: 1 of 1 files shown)


eloots:
@alexarchambault 
Maybe the `cs` native launcher installation instructions could be further modified:

- Install `cs` following the instructions that are already in the doc.
- Use `./cs install cs` to install `cs`.
- Remove the local `cs` executable.

WDYT?

alexarchambault:
It'd be nice to have those instructions too, yes. Maybe right after something like "if you wish to put `cs` in your `PATH`, you can do the following: …". And we can also mention that this allows to later update the launcher with `cs update cs`.

eloots:
> It'd be nice to have those instructions too, yes.

@alexarchambault Just updated the instructions along the lines of what you suggested.

alexarchambault (approved):
Thanks @eloots!

Issue #1951, opened 2021-01-25
Symlinking `~/.local/share/coursier/bin/cs` to `~/bin/cs` fails to run `cs`


bs76:
When I link `~/.local/share/coursier/bin/cs` to `~/bin/cs` and then run `~/bin/cs` it fails. with:
```
./cs: 4: exec: /home/<whatever>/tmp/cs/.cs.aux: not found
```

The `cs` script uses `dirname $0` which resolves to dir of the symlink, not the location the symlink point to.
cs should resolve the symlink (a chain of them) to the destination cs script using readlink/realpath/stat

I want to symlink `cs` to location which is under yadm , I do not want to add `~/.local/share/coursier/bin` as a tracked location.

PR #1952, opened 2021-01-25
fixing symlink resolution to `cs`

bs76:
Added readlink/realpath/stat piped to resolve actual path of `cs` when symlinked. I can not test on MacOS so I added realpath and stat based on: [stackoverflow](https://stackoverflow.com/questions/1055671/how-can-i-get-the-behavior-of-gnus-readlink-f-on-a-mac#4031502)

Could you please advise how to test ?

Please see: #1951

### DIFF
--- a/modules/install/src/main/scala/coursier/install/InstallDir.scala
+++ b/modules/install/src/main/scala/coursier/install/InstallDir.scala
@@ -370,7 +370,7 @@ import scala.util.control.NonFatal
               else
                 baseNativePreamble
                   .withKind(Preamble.Kind.Sh)
-                  .withCommand(""""$(cd "$(dirname $(readlink "$0"))"; pwd)/""" + auxName(dest0.getFileName.toString, "") + "\"") // FIXME needs directory
+                  .withCommand(""""$(dirname "$(readlink -f "$0" || realpath "$0" || stat -f%Y "$0" )")"/""" + auxName(dest0.getFileName.toString, "") + "\"") // FIXME needs directory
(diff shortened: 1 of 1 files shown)


modules/install/src/main/scala/coursier/install/InstallDir.scala
```diff
@@ -370,7 +370,7 @@ import scala.util.control.NonFatal
               else
                 baseNativePreamble
                   .withKind(Preamble.Kind.Sh)
-                  .withCommand(""""$(cd "$(dirname "$0")"; pwd)/""" + auxName(dest0.getFileName.toString, "") + "\"") // FIXME needs directory
+                  .withCommand(""""$(dirname "$(readlink -f "$0" || realpath "$0" || stat -f%Y "$0" )")/""" + auxName(dest0.getFileName.toString, "") + "\"") // FIXME needs directory
```
alexarchambault, line 373: On macOS, it seems the failing `readlink` prints some junk:
```text
$ sbt cli/pack
$ modules/cli/target/pack/bin/coursier install --install-dir test-install cs
$ ./test-install/cs --help
readlink: illegal option -- f
usage: readlink [-n] [file ...]
Coursier 2.0.9
Usage: cs [options] [command] [command-options]
…
```
Some `2>/dev/null` should be needed, I guess.

alexarchambault (commented):
Thanks @bs76!

Sorry for the late review. This might just need a bit more tweaking for macOS.

Issue #1541, opened 2020-02-04
lmcoursier.CoursierDependencyResolution#update should use passed configuration


OlegYch:
In particular ignoring 'missingOk' causes 'updateClassifiers' sbt task (used by intellij to import projects) to fail in some cases.
The issue was first reported at https://github.com/sbt/sbt/issues/5374

OlegYch:
so, on CI 'sbt-lm-coursier/scripted shared-2/update-sbt-classifiers' fails as expected (https://travis-ci.org/coursier/sbt-coursier/jobs/646427306#L1731) with this change https://github.com/coursier/sbt-coursier/commit/404af50b6d8ee4df7ce7e07406656ebc8cc19575

unfortunately locally it fails with 'assertion failed: Any of WrappedArray((org.scala-lang,scala-library)) not found'
and 'sbt-coursier/scripted shared-2/update-sbt-classifiers' does not fail at all, though it does print 'Found module 		org.webjars:npm:4.2.0:default: (MISSING) Artifact(npm, src, jar, Some(sources), Vector(), Some(https://repo1.maven.org/maven2/org/webjars/npm/4.2.0/npm-4.2.0-sources.jar), Map(), None)'

so i'm not sure how to proceed, it looks like in some cases missing artifact is treated as error and in others it doesn't

mikail-khan:
The only way around this I have found so far is to drop back down to 1.3.6, let the project resolve in terms of deps (or whatever it needs) and then move back up to 1.3.8 :frowning_face:

dcsobral:
This is related to sbt/sbt#5465 and sbt/sbt#5078.

andreaTP:
@dcsobral and @OlegYch thanks a lot for narrowing this error down!
I don't have a fix but just found a workaround:
```
addSbtPlugin("com.typesafe.play" % "sbt-plugin" % "2.8.0" exclude ("org.webjars", "npm"))
libraryDependencies += "org.webjars" % "npm" % "4.2.0"
```

Basically, excluding from the transitive dependencies of the sbt plugins the offending ones and adding them back as normal `libraryDependencies` make `updateSbtClassifiers` working.

hepin1989:
Yes，Yes even override will not works

dcsobral:
[...]

PR #1966, opened 2021-02-13
Ignore HTTP 403 Forbidden on optional artifacts

### LINKED ISSUE #1961: "Server returned HTTP response code: 403" on importing sbt project to IntelliJ behind proxy
Ref https://github.com/coursier/coursier/issues/1541

This was originally reported by @texasbruce as https://github.com/sbt/sbt/issues/6300, as a resubmission of the discussion in https://github.com/sbt/sbt/issues/5501

## steps

1. Make sure IntelliJ SBT plugin has checked option "Download Library sources"
2. Import a project with any dependency that does not have source jar in the artifact repo and only binary jar
3. Import fails with 

```
[error] (updateClassifiers) lmcoursier.internal.shaded.coursier.error.FetchError$DownloadingArtifacts: Error fetching artifacts:
[error] https://xxxxx/xxxx/xxxxx/xxxxxxx-sources.jar: download error: Caught java.io.IOException: Server returned HTTP response code: 403 for URL: https://xxxxx/xxxx/xxxxx/xxxxxxx-sources.jar (Server returned HTTP response code: 403 for URL: https://xxxxx/xxxx/xxxxx/xxxxxxx-sources.jar) while downloading https://xxxxx/xxxx/xxxxx/xxxxxxx-sources.jar
[error] Total time: 33 s, completed 8-Feb-2021 11:26:33 AM
```

## problem

One dependency without source shouldn't make the whole project import fails. 

## expectation

Source of dependency without source jar should be skipped and project should be imported successfully in IntelliJ

## notes

I (@eed3si9n) implemented  `missingOk` in https://github.com/coursier/sbt-coursier/pull/212, and later attempted to fix fetch error (source file that doesn't exist) in https://github.com/coursier/sbt-coursier/pull/224, but I closed it in favor of https://github.com/coursier/sbt-coursier/pull/232.

In https://github.com/coursier/coursier/issues/1541#issuecomment-628755175 Alex wrote:

> FYI everyone, coursier considers artifacts from Maven repositories as optional by default. That means if a javadoc or source JAR is missing (with a 404 status), `updateClassifiers` will not fail. There's no need for `missingOk` in most cases thanks to that.

I wonder if it's to do with 404 vs 403.

eed3si9n:
Fixes #1961 / https://github.com/sbt/sbt/issues/6300
Ref #1541

I think some corporate proxies are returning HTTP 403 Forbidden when
accessing a missing source JAR files.
Given Coursier does not ignore these errors, it would have the effect of
`updateClassifier` failing whenever the source JARs are missing from an
external repository, resulting in IntelliJ import failure.

I don't have a test case on this.

### DIFF
--- a/modules/cache/jvm/src/main/scala/coursier/cache/ArtifactError.scala
+++ b/modules/cache/jvm/src/main/scala/coursier/cache/ArtifactError.scala
@@ -17,6 +17,11 @@ sealed abstract class ArtifactError(
     case _: ArtifactError.NotFound => true
     case _ => false
   }
+
+  final def forbidden: Boolean = this match {
+    case _: ArtifactError.Forbidden => true
+    case _ => false
+  }
 }
 
 object ArtifactError {
@@ -32,6 +37,13 @@ object ArtifactError {
     val permanent: Option[Boolean] = None
   ) extends ArtifactError(
     "not found",
+    file
+  )
+
+  final class Forbidden(
+    val file: String
+  ) extends ArtifactError(
+    "forbidden",
     file
   )
 
--- a/modules/cache/jvm/src/main/scala/coursier/cache/internal/Downloader.scala
+++ b/modules/cache/jvm/src/main/scala/coursier/cache/internal/Downloader.scala
@@ -215,6 +215,8 @@ import scala.util.control.NonFatal
 
         if (respCodeOpt.contains(404))
           Left(new ArtifactError.NotFound(url, permanent = Some(true)))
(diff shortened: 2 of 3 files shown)


alexarchambault (approved):
Thanks @eed3si9n!

texasbruce:
@eed3si9n our artifactory server is giving different error now with a SSLHandshakeException for the optional artifacts. Why do we not skip all errorous optional artifacts? From the code it seems like we only skip if not found or forbidden

Issue #1774, opened 2020-06-12
cs bootstrap / cs install launcher script not working with private `-r` repositories


chadselph:
I ran into this issue when test [my wrapper](https://github.com/chadselph/coursier-private-repo-boostrap) which loads .m2 credentials for private maven repositories.

To reproduce:

```console
$ cs launch --credentials ... -r https://private/pro com.company:app:0.1.0 -M com.company.Main
```

this works, downloads jars, command runs but when running bootstrap:

```console
$ cs bootstrap --credentials ... -r https://private/pro com.company:app:0.1.0 -M com.company.Main -o app
```
everything downloads and resolves correctly but when running
``` console
$ ./app
Error while downloading https://private/repo/...app-0.1.0-SNAPSHOT.jar: 
Server returned HTTP response code: 401 for URL: https://... ignoring it
```

it suddenly forgets it has already downloaded the dependencies (and doesn't try to use credentials even if they're written to the credentials.properties file).

Reproduced in a docker container (openjdk11-jre-slim) with `cs --version` `2.0.0-RC6-21`.

alexarchambault:
I confirm credentials are not persisted in bootstraps. Bootstraps are only for repositories with no authentication for now.

Ideally, I'd like [global credentials](https://get-coursier.io/docs/other-credentials.html#property-file) to be used from bootstraps too, but they aren't yet…

chadselph:
yeah bootstraps should definitely not contain the credentials, but it's already downloaded the jars. It's just looking for them in the wrong place: `.cache/coursier/v1/https/user%40server/` vs `.cache/coursier/v1/https/user%40server/`. So, this bootstrap will not be sharable, but it should still run, right? If not, why did the bootstrap command even download them?

The use case I have in actually for the install command. I have an internal URL app channel which points to the internal maven repo.

chadselph:
I'll also mention I found two workarounds.
For cs bootstrap using `--manifest-jar` works
For using app channels, using a standalone or hybrid launcher works.

That being said including the username in the file cache path still seems wrong to me, could there at least be an option to turn it off?

greg-a-atkinson:
> I confirm credentials are not persisted in bootstraps. Bootstraps are only for repositories with no authentication for now.
> 
> Ideally, I'd like [global credentials](https://get-coursier.io/docs/other-credentials.html#property-file) to be used from bootstraps too, but they aren't yet…

Hi @alexarchambault 

[...]

PR #1991, opened 2021-03-07
Enhance bootstrap launcher to download artifacts using global credentials.

greg-a-atkinson:
Possible implementation for issue: https://github.com/coursier/coursier/issues/1774

Within my organisation we use a private authenticated Artifactory instance to serve all internal and external repositories. Ideally we would like to retrieve artifacts through this repository and not via a proxy. (A proxy requires domain credentials, while Artifactory only requires a user specific Artifactry API key, which more secure for adding to configuration files/properties/variables than domain credentials.)

When I configure the following `mirror.properties`:

```
central.from=https://repo1.maven.org/maven2
central.to=https://private.artifactory.com/artifactory/mavencentral
jcenter.from=https://jcenter.bintray.com
jcenter.to=https://private.artifactory.com/artifactory/jcenter
typesafe.from=https://repo.typesafe.com/typesafe/ivy-releases
typesafe.to=https://private.artifactory.com/artifactory/typesafe-ivy-releases
```

Most artifact retrievals succeed, but artifacts retrieved by the bootstrap launcher fail. For example, the Scala Metals plugin for VSCode fails to download various artifacts due to `401 Unauthorized` errors when the bootstrap launcher does not use the configured [global credentials](https://get-coursier.io/docs/other-credentials.html#property-file):

```
Java home: C:\Users\user\working\tools\jdk1.8.0_261-x64
Metals version: 0.9.10
Error while downloading https://private.artifactory.com/artifactory/virtual-java/com/chuusai/shapeless_2.12/2.3.3/shapeless_2.12-2.3.3.jar: Server returned HTTP response code: 401 for URL: https://private.artifactory.com/artifactory/virtual-java/com/chuusai/shapeless_2.12/2.3.3/shapeless_2.12-2.3.3.jar, ignoring it
Error while downloading https://private.artifactory.com/artifactory/virtual-java/com/github/alexarchambault/case-app-annotations_2.12/2.0.0-M9/case-app-annotations_2.12-2.0.0-M9.jar: Server returned HTTP response code: 401 for URL: https://private.artifactory.com/artifactory/virtual-java/com/github/alexarchambault/case-app-annotations_2.12/2.0.0-M9/case-app-annotations_2.12-2.0.0-M9.jar, ignoring it
Error while downloading https://private.artifactory.com/artifactory/virtual-scala/com.lightbend/emoji_2.12/1.2.1/jars/emoji_2.12.jar: Server returned HTTP response code: 401 for URL: https://private.artifactory.com/artifactory/virtual-scala/com.lightbend/emoji_2.12/1.2.1/jars/emoji_2.12.jar, ignoring it
Error while downloading https://private.artifactory.com/artifactory/virtual-java/com/github/alexarchambault/argonaut-shapeless_6.2_2.12/1.2.0-M11/argonaut-shapeless_6.2_2.12-1.2.0-M11.jar: Server returned HTTP response code: 401 for URL: https://private.artifactory.com/artifactory/virtual-java/com/github/alexarchambault/argonaut-shapeless_6.2_2.12/1.2.0-M11/argonaut-shapeless_6.2_2.12-1.2.0-M11.jar, ignoring it
```

The bootstrap launcher is a Java-only module, which makes sense as it will only have use of the core Java libraries before any dependencies are downloaded, but as a consequence it does not have access to the current Scala-based `Credentials` functionality. If ideally [global credentials](https://get-coursier.io/docs/other-credentials.html#property-file) is used from bootstraps, the this could be implemented in a number of ways, each with different pros and cons:

1. The `bootstrap-launcher` module is enhanced with just enough Java-based functionality to utilize the [global credentials](https://get-coursier.io/docs/other-credentials.html#property-file) without any changes to the existing, out-of-scope Scala-based credentials functionality. Pros: No change to the the Scala `Credentials` API or implementation. Cons: Some duplication.
2. A new or exisitng Java-only module, like `paths`, could be enhanced with the complete credentials functionality, the Scala API updated to delegate to the Java implementation and the `boostrap-launcher` updated to utilize the common credentials functionality. Pros: A single credentials implementation. Cons: Change, compatibility and larger initiative.
3. Something else?

Is there a preferred implementation in mind?

What do you think?

Happy to discuss.

### DIFF
--- a/modules/bootstrap-launcher/src/it/scala/coursier/bootstrap/launcher/DownloadTests.scala
+++ b/modules/bootstrap-launcher/src/it/scala/coursier/bootstrap/launcher/DownloadTests.scala
@@ -1,0 +1,96 @@
+package coursier.bootstrap.launcher
+
+import utest._
+
+import coursier.bootstrap.launcher.credentials.Credentials
+import coursier.bootstrap.launcher.credentials.DirectCredentials
+import coursier.paths.CachePath
+
+import scala.jdk.CollectionConverters._
+import java.nio.file.Files
+import java.io.File
+import java.nio.file.Path
+import java.net.URL
+import java.nio.file.Paths
+import java.net.URI
+import java.util.Collections
+
+object DownloadTests extends TestSuite {
+  
+  private val testRepository = Option(System.getenv("TEST_REPOSITORY")).getOrElse(sys.error("TEST_REPOSITORY not set"))
+  private val testRepositoryUser = Option(System.getenv("TEST_REPOSITORY_USER")).getOrElse(sys.error("TEST_REPOSITORY_USER not set"))
+  private val testRepositoryPassword = Option(System.getenv("TEST_REPOSITORY_PASSWORD")).getOrElse(sys.error("TEST_REPOSITORY_PASSWORD not set"))
+  private val testRepositoryHost = new URI(testRepository).getHost
+
+  private def deleteRecursive(f: File): Unit = {
+    if (f.isDirectory)
+      f.listFiles().foreach(deleteRecursive)
+    if (f.exists())
+      f.delete()
+  }
+
+  private def withTmpDir[T](f: Path => T): T = {
+    val dir = Files.createTempDirectory("coursier-test")
+    val shutdownHook: Thread =
+      new Thread {
+        override def run() =
+          deleteRecursive(dir.toFile)
+      }
+    Runtime.getRuntime.addShutdownHook(shutdownHook)
+    try f(dir)
+    finally {
+      deleteRecursive(dir.toFile)
+      Runtime.getRuntime.removeShutdownHook(shutdownHook)
+    }
+  }
+
+  val tests = Tests {
+
+    test("public URL") {
+      withTmpDir { dir =>
+        val remoteUrls = Seq(
+          new URL("https://repo1.maven.org/maven2/com/chuusai/shapeless_2.12/2.3.3/shapeless_2.12-2.3.3.jar")
+        )
+        val download = new Download(1, dir.toFile, Collections.emptyList())
+        val localUrls = download.getLocalURLs(remoteUrls.asJava).asScala
+        assert(remoteUrls.size == localUrls.size)
+        assert(localUrls
+          .map(url => Paths.get(url.toURI).toFile)
+          .foldLeft(true)(_ && _.exists))
+      }
+    }
+
+    test ("private URL with credentials in the URL") {
+      withTmpDir { dir =>
+        val remoteUrls = Seq(
+          new URL(s"$testRepository/com/abc/test/0.1/test-0.1.pom"
+            .replaceFirst(testRepositoryHost, s"$testRepositoryUser:$testRepositoryPassword@$testRepositoryHost"))
+        )
+        val download = new Download(1, dir.toFile, Collections.emptyList())
+        val localUrls = download.getLocalURLs(remoteUrls.asJava).asScala
+        assert(remoteUrls.size == localUrls.size)
+        assert(localUrls
+          .map(url => Paths.get(url.toURI).toFile)
+          .foldLeft(true)(_ && _.exists))
+      }
+    }
+
+    test ("priavte URL with configured credentials") {
+      withTmpDir { dir =>
+        val remoteUrls = Seq(
+          new URL(s"$testRepository/com/abc/test/0.1/test-0.1.pom")
+        )
+        val download = new Download(1, dir.toFile, Collections.singletonList(
+          new DirectCredentials(testRepositoryHost, testRepositoryUser, testRepositoryPassword).withMatchHost(true).withHttpsOnly(false)
+        ))
+        val localUrls = download.getLocalURLs(remoteUrls.asJava).asScala
+        assert(remoteUrls.size == localUrls.size)
+        assert(localUrls
+          .map(url => Paths.get(url.toURI).toFile)
+          .foldLeft(true)(_ && _.exists))
+      }
+    }
+
+  }
+
+}
--- a/build.sbt
+++ b/build.sbt
@@ -218,9 +218,16 @@ lazy val catsJs = cats.js
 lazy val `bootstrap-launcher` = project("bootstrap-launcher")
   .enablePlugins(SbtProguard)
   .disablePlugins(MimaPlugin)
+  .configs(Integration)
   .settings(
     pureJava,
     dontPublish,
+    hasITs,
+    utest,
+    libraryDependencies ++= Seq(
+      Deps.collectionCompat % Test,
+      Deps.java8Compat % Test
+    ),
     addPathsSources,
     addWindowsAnsiPsSources,
     mainClass.in(Compile) := Some("coursier.bootstrap.launcher.Launcher"),
--- a/modules/bootstrap-launcher/src/main/java/coursier/bootstrap/launcher/Download.java
+++ b/modules/bootstrap-launcher/src/main/java/coursier/bootstrap/launcher/Download.java
@@ -4,33 +4,45 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.HttpURLConnection;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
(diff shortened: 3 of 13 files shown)


alexarchambault:
Thanks a lot for taking a cut at this!

About how to add this, I think your point 2. should be the way to go, in the long term. The [mirror logic](https://github.com/coursier/coursier/blob/42261d4739c5cf7eb6f6fef150140e71e98f21c5/modules/paths/src/main/java/coursier/paths/Mirror.java) is already implemented this way, via `paths` (whose name should probably be changed now, as it does more than just dealing with paths). But it doesn't seem straightforward without breaking binary compatibility, yes. And the companion objects of `coursier.credentials.{Credentials, DirectCredentials, FileCredentials}` make it harder to write binary compatibility equivalents in Java.

I guess we can leave it as is in a first time (so your point 1., and the current state of the PR), and work on making this PR mergeable. And at a later point in time, either find a way to use the Java-based `Credentials` everywhere, or wait for the next version bump and break binary compatibility.

In order to run the integration tests you added, we need to start an authenticated mirror repository, right? Do you think you can add that? (I've got to run right now, but there are examples of Sonatype instances being spawn in the code, I can point them later today, if needed).

greg-a-atkinson:
Thanks for your reply. I agree.

Implementation (1) is a subset and requirement for implementation (2) so I thought it a reasonable place to start. In fact I had ported more of the code than is included in the pull request, but chopped it down to what was absolutely necessary.

I'll have a look at spinning up a Sonatype instance in the integration test. I had seen logging for this in the test suite, so I'm sure I can find an example. I developed the change against an Artifactory instance I spun up in a container so, I expect integrating with an in-memory repository should be straight forward. Thanks for the suggestion.

alexarchambault:
Spawning an Artifactory would be fine too. A Sonatype server is spawned via docker around [here](https://github.com/coursier/coursier/blob/ce251d0288bf029e4eeb8700e8990178559a005d/modules/proxy-tests/src/it/scala/coursier/test/CentralNexus2ProxyTests.scala#L7-L11) for example.

By the way, if it's more practical, I think it should be do-able to write the tests in Scala (while keeping the main module in Java). And [`DockerServer.scala`](https://github.com/coursier/coursier/blob/7bba8ce7881a90fdc386c6533f08a114ae964944/modules/proxy-tests/src/it/scala/coursier/test/DockerServer.scala) could be moved there if needed.

greg-a-atkinson:
I found `AuthenticatedProxyTests`, `CentralNexus2ProxyTests` and `CentralNexus3ProxyTests` depending on `DockerServer.scala`, but also found `AuthenticationTests` and `HttpAuthenticationTests` simply depending on the environment variables `TEST_REPOSITORY`, `TEST_REPOSITORY_USER` and `TEST_REPOSITORY_PASSWORD`.

I updated `DownloadTest` to utilise the test repository environment variables and these seemed to do the trick.

What do you think of the updated pull request?

greg-a-atkinson:
Yes and it should be possible to convert the tests. Let me update the PR...

greg-a-atkinson:
Hi @alexarchambault.

The `DownloadTest` integration test has been updated to test authentication by accessing the `TEST_REPOSITORY` and the tests are in Scala while the `main` source tree is Java-only.

Please let me know if there's anything else.

Thanks

modules/bootstrap-launcher/src/it/scala/coursier/bootstrap/launcher/DownloadTests.scala
```diff
@@ -0,0 +1,105 @@
...
+      Runtime.getRuntime.removeShutdownHook(shutdownHook)
+    }
+  }
+
+  val tests = Tests {
+
+    test("public URL") {
+      withTmpDir { dir =>
+        try {
+          Download.setCacheDir(dir.toFile)
```
alexarchambault, line 50: I think the tests might run in parallel, which would be a problem here… To work around that, we can remove the static initializer on `Download`, and make its methods non-static. https://github.com/coursier/coursier/pull/2000 does that, don't hesitate to rebase your work on it.

It should allow to just do
```scala
val download = new Download(dir.toFile, …)
download.getLocalURLs(…)
```
so that each test has its own `Download` instance.

alexarchambault (commented):
Thanks @greg-a-atkinson!

Looks good overall, I only left some minor comments.

modules/bootstrap-launcher/src/it/scala/coursier/bootstrap/launcher/DownloadTests.scala
greg-a-atkinson: This is fine with me. I see that the `bootstrap-launcher` doesn't enforce source and binary compatibility. Are you going to merge #2000? Then I'll rebase.

Thanks

alexarchambault: There's no binary compatibility checks for `bootstrap-launcher`, as it's not meant to be used as a library (the same applies to the `cli` module). I just merged #2000, just feel free to rebase.

greg-a-atkinson: That's done: branch rebased and review comments applied. Please let me know if I missed anything. Thanks

modules/bootstrap-launcher/src/it/scala/coursier/bootstrap/launcher/DownloadTests.scala
```diff
@@ -0,0 +1,96 @@
...
+  }
+
+  val tests = Tests {
+
+    test("public URL") {
+      withTmpDir { dir =>
+        val remoteUrls = Seq(
+          new URL("https://repo1.maven.org/maven2/com/chuusai/shapeless_2.12/2.3.3/shapeless_2.12-2.3.3.jar")
+        )
+        val download = new Download(1, CachePath.defaultCacheDirectory(), Collections.emptyList())
```
alexarchambault, line 54: Shouldn't `dir` be used here?
```suggestion
        val download = new Download(1, dir, Collections.emptyList())
```

modules/bootstrap-launcher/src/it/scala/coursier/bootstrap/launcher/DownloadTests.scala
```diff
@@ -0,0 +1,96 @@
...
+      }
+    }
+
+    test ("private URL with credentials in the URL") {
+      withTmpDir { dir =>
+        val remoteUrls = Seq(
+          new URL(s"$testRepository/com/abc/test/0.1/test-0.1.pom"
+            .replaceFirst(testRepositoryHost, s"$testRepositoryUser:$testRepositoryPassword@$testRepositoryHost"))
+        )
+        val download = new Download(1, CachePath.defaultCacheDirectory(), Collections.emptyList())
```
alexarchambault, line 69: Same as above:
```suggestion
        val download = new Download(1, dir, Collections.emptyList())
```

modules/bootstrap-launcher/src/it/scala/coursier/bootstrap/launcher/DownloadTests.scala
```diff
@@ -0,0 +1,96 @@
...
+          .foldLeft(true)(_ && _.exists))
+      }
+    }
+
+    test ("priavte URL with configured credentials") {
+      withTmpDir { dir =>
+        val remoteUrls = Seq(
+          new URL(s"$testRepository/com/abc/test/0.1/test-0.1.pom")
+        )
+        val download = new Download(1, CachePath.defaultCacheDirectory(), Collections.singletonList(
```
alexarchambault, line 83: Same as above:
```suggestion
        val download = new Download(1, dir, Collections.singletonList(
```

modules/bootstrap-launcher/src/it/scala/coursier/bootstrap/launcher/DownloadTests.scala
greg-a-atkinson: Thanks for double checking. Yes, `CachePath.defaultCacheDirectory()` should have been `dir.toFile`. PR updated and checks out.

alexarchambault (approved):
Thanks @greg-a-atkinson!

greg-a-atkinson:
Great. Thank you @alexarchambault

PR #1995, opened 2021-03-09
Allow passing explicit classloader for CacheUrl.url

MasseGuillaume:
This is to be able to define protocol handlers in sbt.

I splitted the package name with the mkString function to avoid
shading the name.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -144,11 +144,13 @@ lazy val cache = crossProject("cache")(JSPlatform, JVMPlatform)
     libraryDependencies ++= Seq(
       Deps.svm % Provided,
       Deps.windowsAnsi
-    )
+    ),
+    classloadersForCustomProtocolTest(customProtocol.jvm),
   )
   .jsSettings(
     name := "fetch-js",
-    libs += Deps.cross.scalaJsDom.value
+    libs += Deps.cross.scalaJsDom.value,
+    classloadersForCustomProtocolTest(customProtocol.js),
   )
   .settings(
     shared,
@@ -175,6 +177,9 @@ lazy val cache = crossProject("cache")(JSPlatform, JVMPlatform)
 lazy val cacheJvm = cache.jvm
 lazy val cacheJs = cache.js
 
+lazy val customProtocol = crossProject("customProtocol")(JSPlatform, JVMPlatform)
+  .settings(dontPublish)
+  
 lazy val scalaz = crossProject("interop", "scalaz")(JSPlatform, JVMPlatform)
   .dependsOn(cache, tests % "test->test")
   .jsConfigure(_.disablePlugins(MimaPlugin))
--- a/build.sbt
+++ b/build.sbt
@@ -145,12 +145,11 @@ lazy val cache = crossProject("cache")(JSPlatform, JVMPlatform)
       Deps.svm % Provided,
       Deps.windowsAnsi
     ),
-    classloadersForCustomProtocolTest(customProtocol.jvm),
+    classloadersForCustomProtocolTest(customProtocolForTest),
   )
   .jsSettings(
     name := "fetch-js",
-    libs += Deps.cross.scalaJsDom.value,
-    classloadersForCustomProtocolTest(customProtocol.js),
+    libs += Deps.cross.scalaJsDom.value
   )
   .settings(
     shared,
@@ -177,8 +176,7 @@ lazy val cache = crossProject("cache")(JSPlatform, JVMPlatform)
 lazy val cacheJvm = cache.jvm
 lazy val cacheJs = cache.js
 
-lazy val customProtocol = crossProject("customProtocol")(JSPlatform, JVMPlatform)
-  .settings(dontPublish)
+lazy val customProtocolForTest = project("custom-protocol-for-test").settings(dontPublish)
   
 lazy val scalaz = crossProject("interop", "scalaz")(JSPlatform, JVMPlatform)
   .dependsOn(cache, tests % "test->test")
--- a/doc/docs/extra.md
+++ b/doc/docs/extra.md
@@ -95,7 +95,7 @@ class CustomprotocolHandler extends URLStreamHandlerFactory {
 
 Coursier will search for your plugin with the following order on various classloaders:
 
-1. The first classloader of those provided via the API (see `FileCache.withClassLoaders`)
+1. Custom classloaders provided via the API (see `FileCache.withClassLoaders`)
 2. `Thread.currentThread().getContextClassLoader`
 3. The classloader that loaded coursier itself (more precisely, `coursier.cache.CacheUrl.getClass.getClassLoader`)
 
--- a/doc/docs/extra.md
+++ b/doc/docs/extra.md
@@ -71,7 +71,7 @@ By default, coursier and sbt-coursier handle the following protocols:
 * `file://`
 * protocols supported by `java.net.URL` (not thoroughly tested). See [javadoc](https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/net/URL.html#%3Cinit%3E(java.lang.String,java.lang.String,int,java.lang.String)) for more details.
 
-Other protocols can be added by implementing `java.net.URLStreamHandler` under the `coursier.cache.protocol` package name:
+Other protocols can be added by implementing `java.net.URLStreamHandlerFactory` under the `coursier.cache.protocol` package name:
 
 ```scala
 package coursier.cache.protocol
@@ -83,23 +83,21 @@ import java.io.FileInputStream
 
 class CustomprotocolHandler extends URLStreamHandlerFactory {
   def createURLStreamHandler(protocol: String): URLStreamHandler = new URLStreamHandler {
-    protected def openConnection(url: URL): URLConnection = {
+    def openConnection(url: URL): URLConnection =
       new URLConnection(url) {
         def connect(): Unit = ()
-        override def getInputStream(): InputStream = {
+        override def getInputStream(): InputStream =
           new FileInputStream(new File(new File(".").getAbsolutePath() + url.getPath()))
-        }
       }
-    }
   }
 }
 ```
 
 Coursier will search for your plugin with the following order on various classloaders:
 
-1. The first classloader of the provided classloaders (see `FileCache.withClassLoaders`)
+1. The first classloader of those provided via the API (see `FileCache.withClassLoaders`)
 2. `Thread.currentThread().getContextClassLoader`
-3. `getClass.getClassLoader`
+3. The classloader that loaded coursier itself (more precisely, `coursier.cache.CacheUrl.getClass.getClassLoader`)
 
 
 Real Word Examples:
--- a/doc/docs/extra.md
+++ b/doc/docs/extra.md
@@ -64,5 +64,44 @@ From sbt, add the setting `coursierUseSbtCredentials := true` for sbt-coursier t
 
 ## Extra protocols
 
-By default, coursier and sbt-coursier handle the `http://`, `https://`, and `file://` protocols. It should also be fine
-by protocols supported by `java.net.URL` (not thoroughly tested). Support for other protocols can be added via plugins. [coursier-s3](https://github.com/paulblei/coursier-s3), a plugin for S3, is under development, and illustrates how to write such plugins.
+By default, coursier and sbt-coursier handle the following protocols:
+
+* `http://`
+* `https://`
+* `file://`
+* protocols supported by `java.net.URL` (not thoroughly tested). See [javadoc](https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/net/URL.html#%3Cinit%3E(java.lang.String,java.lang.String,int,java.lang.String)) for more details.
+
+Other protocols can be added by implementing `java.net.URLStreamHandler` under the `coursier.cache.protocol` package name:
+
+```scala
+package coursier.cache.protocol
+
+import java.io.File
+import java.io.InputStream
+import java.net.{URL, URLConnection, URLStreamHandler, URLStreamHandlerFactory}
+import java.io.FileInputStream
+
+class CustomprotocolHandler extends URLStreamHandlerFactory {
+  def createURLStreamHandler(protocol: String): URLStreamHandler = new URLStreamHandler {
+    protected def openConnection(url: URL): URLConnection = {
+      new URLConnection(url) {
+        def connect(): Unit = ()
+        override def getInputStream(): InputStream = {
+          new FileInputStream(new File(new File(".").getAbsolutePath() + url.getPath()))
+        }
+      }
+    }
+  }
+}
+```
+
+Coursier will search for your plugin with the following order on various classloaders:
+
+1. The first classloader of the provided classloaders (see `FileCache.withClassLoaders`)
+2. `Thread.currentThread().getContextClassLoader`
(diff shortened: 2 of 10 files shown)


alexarchambault (commented):
I think you'll also need to allow to pass the custom class loader to `FileCache` too, that will pass it to `Downloader`, that will itself pass it to `ConnectionBuilder`, that will need to pass it to `CacheUrl.urlConnectionMaybePartial`, so that `urlConnectionMaybePartial` can use it when it calls `CacheUrl.url`.

`FileCache`, `Downloader`, and `ConnectionBuilder`, are all [data classes](https://github.com/alexarchambault/data-class). Adding a field like
```scala
@since("2.0.13")
  classLoaders: Seq[ClassLoader] = Nil
```
should work (and not break binary compatibility).

modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
```diff
@@ -20,21 +20,34 @@ object CacheUrl {
...
-          .orElse(clsOpt(getClass.getClassLoader))
+        val clsOpt0: Option[Class[_]] = {
+          def fromProvided = classLoaders.toStream.map(clsOpt)
+          def fromThread = clsOpt(Thread.currentThread().getContextClassLoader)
+          def fromClass = clsOpt(getClass.getClassLoader)
+
+          fromProvided.append(List(fromThread, fromClass)).collectFirst {
+            case Some(cls) => cls
+          }
+        }
```
alexarchambault, line 50: I'm not sure this lazily evaluates `fromClass`... Maybe we can use iterators instead?
```suggestion
        val clsOpt0: Option[Class[_]] =
          (classLoaders ++ Iterator(Thread.currentThread().getContextClassLoader, getClass.getClassLoader))
            .flatMap(clsOpt(_).iterator)
            .toStream
            .headOption
```

alexarchambault (commented):
Thanks @MasseGuillaume!

I'm wondering whether we can add a test case for that in [this file](https://github.com/coursier/coursier/blob/2ec154561702432f8ca34885c960116576cdfae2/modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala)? (The tests at the end of the file are simpler than the top ones.) Without doing too complicated things... Maybe we can define a custom handler in a new module, returning hard-coded data at some pre-defined URLs, and have sbt pass us the class path of that module via Java properties (which should work if we run those tests by forking).

MasseGuillaume:
Thanks Alex for you help. I will add the test tonight. I have those kinds of tests in the past, it's not too hard with a bunch of sbt magic.

MasseGuillaume:
I added some documentation for the custom protocol handler and the test you mentioned. It was a good idea because my implementation was incorrect. I'm ready for another round of review. Also, if you can take a look at: https://github.com/coursier/sbt-coursier/pull/327  I think it's almost there :)

build.sbt
```diff
@@ -144,11 +144,13 @@ lazy val cache = crossProject("cache")(JSPlatform, JVMPlatform)
...
       Deps.windowsAnsi
-    )
+    ),
+    classloadersForCustomProtocolTest(customProtocol.jvm),
   )
   .jsSettings(
     name := "fetch-js",
-    libs += Deps.cross.scalaJsDom.value
+    libs += Deps.cross.scalaJsDom.value,
+    classloadersForCustomProtocolTest(customProtocol.js),
```
alexarchambault, line 153: This shouldn't be needed, the code you changed is JVM-only.
```suggestion
    libs += Deps.cross.scalaJsDom.value
```

build.sbt
```diff
@@ -175,6 +177,9 @@ lazy val cache = crossProject("cache")(JSPlatform, JVMPlatform)
 lazy val cacheJvm = cache.jvm
 lazy val cacheJs = cache.js
 
+lazy val customProtocol = crossProject("customProtocol")(JSPlatform, JVMPlatform)
```
alexarchambault, line 180: You shouldn't need to compile that in Scala.JS. `FileCache` is JVM-only.

doc/docs/extra.md
```diff
@@ -64,5 +64,44 @@ From sbt, add the setting `coursierUseSbtCredentials := true` for sbt-coursier t
...
-By default, coursier and sbt-coursier handle the `http://`, `https://`, and `file://` protocols. It should also be fine
-by protocols supported by `java.net.URL` (not thoroughly tested). Support for other protocols can be added via plugins. [coursier-s3](https://github.com/paulblei/coursier-s3), a plugin for S3, is under development, and illustrates how to write such plugins.
+By default, coursier and sbt-coursier handle the following protocols:
+
+* `http://`
+* `https://`
+* `file://`
+* protocols supported by `java.net.URL` (not thoroughly tested). See [javadoc](https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/net/URL.html#%3Cinit%3E(java.lang.String,java.lang.String,int,java.lang.String)) for more details.
+
+Other protocols can be added by implementing `java.net.URLStreamHandler` under the `coursier.cache.protocol` package name:
```
alexarchambault, line 74:
```suggestion
Other protocols can be added by implementing `java.net.URLStreamHandlerFactory` under the `coursier.cache.protocol` package name:
```
?

doc/docs/extra.md
```diff
@@ -64,5 +64,44 @@ From sbt, add the setting `coursierUseSbtCredentials := true` for sbt-coursier t
...
+    protected def openConnection(url: URL): URLConnection = {
+      new URLConnection(url) {
+        def connect(): Unit = ()
+        override def getInputStream(): InputStream = {
+          new FileInputStream(new File(new File(".").getAbsolutePath() + url.getPath()))
+        }
+      }
+    }
+  }
+}
```
alexarchambault, line 95:
```suggestion
class CustomprotocolHandler extends URLStreamHandlerFactory {
  def createURLStreamHandler(protocol: String): URLStreamHandler = new URLStreamHandler {
    def openConnection(url: URL): URLConnection =
      new URLConnection(url) {
        def connect(): Unit = ()
        override def getInputStream(): InputStream =
          new FileInputStream(new File(new File(".").getAbsolutePath() + url.getPath()))
      }
  }
}
```
NIT removed some braces for clarity

doc/docs/extra.md
```diff
@@ -64,5 +64,44 @@ From sbt, add the setting `coursierUseSbtCredentials := true` for sbt-coursier t
...
+```
+
+Coursier will search for your plugin with the following order on various classloaders:
+
+1. The first classloader of the provided classloaders (see `FileCache.withClassLoaders`)
```
alexarchambault, line 100:
```suggestion
1. The first classloader of the classloaders provided via the API (see `FileCache.withClassLoaders`)
```

doc/docs/extra.md
```diff
@@ -64,5 +64,44 @@ From sbt, add the setting `coursierUseSbtCredentials := true` for sbt-coursier t
...
+2. `Thread.currentThread().getContextClassLoader`
+3. `getClass.getClassLoader`
```
alexarchambault, line 102:
```suggestion
3. The classloader that loaded coursier itself (more precisely, `coursier.cache.CacheUrl.getClass.getClassLoader`)
```

modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
alexarchambault, line 50: I think the previous changes were reversed
```suggestion
        val clsOpt0: Option[Class[_]] =
          (classLoaders.iterator ++ Iterator(Thread.currentThread().getContextClassLoader, getClass.getClassLoader))
            .flatMap(clsOpt(_).iterator)
            .toStream
            .headOption
```
(there was a typo in my original suggestion too, it was missing `.iterator` in `classLoaders.iterator`)

modules/cache/jvm/src/main/scala/coursier/cache/ConnectionBuilder.scala
```diff
@@ -18,7 +18,9 @@ import dataclass.data
   hostnameVerifierOpt: Option[HostnameVerifier] = None,
   method: String = "GET",
   maxRedirectionsOpt: Option[Int] = Some(20),
-  proxy: Option[Proxy] = None
+  proxy: Option[Proxy] = None,
+  @since("2.0.13")
```
alexarchambault, line 22: It shoud be `2.0.16` now.
```suggestion
  @since("2.0.16")
```

modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala
```diff
@@ -36,7 +36,9 @@ import scala.util.control.NonFatal
   sslSocketFactoryOpt: Option[SSLSocketFactory] = None,
   hostnameVerifierOpt: Option[HostnameVerifier] = None,
   retry: Int = CacheDefaults.defaultRetryCount,
-  bufferSize: Int = CacheDefaults.bufferSize
+  bufferSize: Int = CacheDefaults.bufferSize,
+  @since("2.0.13")
```
alexarchambault, line 40: Same as above
```suggestion
  @since("2.0.16")
```

modules/cache/jvm/src/main/scala/coursier/cache/internal/Downloader.scala
```diff
@@ -36,7 +36,9 @@ import scala.util.control.NonFatal
   sslRetry: Int = CacheDefaults.sslRetryCount,
   sslSocketFactoryOpt: Option[SSLSocketFactory] = None,
   hostnameVerifierOpt: Option[HostnameVerifier] = None,
-  bufferSize: Int = CacheDefaults.bufferSize
+  bufferSize: Int = CacheDefaults.bufferSize,
+  @since("2.0.13")
```
alexarchambault, line 40: Same as above
```suggestion
  @since("2.0.16")
```

modules/cache/jvm/src/main/scala/coursier/cache/internal/Downloader.scala
```diff
@@ -692,9 +696,22 @@ object Downloader {
           case _: javax.net.ssl.SSLException if retry >= 1 =>
             // TODO If Cache is made an (instantiated) class at some point, allow to log that exception.
             None
+
+          case UnknownProtocol(e, msg0) =>
+            val docUrl = "https://get-coursier.io/docs/extra.html#extra-protocols"
+
+            val msg = List(
+              s"Caught ${e.getClass().getName()} (${msg0}) while downloading $url.",
+              s"Visit $docUrl to learn how to handle custom protocols.",
```
alexarchambault, line 705: :+1:

modules/cache/jvm/src/main/scala/coursier/cache/internal/Downloader.scala
```diff
@@ -711,6 +728,17 @@ object Downloader {
     helper(sslRetry)
   }
 
+  private object UnknownProtocol {
+    def unapply(t: Throwable): Option[(MalformedURLException, String)] = t match {
+      case ex: MalformedURLException => 
+        Option(ex.getMessage()) match {
+          case Some(msg) if msg.startsWith("unknown protocol: ") => Some((ex, msg))
+          case _ => None
+        }
```
alexarchambault, line 737: `ex.getMessage` shouldn't be too costly to compute, I guess the following should be fine:
```suggestion
      case ex: MalformedURLException if ex.getMessage.startsWith("unknown protocol: ") => 
        Some((ex, ex.getMessage))
```

alexarchambault: We could even not return the message here, and call `getMessage` when we match `UnknownProtocol` above.

modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala
```diff
@@ -926,6 +926,69 @@ object FileCacheTests extends TestSuite {
...
+            "Visit https://get-coursier.io/docs/extra.html#extra-protocols to learn how to handle custom protocols."
+          ).mkString(" ")
+
+        res match {
+          case Left(error: ArtifactError.DownloadError) =>
+            println("---")
+            println(error.reason)
+            println()
+            println(expectedReason)
+            println("---")
```
alexarchambault, line 952: Debugging stuff?
```suggestion
```

modules/cache/jvm/src/test/scala-2.12/coursier/cache/FileCacheTests.scala
```diff
@@ -926,6 +926,69 @@ object FileCacheTests extends TestSuite {
...
+              FileCache()
+                .withClassLoaders(Seq(classloader))
+                .withLocation(dir.toFile)
+                .file(artifact)
+                .run
+                .future
+            }
+
+            res match {
+              case Right(file: File) =>
```
alexarchambault, line 980: No need of the type ascription, it could make that code still compile if we were to change the return type of `FileCache.file`, which we don't want:
```suggestion
              case Right(file) =>
```

modules/customProtocol/shared/src/main/scala/coursier/cache/protocol/CustomprotocolHandler.scala
```diff
@@ -0,0 +1,19 @@
...
+import java.net.{URL, URLConnection, URLStreamHandler, URLStreamHandlerFactory}
+import java.io.FileInputStream
+
+class CustomprotocolHandler extends URLStreamHandlerFactory {
+  def createURLStreamHandler(protocol: String): URLStreamHandler = new URLStreamHandler {
+    protected def openConnection(url: URL): URLConnection = {
+      new URLConnection(url) {
+        def connect(): Unit = ()
+        override def getInputStream(): InputStream = {
+          new FileInputStream(new File(new File(".").getAbsolutePath() + url.getPath()))
```
alexarchambault, line 14: Going from `File` to `String` to `File` again seems odd. Wouldn't the following work?
```suggestion
          new FileInputStream(new File(new File("."), url.getPath.stripPrefix("/")))
```

alexarchambault: The `stripPrefix` I added may even not be needed.

project/Settings.scala
```diff
@@ -557,4 +557,21 @@ object Settings {
 
   lazy val javaMajorVer = sys.props("java.version").takeWhile(_.isDigit).toInt
 
+  def classloadersForCustomProtocolTest(fromProject: Project): Setting[Seq[Task[Seq[java.io.File]]]] = 
+    Test / sourceGenerators += Def.task {
+      val customLoaderClasspath = (fromProject / Compile / fullClasspath).value
+      val dq = '"'
+      val files = customLoaderClasspath.files.map(f => 
+        dq + f.toURI().toURL().toString + dq
```
alexarchambault, line 565: The `toURL` is unnecessary here
```suggestion
        dq + f.toURI.toString + dq
```

alexarchambault (commented):
Thanks @MasseGuillaume! Left some more comments.

MasseGuillaume:
Thanks for the feedback! I rebased all the commits and the CI is running at: https://github.com/coursier/coursier/runs/2167946155

alexarchambault (approved):

alexarchambault:
Merging, thanks!

PR #2007, opened 2021-03-17
Allow search in channels for installable apps

### LINKED ISSUE #1993: list of applications
Hi there, I'm wondering the only way how to get a list of installable apps is via:
```
unzip -l "$(cs fetch io.get-coursier:apps:1.0.7)"
```   

If so, wouldn't be good to add a command line option that can display this list? I mean a newbie (or even a user that do not use `cs` on daily basis) will not remember `unzip -l "$(cs fetch io.get-coursier:apps:1.0.7)"` (not very user friendly)

rrramiro:
Hi there, I tried to implement `cs search` command to allow to list install-able applications.
@alexarchambault I would appreciate more directions.

Fixes #1993

### DIFF
--- a/modules/cli/src/main/scala/coursier/cli/Coursier.scala
+++ b/modules/cli/src/main/scala/coursier/cli/Coursier.scala
@@ -19,6 +19,7 @@ import coursier.cli.launch.Launch
 import coursier.cli.publish.Publish
 import coursier.cli.resolve.Resolve
 import coursier.cli.setup.{Setup, SetupOptions}
+import coursier.cli.search.Search
 import coursier.core.Version
 import coursier.install.InstallDir
 import coursier.launcher.internal.{FileUtil, Windows}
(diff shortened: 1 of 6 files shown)


alexarchambault (approved):
Thanks!

I guess `Search` could be moved up in `Coursier.scala` and `CoursierCommand.scala` to preserve the ordering, and some things could be factored in `Channels.scala`, but this can be fixed later...

rrramiro:
Thanks, sorry, I haven't notice the order ...

alexarchambault:
> Thanks, sorry, I haven't notice the order ...

No problem, some things need to be fixed in case-app, so that we don't have these annoying matches in `Coursier.scala`, and preserving that order becomes less of a hindrance...

xhudik:
@rrramiro  - so the command should look like: `cs search` and it will output a list of apps that can be installed via `coursier`?

rrramiro:
Yes, running `cs search` with no argument displays all available applications in the [Main channel](https://github.com/coursier/apps).

`cs search` takes mostly the same options as `cs complete`.
The only difference is `--scala-version`, which it's not used for the search.
For example, it allows to get all available apps from the "Main channel" and the "Contrib channel" with the `--contrib` option:
```
$ cs search --contrib
```

Any arguments filter the result if they're contained in the application name.
For example, the following command lists all applications having `meta` or `fmt` in their names
```
$ cs search meta fmt
metac
metals
metals-emacs
metap
metap-native
scalafmt
```

PR #2012, opened 2021-03-21
Document how to get/set the cache directory in SBT

ScalaWilliam:
https://github.com/coursier/coursier/issues/1483

https://github.com/sbt/sbt/issues/5120

Doing this as most users informing me about cache location confusion use Coursier through SBT, rather than standalone.

### DIFF
--- a/doc/docs/cache.md
+++ b/doc/docs/cache.md
@@ -4,7 +4,7 @@ title: Cache
 
 ## Path format
 
-The cache is supposed to cache static resources, that don't change often (snapshot artifacts being the exception - these then follow TTL).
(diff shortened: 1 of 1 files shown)


alexarchambault (approved):
Thanks!

ScalaWilliam:
Hi @alexarchambault just noticed this is still not updated on the Coursier website, any ideas if there's another step?

ScalaWilliam:
@alexarchambault

alexarchambault:
It didn't end up on the website, as something went wrong during the last release (issues generating the Windows launcher). These should be fixed in the next release, and the website should be updated then.

PR #2017, opened 2021-03-25
Refactoring pattern matching coproduct options

rrramiro:
Hi,
I wanted to fix the order of the commands and I found a way to do the "pattern matching" with coproducts.
I understood that this part will be changed anyway. But I wanted to share it.


alexarchambault (approved):
Nice, thanks!

PR #2020, opened 2021-03-31
Fixes Resolution to contain actual version

### LINKED ISSUE #2009: Resolution contains version range
## steps

Resolve `com.nimbusds:nimbus-jose-jwt:4.41.1`

## problem

`Resolution#dependencyArtifacts` returns `Dependency` value with version range inside.

## expectation

As a dependency resolver used in various tooling, my expectation is that Coursier would return the metadata of the final resolution in `Resolution`, not just the URLs but also the `Dependency` information with concrete version number that it settled on.

## note

For now I'm working around this using this function

```scala
  def actualDependency(d0: Dependency, projectCache: ProjectCache): Dependency =
    projectCache.get(d0.moduleVersion) match {
      case Some((_, p)) => d0.withVersion(p.actualVersion)
      case _            => d0
    }
```

but I don't think we should have to do this.

eed3si9n:
Fixes #2009

Problem
--------
Resolution#dependencyArtifacts returns Dependency value with interval inside.

Solution
--------
Since Project contains the actual version, this make the change to copy that value in.

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Resolution.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Resolution.scala
@@ -1286,7 +1286,7 @@ object Resolution {
           Some(classifiers.getOrElse(Nil) ++ Seq(dep.attributes.classifier))
 
       (pub, artifact) <- source.artifacts(dep, proj, classifiers0)
-    } yield (dep, pub, artifact)
+    } yield (dep.withVersion(proj.actualVersion), pub, artifact)
 
 
   @deprecated("Use the artifacts overload accepting types and classifiers instead", "1.1.0-M8")
--- a/modules/coursier/shared/src/test/scala/coursier/ResolveTests.scala
+++ b/modules/coursier/shared/src/test/scala/coursier/ResolveTests.scala
@@ -217,6 +217,25 @@ object ResolveTests extends TestSuite {
             await(validateDependencies(res))
           }
 
+          test("resolveConcreteVersion") - async {
+
+            val res = await {
+              resolve0
+                .addDependencies(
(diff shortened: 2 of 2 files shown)


alexarchambault (approved):
Thanks @eed3si9n!

alexarchambault:
Sorry, I might approved that too early. The error seems legit (I can reproduce it locally with `sbt cli/testOnly coursier.cli.FetchTests`). Feel free to have a look. I can find time to look at it myself later this week else.

eed3si9n:
```
X coursier.cli.FetchTests.grpc-core should have dependencies 79ms 
  utest.AssertionError: assert(grpcCoreNode.dependencies.toSet == Set(
  grpcCoreNode: coursier.cli.util.DepNode = DepNode(io.grpc:grpc-core:1.29.0,Some(/home/runner/.cache/coursier/v1/https/repo1.maven.org/maven2/io/grpc/grpc-core/1.29.0/grpc-core-1.29.0.jar),Set(),Set(),Set())
    utest.asserts.Asserts$.assertImpl(Asserts.scala:30)
    coursier.cli.FetchTests$.$anonfun$tests$211(FetchTests.scala:998)
    coursier.cli.FetchTests$.$anonfun$tests$211$adapted(FetchTests.scala:982)
    coursier.cli.TestUtil$.withFile(TestUtil.scala:18)
    coursier.cli.FetchTests$.$anonfun$tests$210(FetchTests.scala:982)
```

https://github.com/coursier/coursier/blob/d2e1b696de1642178cdad432956216eb6996bb95/modules/cli/src/test/scala/coursier/cli/FetchTests.scala#L998-L1012

Yea, so I probably broke something.

PR #2024, opened 2021-04-03
Handle gzip content encoding

### LINKED ISSUE #1281: Support gzip content encoding
Currently when the server responds with a gzip encoded stream, files get saved compressed and don't pass the checksum afterwards so coursier exits with an error.

It's probably happening because `HttpURLConnection` doesn't seem to do anything with gzip encoding so `HttpURLConnection.getInputSteam` needs to be wrapped with `GZIPInputStream` if the gzip encoding is present in the response headers.

I'm using coursier via [`rules_jvm_external`](https://github.com/bazelbuild/rules_jvm_external) if that's relevant.

kitterion:
Fixes #1281

I have a server that responds with gzip encoded content and it breaks coursier because checksums don't match after downloading. I thought I'd fix it here instead of maintaining a locally modified version.

### DIFF
--- a/modules/cache/jvm/src/main/scala/coursier/cache/internal/Downloader.scala
+++ b/modules/cache/jvm/src/main/scala/coursier/cache/internal/Downloader.scala
@@ -243,10 +243,11 @@ import scala.util.control.NonFatal
 
           val lastModifiedOpt = Option(conn.getLastModified).filter(_ > 0L)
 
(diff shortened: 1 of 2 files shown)


alexarchambault (approved):
Thanks!

I just pushed an extra commit, to fix minor style-related points.

kitterion:
No worries, I'm not very familiar with scala, thanks for fixing it.

PR #2025, opened 2021-04-05
Use JNI rather than PowerShell to do Windows system calls

alexarchambault:


alexarchambault:
(I'd like to setup some Windows CI test jobs prior to merging that.)

eed3si9n:
Thank you for working on this on behalf of all the Windows users :)

PR #2035, opened 2021-05-02
Improve performance of json generation

kitterion:
I am using coursier via [rules_jvm_external](https://github.com/bazelbuild/rules_jvm_external), I added a dependency and was very surprised to see coursier failing with a timeout. It turned out, the dependency graph size increased from about 200 nodes to about 400, and coursier started eating multiple gigs of memory and grinding to a halt during json generation.

The code doing the recursive walk of the dependency tree is inefficient. It tries for every node to calculate the set of its transitive dependencies. As far as I understand, it does a DFS from every node in parallel. Without any memoization it recalculates transitive deps of pretty much every node quite a lot of times.
I essentially added that missing memoization, now for each node it remembers the set of its transitive deps. It drastically reduces memory consumption and runtime on large dependency graphs.

Also, I'm not very familiar with scala, so feel free to point out more idiomatic ways of doing things.


alexarchambault:
Merged via https://github.com/coursier/coursier/pull/2105.

PR #2043, opened 2021-05-16
Update ttl.md

unkarjedy:
... based on coursier implementation:
https://github.com/coursier/coursier/blob/d5ad55d1dcb025084ba9bd994ea47ceae0608a8f/modules/cache/jvm/src/main/scala/coursier/cache/CacheDefaults.scala#L42
and sbt shading rules:
https://github.com/coursier/sbt-coursier/blob/master/build.sbt#L57


alexarchambault (approved):
Thanks, and sorry for the (very) late review :|

(For the anecdote, I've been wanting to try your PR locally shortly after you opened it, but had issues with running the website locally at the time… which was the last straw that made me want to [switch the build of coursier to Mill](https://github.com/coursier/coursier/pull/2067), which took a bit of time.)

I'm going to merge, I only pushed a minor change on top of your changes.

PR #2050, opened 2021-05-26
Python native libs

kiendang:
I changed a few things regarding the python native libs.

 - Use `sys.exec_prefix` instead of `sys.prefix`. Often they are the same but when they are not, [exec_prefix](https://docs.python.org/3/library/sys.html#sys.exec_prefix) is where the native libs are located.
 - Let python handle `PYTHONHOME` itself. So basically call `PYTHONHOME=<...> python -c "import sys;print(sys.exec_prefix)"` instead of dealing with it in coursier since [sometimes it gets complicated](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONHOME).
 - `exec_prefix` can be shared among multiple python installations. Without further specification ScalaPy might pick up the wrong python libs. For example if the `python` on `PATH` is 3.9 but there is a `python3.7` also installed with the same `exec_prefix`, ScalaPy would pick up the `python3.7` lib [instead](
https://github.com/shadaj/scalapy/blob/6e4f09dbed07ea5bf55f2cd086c7929276c08b8b/core/jvm/src/main/scala/me/shadaj/scalapy/interpreter/CPythonAPI.scala#L10-L24). To fix this I get the correct name of the python lib (`python` + `LDVERSION`) and set it to `SCALAPY_PYTHON_LIBRARY`.
- Users can specify another python other than the one on `PATH` by setting the [PYTHONEXECUTABLE](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONEXECUTABLE) env var.

I did went down the `sysconfig.get_config_var('LIBPL')` path but turned out it would complicate things and is not necessary anw.

### DIFF
--- a/modules/cli/src/main/scala/coursier/cli/launch/Launch.scala
+++ b/modules/cli/src/main/scala/coursier/cli/launch/Launch.scala
@@ -289,7 +289,11 @@ object Launch extends CaseApp[LaunchOptions] {
             .asScala
             .map(e => (e.getKey, e.getValue))
             .toVector
-          (props, EnvironmentUpdate(Seq("PYTHONHOME" -> home), Nil))
+          val envVars =
+            if (params.shared.python)
+              Seq("PYTHONHOME" -> home, "SCALAPY_PYTHON_LIBRARY" -> Jep.pythonLDLibrary())
+            else Seq("PYTHONHOME" -> home)
+          (props, EnvironmentUpdate(envVars, Nil))
         } catch {
           case NonFatal(e) =>
             if (params.shared.resolve.output.verbosity >= 1)
--- a/modules/paths/src/main/java/coursier/paths/Jep.java
+++ b/modules/paths/src/main/java/coursier/paths/Jep.java
@@ -118,6 +118,48 @@ public class Jep {
     return version;
   }
 
+  public static String pythonExecutable() throws Exception {
+    String fromEnv = System.getenv("PYTHONEXECUTABLE");
+    if (fromEnv != null && !fromEnv.isEmpty() && existsInPath(fromEnv))
+      return fromEnv;
+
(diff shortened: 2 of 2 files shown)


alexarchambault (commented):
Nice work!

We might want to unit test that logic if it were to be tweaked again…

alexarchambault:
About unit testing this, I was thinking about something along those lines:
- move the python-related methods to their own class
- make them non-static
- pass arguments such as `Map<String, String> env`, `Map<String, String> props`, `Function2<List<String>, Map<String, String>, String> callProcess`, `boolean isWindows` to the new class, and maybe even `java.nio.file.FileSystem fs`
- use those arguments in the python logic
- add back static methods, creating a class instance with sensible defaults for the arguments, calling the non-static methods.

Then in unit tests, we can create class instances with mock environment /  properties, returning mock output when some processes are called, and maybe even use mock file systems, and check that we get the right python paths. [These tests](https://github.com/coursier/coursier/blob/2af4130839249929680bcedee037f45ee47c6b64/modules/jvm/src/test/scala/coursier/jvm/JavaHomeTests.scala) follow a similar philosophy, using [a mock environment, a mock process runner](https://github.com/coursier/coursier/blob/2af4130839249929680bcedee037f45ee47c6b64/modules/jvm/src/test/scala/coursier/jvm/JavaHomeTests.scala#L90-L92).

The tests can be written in Scala, and added to the [`bootstrap-launcher` tests](https://github.com/coursier/coursier/tree/master/modules/bootstrap-launcher/src/test) for example.

We may not want to write those right now, as this might be a bit cumbersome overall… Maybe this can be done if those python methods are to be tweaked again in the coming weeks.

kiendang:
Superseded by #2532

PR #2051, opened 2021-05-26
Homebrew set depends on openjdk for all macOS versions

kiendang:
`depends_on java` has been deprecated so just `depends_on openjdk` for all mac versions now. Plus minor styling stuff from `brew audit`.


alexarchambault (approved):
Thanks @kiendang!

PR #2064, opened 2021-06-25
docs: sbt uses coursier by default nowadays

raboof:
Some of the docs were still written from the perspective that
sbt only uses coursier via the plugin. Moved things around a bit
to better reflect the fact that on recent versions of sbt, coursier
is the default (while still keeping the old instructions for sbt <1.3.x
users, but less prominently).

Unfortunately I failed to build the docs locally, so I haven't
checked the rendering.


raboof:
(build failure seemed unrelated, rebased since master seems green now)

alexarchambault (approved):
Thanks @raboof!

PR #2067, opened 2021-06-30
Switch build to Mill

alexarchambault:


alexarchambault:
(Need to update / add developer doc prior to merging this.)

DEVELOP.md
```diff
@@ -0,0 +1,141 @@
...
+```
+
+### IDE
+
+#### Metals
+
+[Metals](https://scalameta.org/metals) is the recommended IDE to develop on coursier.
+
+Using the default import functionality of Metals is _not_ recommended. Instead,
+generate bloop configuration files with
```
ckipp01, line 38: I've seen you write this in a couple different repos now. Is there anything on the Metals side that we could do to improve the import situation with Mill? What makes you strongly recommend not relying on the built-in functionality? I'd love to improve this so we don't have to recommend users not to utilize the built-in functionality. Let me know your thoughts @alexarchambault .

Issue #1969, opened 2021-02-14
Empty User-Agent header can trigger WAF blocking


q3k:
Coursier seems to set an empty 'User-Agent' header in the requests it makes to external repositories:

```
$ coursier resolve -r http://127.0.0.1:1337/nexus/content/repositories/snapshots org.spigotmc:spigot-api:1.15.2-R0.1-20200624.001023-124
````

```
$ nc -l 1337
GET /nexus/content/repositories/snapshots/org/spigotmc/spigot-api/1.15.2-R0.1-SNAPSHOT/spigot-api-1.15.2-R0.1-20200624.001023-124.pom HTTP/1.1
User-Agent: 
Host: 127.0.0.1:1337
Accept: text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2
Connection: keep-alive
```

This can trigger (web application) firewalls and make them reject the requests, as some security scanning automation out there seems to use an empty UA string. For example, this is the case with `hub.spigotmc.org`: `coursier resolve -r https://hub.spigotmc.org/nexus/content/repositories/snapshots org.spigotmc:spigot-api:1.15.2-R0.1-20200624.001023-124` fails with a 403 due to a CloudFlare configuration on their side.

I think it would be advisable for coursier to set the User-Agent string to some non-zero value (and maybe identify itself as coursier at a given version), or at least for users to be able to override what UA is set.

alexarchambault:
IIRC, this was made early in the development of coursier, because of Maven Central or maybe some other Sonatype repository not returning the right content with the default User-Agent. Checking in https://github.com/coursier/coursier/pull/1974 if keeping the default User-Agent breaks anything.

Issue #1649, opened 2020-04-06
Can't use Microsoft Azure DevOps Maven Repository as Resolver 


mantovani:
I don't want to do cross-posting but looks like there's a fix need from Coursier as well,

https://github.com/sbt/sbt/issues/5492

eed3si9n:
Here are steps and problem from https://github.com/sbt/sbt/issues/5492

## steps

Add a Microsoft [Azure DevOps Maven](https://docs.microsoft.com/en-us/azure/devops/artifacts/get-started-maven?view=azure-devops) repository as resolver:

```sbt
lazy val scala211 = "2.11.12"

ThisBuild / organization := "com.foo"
ThisBuild / scalaVersion := scala211
ThisBuild / useCoursier := true

name := "baz"
crossScalaVersions := Seq(scala211)
libraryDependencies ++= Seq(
  "foo" %% "bar" % 0.1 % "provided"
)
resolvers +=   "nabu-releases" at "https://pkgs.dev.azure.com/foo/Foo/_packaging/foo-releases/maven/v1"
credentials += Credentials(Path.userHome / ".sbt" / ".credentials")
```

sbt version
```
[info] This is sbt 1.3.9
```

## problem

Coursier throws:

```
[warn]         Note: Unresolved dependencies path:
[error] stack trace is suppressed; run last update for the full output
[error] (update) sbt.librarymanagement.ResolveException: Error downloading com.foo:bar_2.11:1.0.0
[error]   Not found
[error]   Not found
[error]   not found: /Users/mantovani/.ivy2/local/com.foo/bar_2.11/1.0.0/ivys/ivy.xml
[error]   not found: https://repo1.maven.org/maven2/com/foo/bar_2.11/1.0.0/bar_2.11-1.0.0.pom
[error]   checksum format error: /Users/mantovani/Library/Caches/Coursier/v1/https/pkgs.dev.azure.com/Foo-backend/Foo/_packaging/Foo-releases/maven/v1/com/foo/bar_2.11/1.0.0/bar_2.11-1.0.0.pom.sha1
```

## notes

```
sbt:foo> show credentials
[info] * FileCredentials("/Users/mantovani/.sbt/sonatype_credential")
[info] * FileCredentials("/Users/mantovani/.sbt/.credentials")
```

```
cat /Users/mantovani/.sbt/.credentials

realm=foo
host=pkgs.dev.azure.com
user=foo
password=xxx
```

The checksum actually has a HTML from Azure:

```
[...]

PR #2068, opened 2021-07-03
Allow users to reject 203 responses

markoutso:
Microsoft (Azure) proxies return a 203 response when unathenticated.
The response contains html code that prompts the user to login.

In this the pull request there's also code that propagates the auth realm to downstream functions.


alexarchambault:
Thanks for opening this!

At first I'm a bit wary to enable this by default… My understanding of the [203 status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/203) is that the body may have been modified by a proxy, but is still valid… (My guess is it could be used if a proxy reformats checksum files or directory listings for example, while preserving their actual value.) It seems the Azure servers are faulty here.

If this feature were to be added, I think it'd be better to enable it manually, either for specific repositories or all requests.

That said, it seems tweaking the User Agent may just solve that, per https://github.com/sbt/sbt/issues/5483. I'd like to preserve the default Java user agent, as clearing it has been a problem in other contexts (see https://github.com/coursier/coursier/issues/1969). Maybe that'll just solve the issue here. It could be made customizable too.

alexarchambault:
@markoutso Is there a public repository that could be used to reproduce that?

markoutso:
I understand your hesitation.
Having this as an opt-in feature was my initial intention but after some thought I ended up believing that it's a bad thing when an artifact proxy is telling you that the payload is not the original one.

In any case, I will update the pull request.

I don't think it's a matter of user-agent though.
If you configure the repository with the correct authentication then you can download without any problems as I described [here](https://github.com/coursier/coursier/issues/1649#issuecomment-851058821).

markoutso:
@alexarchambault Any news on this?

modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
```diff
@@ -231,6 +225,7 @@ object CacheUrl {
     credentials: Seq[DirectCredentials] = Nil,
     sslSocketFactoryOpt: Option[SSLSocketFactory] = None,
     hostnameVerifierOpt: Option[HostnameVerifier] = None,
+    authRealmOpt: Option[String] = None,
```
alexarchambault, line 228: Adding this one shouldn't be needed. It's a deprecated method, that shouldn't be used anymore.

modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
```diff
@@ -277,6 +273,7 @@ object CacheUrl {
     autoCredentials: Seq[DirectCredentials],
     sslSocketFactoryOpt: Option[SSLSocketFactory],
     hostnameVerifierOpt: Option[HostnameVerifier],
+    authRealmOpt: Option[String],
```
alexarchambault, line 276: Same as above.

modules/cache/jvm/src/main/scala/coursier/cache/ConnectionBuilder.scala
```diff
@@ -16,6 +16,7 @@ import dataclass.data
   autoCredentials: Seq[DirectCredentials] = Nil,
   sslSocketFactoryOpt: Option[SSLSocketFactory] = None,
   hostnameVerifierOpt: Option[HostnameVerifier] = None,
+  authRealmOpt: Option[String] = None,
```
alexarchambault, line 19: This field should be added in last position, after a `@since("2.0.17")` annotation, not to break binary compatibility.

modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala
```diff
@@ -31,10 +31,12 @@ import scala.util.control.NonFatal
   localArtifactsShouldBeCached: Boolean = false,
   followHttpToHttpsRedirections: Boolean = true,
   followHttpsToHttpRedirections: Boolean = false,
+  rejectNonAuthoritativeResponses: Boolean = false,
   maxRedirections: Option[Int] = CacheDefaults.maxRedirections,
   sslRetry: Int = CacheDefaults.sslRetryCount,
   sslSocketFactoryOpt: Option[SSLSocketFactory] = None,
   hostnameVerifierOpt: Option[HostnameVerifier] = None,
+  authRealmOpt: Option[String] = None,
```
alexarchambault, line 39: Just like for `ConnectionBuilder` above, these new fields should be added in last position, and after a `@since("2.0.17")` annotation, not to break binary compatibility.

modules/cache/jvm/src/main/scala/coursier/cache/internal/Downloader.scala
```diff
@@ -33,10 +33,12 @@ import scala.util.control.NonFatal
   localArtifactsShouldBeCached: Boolean = false,
   followHttpToHttpsRedirections: Boolean = true,
   followHttpsToHttpRedirections: Boolean = false,
+  rejectNonAuthoritativeResponses: Boolean = false,
   maxRedirections: Option[Int] = CacheDefaults.maxRedirections,
   sslRetry: Int = CacheDefaults.sslRetryCount,
   sslSocketFactoryOpt: Option[SSLSocketFactory] = None,
   hostnameVerifierOpt: Option[HostnameVerifier] = None,
+  authRealmOpt: Option[String] = None,
```
alexarchambault, line 41: Just like above, these should be in last position and after a `@since("2.0.17")`.

alexarchambault (commented):
Thanks! I only added comments about binary compatibility (which can be preserved with only minor modifications). [`DEVELOP.md`](https://github.com/coursier/coursier/blob/99576bdfb87d18dceefc1800f4fe195726dfba19/DEVELOP.md#check-binary-compatibility) indicates how to check it locally. You can also add `-w` right after `./mill` to check it in watch mode, while you are editing your code.

PR #2105, opened 2021-07-07
Improve json generation performance (tweaked #2035)

alexarchambault:
Includes https://github.com/coursier/coursier/pull/2035

### DIFF
--- a/modules/cli/src/main/scala/coursier/cli/util/JsonReport.scala
+++ b/modules/cli/src/main/scala/coursier/cli/util/JsonReport.scala
@@ -59,29 +59,27 @@ object JsonReport {
   def apply[T](roots: IndexedSeq[T], conflictResolutionForRoots: Map[String, String])
               (children: T => Seq[T], reconciledVersionStr: T => String, requestedVersionStr: T => String, getFile: T => Option[String], exclusions: T => Set[String]): String = {
 
-    val rootDeps: ParSeq[DepNode] = roots.par.map(r => {
+    var depToTransitiveDeps = mutable.Map[T, Set[String]]()
+    val rootDeps: Seq[DepNode] = roots.map(r => {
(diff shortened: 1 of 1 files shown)


modules/cli/src/main/scala/coursier/cli/util/JsonReport.scala
```diff
@@ -59,33 +59,26 @@ object JsonReport {
...
-          if (depElems.nonEmpty)
-            flattenDeps(children(elem), ancestors + elem, acc)
-        }
+    val depToTransitiveDeps = new mutable.HashMap[T, Set[String]]
+    def flattenDeps(elem: T): Set[String] =
+      if (depToTransitiveDeps.contains(elem))
+        depToTransitiveDeps(elem)
+      else {
+        val children0 = children(elem)
+        val deps = children0.map(reconciledVersionStr(_)).toSet ++ children0.iterator.flatMap(flattenDeps(_))
```
kitterion, line 68: In my original PR I thought about using `.flatMap` on `children0` but decided against it because it seems to require O(n^2) memory on dense graphs where n is the size of children0 and the whole point was to minimize memory consumption.
My logic is that in a fully connected directed acyclic graph `.flatMap` will generate a sequence of size (n-2) + (n-3) + ... + 1 + 0 = (n-2)(n-3)/2 ~ n^2, and only after generating that sequence will it collapse while being converted to `Set`.

alexarchambault: Yes, but we flatMap on an iterator here, so no large collection actually gets built here, I believe. `flattenDeps` gets called on the first element of `children0`, the corresponding set is added to the one being built for `deps` and discarded straightaway, then the next element of `children0` is processed, etc.

PR #2107, opened 2021-07-07
Be fine with Mill-generated manifests

alexarchambault:
Mill doesn't add Implementation-Vendor-Id and Specification-Title in the
manifests it generates, which the main class detection code uses to pick
a main class.

These changes default to the main class from the first JAR in the class
path (which has been determistic for some time now, and corresponds to the
JAR of the first specified dependency), no matter its
Implementation-Vendor-Id and Specification-Title attributes.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -506,6 +506,9 @@ trait Cli extends CsModule with CoursierPublishModule with Launchers {
     ivy"com.chuusai::shapeless:2.3.7"
   )
   def mainClass = Some("coursier.cli.Coursier")
+  def finalMainClassOpt = T{
+    Right("coursier.cli.Coursier"): Either[String, String]
+  }
   def manifest = T{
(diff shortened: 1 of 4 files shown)


jorhett:
I'm seeing this still on 20.04, on the current release downloaded from the launchers repo:

```shell
$ /tmp/coursier_installerbw_3tcf1/cs setup --yes
Checking if a JVM is installed
Found a JVM installed under /usr/lib/jvm/java-11-openjdk-arm64.

Checking if ~/.local/share/coursier/bin is in PATH
Should we add ~/.local/share/coursier/bin to your PATH via ~/.profile, ~/.bash_profile? [Y/n] Y

Checking if the standard Scala applications are installed
  Found ammonite
  Found cs
  Found coursier
  Found scala
  Found scalac
  Found sbt
No main class found
```

PR #2154, opened 2021-08-16
support fish shell (draft)

pektinasen:
I basically copy-pasted ProfileUpdater and changed the `set` and `update` texts. 

I consider this work in progress and would like to get some feedback what needs to change to get this merged. Some things I'm not satisfied about:

*  `Either[WindowsEnvVarUpdater, Either[ProfileUpdater, FishUpdater]]` -> maybe change it to ADT?
* `def isFish()` not sure where to put it, and if this is the best approach
* Lots of duplication when pattern matching the eithers
* I'm not really comfortable with java.nio. So the whole line about finding the path to "config.fish" is a bit weird.

Did I miss something where a distinction between the shells is also necessary?


alexarchambault (commented):
Thanks for opening this!

I'm fine adding support for editing fish profile files, but there's quite a bit of duplication in the current state of this PR, to the point that I have a hard time figuring out what was specifically added to the cut-n-pasted code.

Wouldn't just handling fish in [`ProfileUpdater`](https://github.com/coursier/coursier/blob/a2142470fc063bc1f143f9eac4bdacb8806dfd0e/modules/env/src/main/scala/coursier/env/ProfileUpdater.scala) itself work? detecting if fish is used from there, just like [how it's done for zsh](https://github.com/coursier/coursier/blob/a2142470fc063bc1f143f9eac4bdacb8806dfd0e/modules/env/src/main/scala/coursier/env/ProfileUpdater.scala#L24).

PR #2188, opened 2021-09-19
Add ArchiveCache, rework JvmCache to use it

alexarchambault:


jneira:
hi, first of al lthanks for the work done here
This pr removed the `cs java --installed` option, where could i find the rationale behind? It seems a useful option and [i even started to add it again](https://github.com/jneira/coursier/tree/add-java-installed) before being aware it was removed here

PR #2197, opened 2021-09-25
Recheck JVM archive extraction directory after acquiring the file lock.

patricklaw:
The existing locking code for extracting JVM archives unconditionally attempts the extraction after acquiring the file lock for the extracted directory, but the check for whether that directory exists happens before the lock is acquired.  This means that while waiting for the lock (or losing the race to get it), another process could successfully extract the archive.  We then successfully acquire the lock, but the extraction fails because the directory now exists.

I believe this should fix the root cause of https://github.com/pantsbuild/pants/issues/12293.

### DIFF
--- a/modules/jvm/src/main/scala/coursier/jvm/JvmCache.scala
+++ b/modules/jvm/src/main/scala/coursier/jvm/JvmCache.scala
@@ -50,39 +50,48 @@ import scala.util.control.NonFatal
     tmpDir: File
   ) =
     withLockFor(dir) {
-      logger0.extracting(entry.id, archive.getAbsolutePath, dir)
-      val dir0 =
-        try {
+      // Check again that the entry doesn't exist, since another process might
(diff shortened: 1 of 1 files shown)


alexarchambault (approved):
Nice catch, thanks @patricklaw!

patricklaw:
Thanks for merging, Alex!  We'd really like to test this in Pants to see if it resolves our CI issues, but after a bit of poking I can't figure out how to cut a binary release of Coursier.  Is that a workflow that you run, or am I missing some documentation somewhere?

alexarchambault:
I cut `2.0.16-157-gcef6b2330` yesterday, but a new issue came up. Hopefully, the whole release will work [now](https://github.com/coursier/coursier/actions/runs/1311703714).

stuhood:
Thanks a lot! We'll try consuming it and report back.

PR #2231, opened 2021-10-19
Introduce `VersionOverride`

### LINKED ISSUE #2195: Merge Scala 3 related apps with Scala 2 related apps
Currenlty the user has to use different app names when using Scala 3 versus Scala 2.

```
$ cs launch scala:2.12.14 -- -version
Scala code runner version 2.12.14 -- Copyright 2002-2021, LAMP/EPFL and Lightbend, Inc.
```

```
$ cs launch scala3-repl:3.0.1 -- -version
Scala code runner version 3.0.1 -- Copyright 2002-2021, LAMP/EPFL
```

We should allow to launch Scala 3 using the name `scala`, since Scala 3 is nothing but the latest version of Scala.
A similar issue exists with `scalac` and `scaladoc`.

This change implies that Scala 3 will be installed by default. That should not prevent to launch, or install, Scala 2 with `cs launch scala:2.13.6 `.

adpi2:
Fixes #2195 

A `VersionOverride` is a way to override the app descriptor based on the app version.
This change is introduced so that we can merge `scalac` and `scala3-compiler` into the same app descriptor:
```json
{
  "mainClass": "dotty.tools.dotc.Main",
  "repositories": [
    "central"
  ],
  "dependencies": [
    "org.scala-lang:scala3-compiler_3:latest.stable"
  ],
  "properties": {
    "scala.usejavacp": "true"
  },
  "versionOverrides": [
    {
      "versionRange": "(,2.max]",
      "mainClass": "scala.tools.nsc.Main",
      "repositories": [
        "central"
      ],
      "dependencies": [
        "org.scala-lang:scala-compiler:latest.stable"
      ],
      "properties": {
        "scala.usejavacp": "true"
      }
    }
  ]
}
```
The above app descriptor works this way:
1. If no version is specified, Coursier will install the latest stable Scala 3 compiler:

```bash
> ./mill -i cli.run install scalac
https://repo1.maven.org/maven2/org/scala-lang/scala3-compiler_3/maven-metadata.xml
  No new update since 2021-10-20 06:09:06
Wrote scalac
> scalac -version
Scala compiler version 3.1.0 -- Copyright 2002-2021, LAMP/EPFL
```
2. However, if the user specifies a version below `2.max`, Coursier will install the Scala 2 compiler whose maven coordinates and
main class are different:

```bash
> ./mill -i cli.run install scalac:2.13.5
Wrote scalac
> scalac -version
Scala compiler version 2.13.5 -- Copyright 2002-2020, LAMP/EPFL and Lightbend, Inc.
```
3. Also, `cs update scalac` will always overwrite `scalac` with the latest stable Scala 3 compiler.

### DIFF
--- a/modules/install/src/main/scala/coursier/install/RawAppDescriptor.scala
+++ b/modules/install/src/main/scala/coursier/install/RawAppDescriptor.scala
@@ -3,8 +3,22 @@ package coursier.install
 import argonaut._
 import cats.data.{NonEmptyList, Validated, ValidatedNel}
 import cats.implicits._
-import coursier.core.{Classifier, Configuration, ModuleName, Resolution, Type}
-import coursier.parse.{DependencyParser, JavaOrScalaModule, ModuleParser, RepositoryParser}
+import coursier.core.{
+  Classifier,
+  Configuration,
+  ModuleName,
+  Resolution,
+  Type,
+  Repository,
+  VersionInterval
+}
+import coursier.parse.{
+  DependencyParser,
+  JavaOrScalaModule,
+  ModuleParser,
+  RepositoryParser,
+  JavaOrScalaDependency
+}
 import dataclass._
 
 import scala.language.implicitConversions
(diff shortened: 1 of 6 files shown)


adpi2:
@alexarchambault I've just rebased to fix the conflicts.

modules/install/src/main/scala/coursier/install/RawAppDescriptor.scala
```diff
@@ -216,6 +246,71 @@ object RawAppDescriptor {
...
+    versionOverrides
+      .map(_.versionRange)
+      .foldLeft[ValidatedNel[String, Seq[VersionInterval]]](Validated.valid(Seq.empty)) {
+        case (validRanges, range) =>
+          validRanges.andThen { ranges =>
+            val conflictingRanges = ranges.filter(_.merge(range).nonEmpty)
+
+            if (conflictingRanges.isEmpty) Validated.valid(ranges :+ range)
+            else {
+              val conflicts = conflictingRanges.map(i => "\"$i\"").mkString("[", ", ", "]")
```
dwijnand, line 307: This isn't interpolating.

alexarchambault: Should be fixed by https://github.com/coursier/coursier/pull/2513.

PR #1974, opened 2021-02-16
Don't clear User-Agent

alexarchambault:

PR #2246, opened 2021-10-29
Set a non-empty User-Agent

julienrf:
Some repositories have a policy that rejects requests with an empty User-Agent value.

This change sets the User-Agent to a valid non-empty value.

### DIFF
--- a/modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
+++ b/modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
@@ -142,7 +142,7 @@ object CacheUrl {
 
         // Early in the development of coursier, I ran into some repositories (Sonatype ones?) not
         // returning the same content for user agent "Java/…".
-        conn0.setRequestProperty("User-Agent", "")
+        conn0.setRequestProperty("User-Agent", "Coursier")
 
         conn0 match {
           case conn1: HttpsURLConnection =>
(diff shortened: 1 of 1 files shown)


julienrf:
Would it be possible to get a review?

modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
```diff
@@ -142,7 +142,7 @@ object CacheUrl {
 
         // Early in the development of coursier, I ran into some repositories (Sonatype ones?) not
         // returning the same content for user agent "Java/…".
-        conn0.setRequestProperty("User-Agent", "")
+        conn0.setRequestProperty("User-Agent", "Coursier")
```
alexarchambault, line 145: Let's just remove this line (and the comment above), the default user agent (something like `Java/…`) should be fine.

alexarchambault (commented):
Thanks. I had tried similar things in the past (https://github.com/coursier/coursier/pull/1974), which I didn't want to merge straightaway, as I wanted to run the tests once without all the POM files in the test fixtures (so that the tests would download everything again with the new user agent).

I just did that in https://github.com/coursier/coursier/pull/1974, right after updating the tests in https://github.com/coursier/coursier/pull/2249, and it went fine.

Let's just keep the default user agent, and we can merge.

modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
julienrf: Done.

alexarchambault (approved):

PR #2250, opened 2021-11-04
Add main and command descriptions in help message

adpi2:
The `cs -h` message looks like this:
```
Usage: coursier <COMMAND>
Coursier is the Scala application and artifact manager.
It can install Scala applications and setup your Scala development environment.
It can also download and cache artifacts from the web.



Commands:
  bootstrap                          Create a binary launcher from a dependency or an application descriptor.
  channel                            Manage additional channels, used by coursier to resolve application descriptors.
  complete-dep, complete-dependency  Auto-complete Maven coordinates.
  fetch                              Transitively fetch the JARs of one or more dependencies or an application.
  get                                Download and cache a file from its URL.
  install                            Install an application from its descriptor.
  java                               Manage installed JVMs and run java.
  java-home                          Print the home directory of a particular JVM.
  launch                             Launch an application from a dependency or an application descriptor.
  list                               List all currently installed application
  publish                            [Experimental] Publish an artifact to a maven repository.
  resolve                            Resolve and print the transitive dependencies of one or more dependencies or an application.
  search                             Search application names from known channels.
  setup                              Setup a machine for Scala development.
  uninstall                          Uninstall one or more applications.
  update                             Update one or more applications.
```

As an example the `cs bootstrap -h` message looks like this:
```
Usage: coursier bootstrap [options] [org:name:version*|app-name[:version]]
Create a binary launcher from a dependency or an application descriptor.
The generated launcher can then be used without cs being installed.

Examples:
$ cs bootstrap org.scalameta::scalafmt-cli:2.4.2 -o scalafmt
$ cs bootstrap scalafmt -o scalafmt

App channel options:
  --channel org:name  Channel for apps
  --contrib           Add contrib channel

<elided>
```

### DIFF
--- a/modules/cli/src/main/scala/coursier/cli/Coursier.scala
+++ b/modules/cli/src/main/scala/coursier/cli/Coursier.scala
@@ -21,6 +21,11 @@ object Coursier extends CommandsEntryPoint {
 
   lazy val progName = (new Argv0).get("coursier")
 
+  override val description =
+    """|Coursier is the Scala application and artifact manager.
+       |It can install Scala applications and setup your Scala development environment.
+       |It can also download and cache artifacts from the web.""".stripMargin
+
   val commands = Seq(
     bootstrap.Bootstrap,
     channel.Channel,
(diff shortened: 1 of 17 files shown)


alexarchambault (approved):
Thanks @adpi2!

PR #2254, opened 2021-11-08
Review the docs to highlight the beginner use case.

sjrd:
This complements recent changes in the `--help` menus.

The documentation used to highlight the artifact resolver use case of coursier first and foremost. We now highlight the beginner use case of setting up and installing applications instead.


alexarchambault (approved):
Thanks, LGTM!

PR #2269, opened 2021-11-19
Beginner-oriented command help messages

adpi2:
Comes after https://github.com/coursier/coursier/pull/2253
Fixes https://github.com/coursier/coursier/issues/2196

- Hide some more options
- Group all options
- Sort all option groups

PR #2272, opened 2021-11-23
Beginner oriented command help messages

sjrd:
- Hide more options
- Group all options
- Sort all option groups

---

This is an updated version of #2269 because @adpi2 is on vacation now.


alexarchambault (approved):
Thanks!

Issue #930, opened 2018-10-05
Add support for proxy authentication


alexarchambault:
AFAIK, proxy authentication isn't automatically handled via [Java properties](https://docs.oracle.com/javase/8/docs/technotes/guides/net/proxies.html). Some extra setup is needed, see e.g. [this question](https://stackoverflow.com/questions/1626549/authenticated-http-proxy-with-java) or https://github.com/almond-sh/almond/pull/182.

The CLI of coursier should allow one to set up proxy authentication, and helpers should be available for that for the API users.

ekrich:
I just added this to the sbt issue so I am copying here as requested by @eed3si9n 

I have some additional information on this issue related to Basic auth and the redirect need to download plugins - normal dependencies download ok.

```
[error]   download error: Caught java.io.IOException: Unable to tunnel through proxy. Proxy returns "HTTP/1.1 407 Proxy
Authentication Required" (Unable to tunnel through proxy. Proxy returns "HTTP/1.1 407 Proxy Authentication Required") while 
downloading https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.scala-native/sbt-scala-native/scala_2.12/sbt_1.0/0.4.0-
M2/ivys/ivy.xml
... (full error below)
```

I am able to access via `curl` with `-L` required for redirect:

```
$ curl -I -L -v  --proxy-user user:pass  https://repo.scala-sbt.org/scalasbt/sbt-plugin-releases/org.scala-native/sbt-scala-native/scala_2.12/sbt_1.0/0.4.0-M2/ivys/ivy.xml
```

I found this SOF: https://stackoverflow.com/questions/41505219/unable-to-tunnel-through-proxy-proxy-returns-http-1-1-407-via-https
Which points to this Oracle notice for jdk 8 and above about Basic auth excluded: https://www.oracle.com/java/technologies/javase/8u111-relnotes.html

I tried adding the following to `JAVA_OPTS` with no luck. I have the same output as the original issue plus these.

```
-Djdk.http.auth.tunneling.disabledSchemes=
-Djdk.http.auth.proxying.disabledSchemes=
```

Somehow I got this to work using `sbt` `1.4.0-M2` but not `1.3.13`. Now I can not repeat. I can get a normal Scala app to work but not a Scala Native that requires a plugin and thus a redirect via a proxy with Basic auth.

```
[warn]
[warn]  Note: Some unresolved dependencies have extra attributes.  Check that these dependencies exist with the requested attributes.
[...]

PR #2278, opened 2021-11-29
Mention about `proxyUser` and `proxyPassword` in docs

kpodsiad:
Proxy authentication support [was added](https://github.com/coursier/coursier/issues/930) a long time ago, but such capability wasn't mentioned in docs.


alexarchambault (commented):
Thanks @kpodsiad! I slightly changed the wording as a suggestion.

If you're ok with it, we can merge this PR right after.

kpodsiad:
Thanks for the suggestion @alexarchambault, let's merge it.

PR #2286, opened 2021-12-06
Case-insensitive OS activation

dwickern:
This profile does not activate due to the capital W in Windows:
```xml
<activation>
  <os>
    <family>Windows</family>
    <arch>amd64</arch>
  </os>
</activation>
```

For my use-case, jvmbrotli uses profiles to pull in its jni libs for each operating system:
https://github.com/nixxcode/jvm-brotli/blob/06a73890dec39b060c80d021a45e80c8ae1a8c32/jvmbrotli/pom.xml#L18-L33

This works for `<family>unix</family>` and `<family>mac</family>` but not `<family>Windows</family>` due to the capitalization:
https://github.com/dwickern/sbt-web-brotli/runs/4423716589?check_suite_focus=true

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Activation.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Activation.scala
@@ -75,16 +75,19 @@ object Activation {
     def isEmpty: Boolean =
       arch.isEmpty && families.isEmpty && name.isEmpty && version.isEmpty
 
+    def archMatch(current: Option[String]): Boolean =
+      archNormalized.forall(current.contains)
+
     def isActive(osInfo: Os): Boolean =
-      archNormalized.forall(osInfo.archNormalized.contains) &&
-        familiesNormalized.forall { f =>
-          if (Os.knownFamilies(f))
-            osInfo.familiesNormalized.contains(f)
-          else
-            osInfo.nameNormalized.exists(_.contains(f))
-        } &&
-        nameNormalized.forall(osInfo.nameNormalized.contains) &&
-        versionNormalized.forall(osInfo.versionNormalized.contains)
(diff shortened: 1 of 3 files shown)


alexarchambault (commented):
Thanks! It looks like this needs to be fixed, yes.

I think I'd prefer to keep the arch / family / … as is in the `Activation.Os` class, and only convert things to lower case in its `isActive` method (maybe keeping lower case values as `private lazy val`s in `Activation.Os`). Some people use the POM parsing capabilities of coursier, and I'd rather keep the original values for them, just in case.

dwickern:
I did as you suggested so that we preserve the original `Os` values. I switched to `Locale.US` for string comparisons, since that's what [plexus-utils](https://github.com/sonatype/plexus-utils/blob/f2beca21c75084986b49b3ab7b5f0f988021dcea/src/main/java/org/codehaus/plexus/util/Os.java#L72) uses.

modules/core/shared/src/main/scala/coursier/core/Activation.scala
```diff
@@ -60,25 +62,29 @@ object Activation {
...
+        case arch     => arch
+      }
+    private lazy val familiesNormalized = families.map(_.toLowerCase(Locale.US))
+    private lazy val nameNormalized     = name.map(_.toLowerCase(Locale.US))
+    private lazy val versionNormalized  = version.map(_.toLowerCase(Locale.US))
+
     def isEmpty: Boolean =
       arch.isEmpty && families.isEmpty && name.isEmpty && version.isEmpty
 
-    def archMatch(current: Option[String]): Boolean =
```
dwickern, line 66: Do we need to keep this for compatibility since it's public?

Issue #2287, opened 2021-12-06
Unable to download snapshot dependencies with 2.1.0-M1


hmemcpy:
I initially reported this as an [sbt issue](https://github.com/sbt/sbt/issues/6737) , as it seems that the latest upgrade cannot resolve snapshot dependencies from sonatype.

I managed to manually reproduce this with just coursier. I had an older v2.0.7 installed, and it fetches the snapshot dependency just fine:

```
> cs fetch --no-default -r central -r sonatype:snapshots org.zeromq:jeromq:0.5.3-SNAPSHOT
https://oss.sonatype.org/content/repositories/snapshots/org/zeromq/jeromq/0.5.3-SNAPSHOT/maven-metadata.xml
  100.0% [##########] 1.2 KiB (111B / s)
https://oss.sonatype.org/content/repositories/snapshots/org/zeromq/jeromq/0.5.3-SNAPSHOT/jeromq-0.5.3-20211128.145727-50.pom
  100.0% [##########] 16.7 KiB (42.8 KiB / s)
https://oss.sonatype.org/content/repositories/snapshots/org/zeromq/jeromq/0.5.3-SNAPSHOT/jeromq-0.5.3-20211128.145727-50.jar
  100.0% [##########] 506.2 KiB (467.4 KiB / s)
/Users/hmemcpy/Library/Caches/Coursier/v1/https/oss.sonatype.org/content/repositories/snapshots/org/zeromq/jeromq/0.5.3-SNAPSHOT/jeromq-0.5.3-20211128.145727-50.jar
/Users/hmemcpy/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/eu/neilalexander/jnacl/1.0.0/jnacl-1.0.0.jar
```

However, upgrading to the latest coursier 2.1.0-M1-1 fails to resolve this:

```
> ll $(which cs)
lrwxr-xr-x  1 hmemcpy  admin    36B Dec  6 10:08 /usr/local/bin/cs -> ../Cellar/coursier/2.1.0-M1-1/bin/cs
> cs fetch --no-default -r central -r sonatype:snapshots org.zeromq:jeromq:0.5.3-SNAPSHOT
https://oss.sonatype.org/content/repositories/snapshots/org/zeromq/jeromq/0.5.3-SNAPSHOT/maven-metadata.xml
  100.0% [##########] 353B (680B / s)
[...]

PR #2294, opened 2021-12-12
Back to empty user agent

alexarchambault:
Fixes https://github.com/coursier/coursier/issues/2287.

### DIFF
--- a/modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
+++ b/modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
@@ -140,6 +140,10 @@ object CacheUrl {
         // handling those ourselves, so that we can update credentials upon redirection
         conn0.setInstanceFollowRedirects(false)
 
+        // Early in the development of coursier, I ran into some repositories (Sonatype ones?) not
+        // returning the same content for user agent "Java/…".
+        conn0.setRequestProperty("User-Agent", "")
+
         conn0 match {
           case conn1: HttpsURLConnection =>
             for (f <- sslSocketFactoryOpt)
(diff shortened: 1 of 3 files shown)


modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
```diff
@@ -140,6 +140,10 @@ object CacheUrl {
         // handling those ourselves, so that we can update credentials upon redirection
         conn0.setInstanceFollowRedirects(false)
 
+        // Early in the development of coursier, I ran into some repositories (Sonatype ones?) not
+        // returning the same content for user agent "Java/…".
+        conn0.setRequestProperty("User-Agent", "")
```
julienrf, line 145: Can we make

```suggestion
        conn0.setRequestProperty("User-Agent", "Coursier")
```

As initially proposed in #2246?

eed3si9n (approved):
Thanks 🙏

modules/cache/jvm/src/main/scala/coursier/cache/CacheUrl.scala
eed3si9n: curl test says `"Coursier"` would work too

```bash
$ curl -H "User-Agent: Coursier" https://oss.sonatype.org/content/repositories/snapshots/org/zeromq/jeromq/0.5.3-SNAPSHOT/maven-metadata.xml
```

so I am happy with either.

alexarchambault: I think I'm going to cut a release with an empty user agent (former behavior), then one with a `Coursier` user agent.

PR #2295, opened 2021-12-12
Set user agent to 'Coursier/2.0'

alexarchambault:
I guess the version should stay at '2.0' for the time being - not to leak the full coursier version, or details about the OS, all around.


julienrf:
Related: #2246

PR #2308, opened 2021-12-21
docs: add curl download instructions for Windows

japm48:
cURL has been included in Windows 10/11 since build 17063 ([source](https://docs.microsoft.com/en-us/virtualization/community/team-blog/2017/20171219-tar-and-curl-come-to-windows)).

The BITS commands are sometimes problematic (slow and unreliable), so this alternative could be useful.


alexarchambault (commented):
Thanks! I actually just [changed these instructions](https://github.com/coursier/coursier/pull/2339), as the launchers are now compressed… I was aware of curl being more widely available now, so I used it when updating the instructions.

So I only kept one line from your changes when resolving conflicts :|

But the new instructions are probably perfectible, so feel free to update them further if you want to…

japm48:
Sorry for the late reply. I think that's fine as it is now, thanks!

PR #2332, opened 2022-01-19
Include pre-release versions in versions from listing

fthomas:
This includes pre-release versions in `MavenRepository#versionsFromListing`. I've no idea why these versions are currently ignored and also haven't found any clue about that in the Git history.

Motivation for this change is https://github.com/typelevel/sbt-typelevel/issues/78. @armanbilge and I found that Scala Steward (which uses Coursier under the hood) does not create updates to pre-release versions of sbt plugins hosted on Maven Central. The reason is that there is no `maven-metadata.xml` for sbt plugins so that Coursier reads the versions from the directory listing but ignores pre-release versions then.

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/maven/MavenRepository.scala
+++ b/modules/core/shared/src/main/scala/coursier/maven/MavenRepository.scala
@@ -245,17 +245,13 @@ object MavenRepository {
             case _              => true
           })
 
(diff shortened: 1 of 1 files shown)


alexarchambault (approved):
Thanks @fthomas! I'm not sure why it was this way, but your changes look legit!

PR #2367, opened 2022-02-09
Set HTTP Accept header

ybasket:
Works around problems with remote repositories like AWS CodeArtifact that return 404 on `maven-metadata.xml` if requested with JVM's default `Accept` HTTP header (which is `text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2` – see [here](https://github.com/openjdk-mirror/jdk/blob/jdk8u/jdk8u/master/src/share/classes/sun/net/www/protocol/http/HttpURLConnection.java#L256)).


ybasket:
The one CI job seems to have failed for no good reason, it's the GitHub cache action that got stuck downloading 🤷

alexarchambault (approved):
Thanks!

PR #2374, opened 2022-02-22
Update launcher.bat

jesseharte:
I needed to make this change locally in sbt.bat after installing it with Coursier in order for my sbt commands to work. Let me know if this is a correct and complete fix.


alexarchambault (approved):
Thanks!

alexarchambault:
> Let me know if this is a correct and complete fix.

It seems correct. `@COMMAND@` is supposed to be substituted by a path to another command, typically.

Issue #2118, opened 2021-07-18
Coursier installs applications under null on Windows


rgwilton:
Depending on how Coursier is run on windows it gets the wrong null path when attempting to install anything.

Running the .exe directly, or under Powershell, or under cmd, all seem to pick up a bad null path:

Checking if ~\Downloads\null\Coursier\data\bin is in PATH
  Should we add ~\Downloads\null\Coursier\data\bin to your PATH? [Y/n]

When I ran it under gitbash then I did manage to install into the right place (under $home/AppData/local/Cousier), but the application links don't work so well because you have to run cs.bat rather than cs.  And even then, if I get Coursier to install Java then it ends up the wrong place.

E.g., 
export JAVA_HOME="C:\Users\rober\null\Coursier\cache\jvm\adopt@1.11.0-11"
export PATH="C:\Users\rober\null\Coursier\cache\jvm\adopt@1.11.0-11\bin;$PATH"

This is using Cousier 2.0.13, but all the recent versions seem to have this problem.

som-snytt:
https://github.com/coursier/coursier/issues/2031

ratbeknurlanov:
Does anybody know why that occurs and how to fix?

rgwilton:
> Does anybody know why that occurs and how to fix?

If you install git for windows (https://gitforwindows.org/) and then run the commands under "Git Bash" (that is installed as part of git for windows) then that seemed to be a workaround for me.

s5bug:
You can work around this without Git Bash by setting `COURSIER_CACHE` and `COURSIER_JVM_CACHE` manually. I have them set to `F:\Win\Program Files\`, but you may want to set them to the default locations according to the documentation. https://get-coursier.io/docs/cache

![image](https://user-images.githubusercontent.com/7751296/128920893-80affa1d-9df4-49b8-9b65-1002c97dae6e.png)

zstcode:
I also found this problem. Using the JAR-based launcher instead of the native one seemed to be a workaround for me.

Issue #2031, opened 2021-04-25
'null' directories on Windows 10


mykyar:
When running cs.exe install ammonite it creates null directory.  With all the stuff Coursier/data.. cache etc. on the running level.

version of Coursier - 2.0.13

ShapelessCat:
Current Coursier document mentions "Note that this must be run with cmd.exe, not PowerShell." in this page https://get-coursier.io/docs/cli-installation.

I also see this issue #1984, and it seems this is an upstream issue. This #1984 has already been closed, and it seems use Coursier through sbt now works (I didn't try). However, I can see this issue is still there when using the latest version of Coursier if we use powershell.  Use Coursier through cmd.exe works well.

scalway:
It still creates null folder in both cmd and powershell :(.

PR #2389, opened 2022-03-12
Add PowerShell and fix CMD install instructions

s5bug:
My only worry with this is #2031 and #2118. I have not tested these instructions on a machine lacking `COURSIER_CACHE`, `COURSIER_BIN_DIR`, and `COURSIER_JVM_CACHE`.

[My own instructions for installing Coursier](https://b.s5.pm/os/2021/08/28/windows-setup.html#java--scala) specify that those environment variables are needed, however I don't know what has changed since that was written.


alexarchambault (approved):
Thanks!

PR #2394, opened 2022-04-01
Print the installed location in `coursier java --available` and restore `--installed` option

### LINKED ISSUE #2300: Unrecognized option: --installed
Hello!

This might just be a problem with my setup, but I can't figure out why `cs java --installed` fails. Any idea?

```
$ brew info coursier
coursier/formulas/coursier: stable 2.1.0-M2-1
Launcher for Coursier
https://get-coursier.io
/usr/local/Cellar/coursier/2.1.0-M2-1 (5 files, 60.4MB) *
  Built from source on 2021-12-15 at 17:56:45
From: https://github.com/coursier/homebrew-formulas/blob/HEAD/coursier.rb
==> Dependencies
Required: openjdk ✔
...
$ cs java --version
openjdk 17 2021-09-14
OpenJDK Runtime Environment (build 17+35-2724)
OpenJDK 64-Bit Server VM (build 17+35-2724, mixed mode, sharing)
$ cs java --available
adopt:1.8.0-181
adopt:1.8.0-192
adopt:1.8.0-202
adopt:1.8.0-212
...
$ cs java --installed
Unrecognized option: --installed
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

jneira:
Hi, trying coursier lately i realized `coursier jvm --installed` mentioned in ~~not too old~~[master](https://github.com/coursier/coursier/blob/master/doc/docs/cli-java.md#managed-jvms) docs was not working anymore:

```
D:\ws>coursier java --installed
Unrecognized option: --installed
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

so i started to work in add it. Later i realized the feature was removed on purpose and commented in the pr which did it: https://github.com/coursier/coursier/pull/2188#issuecomment-1079935827

Well as i dont know what were the reasons to remove it i am opening the pr knowing it could be ignored/discarded.

However, printing the installed location in `--available` seems pretty useful to me too and it could be added independently. `--installed` will be matter of doing `coursier jvm --installed | grep installed`

Maybe it would close #2300

Thanks in advance!

### DIFF
--- a/modules/cli/src/main/scala/coursier/cli/jvm/Java.scala
+++ b/modules/cli/src/main/scala/coursier/cli/jvm/Java.scala
@@ -48,7 +48,7 @@ object Java extends CoursierCommand[JavaOptions] {
       params.output.verbosity
     )
 
-    if (params.available) {
+    if (params.available || params.installed) {
       val task =
         for {
           index <- jvmCache.index.getOrElse(sys.error("should not happen"))
@@ -58,8 +58,13 @@ object Java extends CoursierCommand[JavaOptions] {
                 (name, versionMap) <- map.toVector.sortBy(_._1)
                 version <- versionMap.keysIterator.toVector.map(Version(_)).sorted.map(_.repr)
               } yield s"$name:$version"
-              for (id <- available)
-                System.out.println(id)
+
+              for (id <- available) {
+                val file = jvmCache.getIfInstalled(id).unsafeRun()(coursierCache.ec)
+                val file0 = file.map( f => s"installed at $f").getOrElse("")
+                if (params.available || (params.installed && file.isDefined))
+                  System.out.println(s"$id $file0")
+              }
             }
           }
         } yield maybeError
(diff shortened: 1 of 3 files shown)


modules/cli/src/main/scala/coursier/cli/jvm/Java.scala
```diff
@@ -58,8 +58,13 @@ object Java extends CoursierCommand[JavaOptions] {
                 (name, versionMap) <- map.toVector.sortBy(_._1)
                 version <- versionMap.keysIterator.toVector.map(Version(_)).sorted.map(_.repr)
               } yield s"$name:$version"
-              for (id <- available)
-                System.out.println(id)
+
+              for (id <- available) {
+                val file = jvmCache.getIfInstalled(id).unsafeRun()(coursierCache.ec)
```
jneira, line 63: I guess this is not the right way to compose tasks. But the unique way i can think off is using `sequence` and i am not sure if we have it available

alexarchambault: Yes, no problem, I pushed an extra commit fixing that.

alexarchambault (approved):
Thanks @jneira!

modules/cli/src/main/scala/coursier/cli/jvm/Java.scala
jneira: thanks for continuing the pr, a good opportunity for me to learn the api 🙂

PR #2415, opened 2022-05-12
Add scala-cli to default app list

alexarchambault:


julienrf:
Assuming `scala-cli` will supersede `scala` in the near future, what would the transition look like? Would it be possible to have a `scala-cli` app that would print a warning “'scala-cli' has been renamed to 'scala'. The command name 'scala-cli' will be removed in 2024”?

sjrd:
This PR should also update the list of applications installed in the documentation, at
https://github.com/coursier/coursier/blob/7b7c2c312aea26e850f0cd2cf15e688d0777f819/doc/docs/cli-installation.md?plain=1#L8-L13

julienrf:
It would be good to merge the page [Installation](https://get-coursier.io/docs/cli-installation) with the page [setup](https://get-coursier.io/docs/cli-setup) because they both document the setup process. This was an issue for me because when I looked at the list of applications installed by `cs setup` I couldn’t find the answer in the `setup` documentation page, and I didn’t think it would be in the Installation page. Currently, some information about the installation process is available only in the Installation page, but some other information is available only in the `setup` page, and some information is also redundant between both pages.

alexarchambault:
> Assuming `scala-cli` will supersede `scala` in the near future, what would the transition look like? Would it be possible to have a `scala-cli` app that would print a warning “'scala-cli' has been renamed to 'scala'. The command name 'scala-cli' will be removed in 2024”?

I'm not sure. We'd need to allow app descriptors to be flagged as deprecated, and add support here to read that and warn users.

alexarchambault:
> It would be good to merge the page [Installation](https://get-coursier.io/docs/cli-installation) with the page [setup](https://get-coursier.io/docs/cli-setup) because they both document the setup process.

This, or maybe they can be re-balanced or have more cross-links.

julienrf:
> I'm not sure. We'd need to allow app descriptors to be flagged as deprecated, and add support here to read that and warn users.

I don’t think we need to add a deprecation flag to app descriptors themselves. Would it be possible instead to point the scala-cli app descriptor to an app that would display a deprecation warning at startup?

> This, or maybe they can be re-balanced or have more cross-links.

I think both pages document the same thing. What would you put in cross-links messages? “This page contains only a part of the documentation, please also read _that page_ to get the full picture” sounds like a hint that both pages should be unified.

alexarchambault:
> I don’t think we need to add a deprecation flag to app descriptors themselves. Would it be possible instead to point the scala-cli app descriptor to an app that would display a deprecation warning at startup?

Deprecation for app descriptors has been needed already, with https://github.com/coursier/apps/pull/152 being the latest example of it. It'll probably be added anyway.

> I think both pages document the same thing. What would you put in cross-links messages? “This page contains only a part of the documentation, please also read that page to get the full picture” sounds like a hint that both pages should be unified.

I don't mind either way. If you want to update those pages, feel free to open a PR to do so.

julienrf:
See #2417

PR #2420, opened 2022-05-22
Improve performance for DependencySet.add

craffit:
Analysis with jprofiler showed ~1/5th of time spent in DependencySet.add for an `sbt update` i was running (see profile picture).

Screenshot from 2022-05-22 19-34-12

This is caused by DependencySet.addNoCheck completely rebuilding the 'grouped' HashMap from scratch. Using immutable collections in this PR brought `sbt update` to 111s from 131s prior for my local workload


alexarchambault (approved):
Thanks @craffit!

I didn't run benchmarks or checked the performance of this locally. But this doesn't break tests and looks fine, so that's good.

PR #2441, opened 2022-05-25
fix: make sure JAVA_HOME is used in bootstrapped file

### LINKED ISSUE #2048: JAVA_HOME variable ignored by bash script
***Preliminary note** : [Issue 2519](https://github.com/scalameta/scalafmt/issues/2519) originally submitted to [`scalafmt`](https://github.com/scalameta/scalafmt) project.*
***
## Problem

On MS Windows the Coursier managed batch file
``` 
%LOCALAPPDATA%\Coursier\data\bin\scalafmt.bat
```
contains the following code which looks for `java.exe` *both* through `JAVA_HOME` and `PATH`:
```shell
16: if not "%JAVA_HOME%" == "" goto OkJHome
17:
18: for /f %%j in ("java.exe") do (
19:   set JAVA_EXE="%%~$PATH:j"
20:   goto init
21: )
22:
23: :OkJHome
24: if exist "%JAVA_HOME%\bin\java.exe" (
25:  SET JAVA_EXE="%JAVA_HOME%\bin\java.exe"
26:  goto init
27: )
```
On Linux the Coursier managed bash script
```
$HOME/.local/share/coursier/bin/scalafmt
```
*only* looks for a Java executable in `PATH`.

## Expectation

Both scripts should use `JAVA_HOME` if defined.

## Workaround

In the bash script change the following line
```
27: exec java $JAVA_OPTS "$@"
```
to something like :
```bash
27: [ -x "$JAVA_HOME/bin/java" ] && JAVACMD="$JAVA_HOME/bin/java" || JAVACMD=java
28: exec $JAVACMD $JAVA_OPTS "$@"
```

ckipp01:
I came across this for some Metals users that use coursier to bootstrap
a metals executable. The executable just calls `java` and doesn't
respect `JAVA_HOME`. This change makes sure that in the shell script if
`JAVA_HOME` is set, it's used and if not we fall back to `java`.

Closes #2048

### DIFF
--- a/modules/launcher/src/main/scala/coursier/launcher/Preamble.scala
+++ b/modules/launcher/src/main/scala/coursier/launcher/Preamble.scala
@@ -41,9 +41,13 @@ import scala.io.{Codec, Source}
       .sorted
       // escaping possibly a bit loose :-|
       .map { case (k, v) => s"""export $k="$v"""" }
+
     val lines = command match {
       case None =>
-        val javaCmd = Seq("java") ++
+        val setJavaCmd =
+          """[ -x "$JAVA_HOME/bin/java" ] && JAVA_CMD="$JAVA_HOME/bin/java" || JAVA_CMD=java"""
+
+        val javaCmd = Seq("$JAVA_CMD") ++
           // escaping possibly a bit loose :-|
           javaOpts.map(s => "'" + s.replace("'", "\\'") + "'") ++
           jvmOptionFile.toSeq.map(_ => "${extra_jvm_opts[@]}") ++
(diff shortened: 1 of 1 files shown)


alexarchambault (commented):
Thanks @ckipp01!

That looks great, except for the `modules/test/metadata` submodule change (that submodule shouldn't be touched by this PR, I think)

ckipp01:
> That looks great, except for the modules/test/metadata submodule change (that submodule shouldn't be touched by this PR, I think)

Ah weird, sorry I didn't even notice that! I'll get rid of it.

alexarchambault (approved):
Thanks! I'll merge as soon as the CI is green.

PR #2442, opened 2022-05-30
Faster hashing for exclusions

craffit:
Jprofiler showed (for my local workload) ~50% of coursiers time being spent during getting a hashCode for Dependencies, for which the main contributor is the set of exclusions in the dependency:

Screenshot from 2022-05-30 09-02-54
 
This PR contains:
- An explicit Exclusions data structure which:
  - Keeps the hascode cached
  - Keeps the exclusions structure always minimized
  - Special cases for ExcludeAll/ExcludeNone
- Optimized hashcode for the `Publication` data structure (using a `lazy val` instead of `def`)

This change brought down the `sbt update` invocation on my local machine from 30s to 20s

Caveat (backwards compatibility):
I have tried to keep these changes backwards (binary) compatible, but it is not the case, the following spots are tricky:
- Dependency.exclusions is now a completely different type, not even being a Set, so that is probably breaking
- I added a backwards-compatible `Dependency.withExclusions(exclusions: Set[(Organization, ModuleName)])` function call, but because this now became an overloaded function, every call to `withExclusions(Set.empty)` becomes ambiguous, because the set type arguments cannot be inferred anymore.

@alexarchambault Could you let me know how important binary backwards compatibility is for this library? And for what parts? If it is an issue, I can see what else I can do to avoid breaking binary backwards compatibility


alexarchambault:
> Could you let me know how important binary backwards compatibility is for this library? And for what parts? If it is an issue, I can see what else I can do to avoid breaking binary backwards compatibility

It's fine to break it. I intend to break it even more before `2.1.0`, as time is missing to maintain it or to keep a proper migration path with deprecations.

craffit:
I already made it binary backwards compatible, it was not too much of a fuss. If you want me to revert that and make the code more straightforward without the compatibility hooks and deprecation, i'll do that no problem

craffit:
@alexarchambault I've updated the PR, the backwards compatibility is not an issue anymore. From my side I think this work is done. I've added you as a reviewer. Could you take a look to see whether/what you would like to see changed?

alexarchambault (approved):
Thanks a lot, @craffit!

(I didn't review in minute details, but given that this doesn't break any tests, I'd say the probability that this is incorrect is very very low.)

lefou:
This PR is a compatibility breaker. You can find details in this issue: #2560

PR #2480, opened 2022-07-16
Update cli-installation.md

paxel:
If you only call cs --help scala is not installed and in the later steps confusion is big


paxel:
ah it should probably be `cs setup`

paxel:
my bad

paxel:
coursera linked to the page and the confusion was more by me. nothing wrong here

PR #2489, opened 2022-07-27
Support `--architecture` flag for Java commands

olafurpg:
Previously, it was not possible to use `coursier java-home` to download
an x86_64 architecture JDK on an Apple M1 computer. This may be a
niche use-case, but we can't use any existing aarch64 JDK 8
distributions in the https://github.com/sourcegraph/scip-java build
because none of them give access to the `com.sun` APIs as far as I can
tell. With this commit, we can use `--architecture amd64 --jvm 8` to
populate the boot classpath of our javac compiler plugin so that
`com.sun` APIs are available during compile time.

### DIFF
--- a/modules/cli/src/main/scala/coursier/cli/jvm/SharedJavaOptions.scala
+++ b/modules/cli/src/main/scala/coursier/cli/jvm/SharedJavaOptions.scala
@@ -18,6 +18,9 @@ final case class SharedJavaOptions(
     update: Boolean = false,
 
   @Group(OptionGroup.java)
-    jvmIndex: Option[String] = None
+    jvmIndex: Option[String] = None,
+
+  @Group(OptionGroup.java)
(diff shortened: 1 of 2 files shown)


alexarchambault (approved):
Looks good, thanks @olafurpg!

I guess this is meant to be used on M1, right? Feel free to report how it goes, and fix any other thing that needs fixing.

Also, about M1, there are now `cs` binaries built specifically for it [here](https://github.com/VirtusLab/coursier-m1/releases), that are supposed to be much faster (but don't address flaws in the handling of the architecture internally in coursier…)

olafurpg:
Can confirm that I was able to use `--architecture amd64` on an M1 chip to fetch a JVM where the default `./coursier java-home --jvm 8` was failing. I opened an followup issue https://github.com/coursier/coursier/issues/2510

PR #2538, opened 2022-09-19
Use `3` for Scala 3 Binary Version in Complete

LaurenceWarne:
Hi, I saw that that `3.x` is used by `Complete` when setting the binary version based on the full version for (non `-M`, `-RC`) Scala 3 versions, this PR changes it so `3` is always used to be in line with https://docs.scala-lang.org/scala3/reference/language-versions/binary-compatibility.html (and so completion results are a lot better when just using `withScalaVersion` :slightly_smiling_face:).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,7 @@
 [submodule "tests/metadata"]
 	path = modules/tests/metadata
-	url = https://github.com/coursier/test-metadata.git
+	url = https://github.com/LaurenceWarne/test-metadata.git
+	branch = update-for-scala3-tests
 [submodule "modules/directories"]
 	path = modules/directories
 	url = https://github.com/coursier/directories-jvm.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,6 @@
 [submodule "tests/metadata"]
 	path = modules/tests/metadata
-	url = https://github.com/LaurenceWarne/test-metadata.git
-	branch = update-for-scala3-tests
+	url = https://github.com/coursier/test-metadata.git
 [submodule "modules/directories"]
 	path = modules/directories
 	url = https://github.com/coursier/directories-jvm.git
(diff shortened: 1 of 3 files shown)


alexarchambault (commented):
Thanks @LaurenceWarne, this looks good!

It would have been nice to add non regression tests for this in [`CompleteTests`](https://github.com/coursier/coursier/blob/1df62c141ba696a403ca34cbfbe871a318db639a/modules/coursier/shared/src/test/scala/coursier/complete/CompleteTests.scala). Adding tests to it requires updating some metadata under `modules/metadata` though. This isn't properly documented IIRC. The idea is to add tests, run them while ensuring [this variable](https://github.com/coursier/coursier/blob/1df62c141ba696a403ca34cbfbe871a318db639a/modules/coursier/jvm/src/test/scala/coursier/PlatformTestHelpers.scala#L38-L39) is true, then pushing the additional metadata of the `modules/metadata` git submodule to [this repository](https://github.com/coursier/test-metadata).

LaurenceWarne:
Cool, I'll take a look!

LaurenceWarne:
Is https://github.com/coursier/coursier/pull/2538/commits/de7192291d0eeed7b2dfe4c2336cd65be59274de along the lines of what you had in mind @alexarchambault :slightly_smiling_face: ?

.gitmodules
```diff
@@ -1,6 +1,7 @@
 [submodule "tests/metadata"]
 	path = modules/tests/metadata
-	url = https://github.com/coursier/test-metadata.git
+	url = https://github.com/LaurenceWarne/test-metadata.git
+	branch = update-for-scala3-tests
```
LaurenceWarne, line 4: Temporary until https://github.com/coursier/test-metadata/pull/6, hopefully tests should go green with it

alexarchambault:
Merging, thanks!

PR #2540, opened 2022-09-21
install bash completion too

darklore:
- Installs bash completion script in addition to zsh completion when installed by homebrew.
- Change option name to avoid the installation from "--without-zsh-completions" to "--without-shell-completions"


alexarchambault (approved):
Thanks! I'm not sure how to try that locally, but I guess that should work (and that can be addressed or reverted swiftly if needed…).

Issue #2487, opened 2022-07-26
Coursier doesn't show latest Graalvm-java11 version (22)


carlosedp:
As Jabba index is dead, I reinstalled my whole Coursier to install the latest GraalVM-java11 (22) from the new index.

Weirdly the command `cs java --available |grep graalvm-java11` doesn't show the last 22.x versions of graalvm-java11 on MacOS amd64 but both the index and the local index cache show them:

```
❯ curl -sL https://github.com/coursier/jvm-index/raw/master/index.json |jq '.darwin.amd64."jdk@graalvm-java11"'
{
  "19.3.0": "tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-19.3.0/graalvm-ce-java11-darwin-amd64-19.3.0.tar.gz",
  "19.3.0.2": "tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-19.3.0.2/graalvm-ce-java11-darwin-amd64-19.3.0.2.tar.gz",
  "19.3.1": "tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-19.3.1/graalvm-ce-java11-darwin-amd64-19.3.1.tar.gz",
  "19.3.2": "tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-19.3.2/graalvm-ce-java11-darwin-amd64-19.3.2.tar.gz",
  "19.3.3": "tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-19.3.3/graalvm-ce-java11-darwin-amd64-19.3.3.tar.gz",
  "19.3.4": "tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-19.3.4/graalvm-ce-java11-darwin-amd64-19.3.4.tar.gz",
  "19.3.5": "tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-19.3.5/graalvm-ce-java11-darwin-amd64-19.3.5.tar.gz",
  "19.3.6": "tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-19.3.6/graalvm-ce-java11-darwin-amd64-19.3.6.tar.gz",
  "20.0.0": "tgz+https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-20.0.0/graalvm-ce-java11-darwin-amd64-20.0.0.tar.gz",
[...]

PR #2544, opened 2022-10-09
drop `jabba` and `cs-maven` from pre-installed JVM indexes

danielleontiev:
Both have not been updated for a while therefore there is no need to keep extra code to support them. Moreover, dropping `jabba` fixes https://github.com/coursier/coursier/issues/2487 because downloaded index was modified by `JvmIndex.jabbaIndexGraalvmJava8Hack` function with no respect to index source - so not only `jabba` index was modified by it. For example, for coursier index
https://github.com/coursier/jvm-index it messed up graalvm versions badly and caused the lost of latest graalvm versions.

### DIFF
--- a/doc/docs/cli-java.md
+++ b/doc/docs/cli-java.md
@@ -68,9 +68,8 @@ openjdk:1.14.0
 
 ## JVM index
 
-Currently, the `java` and `java-home` commands rely on
-[the index](https://github.com/shyiko/jabba/blob/master/index.json)
-from the command-line tool [jabba](https://github.com/shyiko/jabba).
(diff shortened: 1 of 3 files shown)


danielleontiev:
https://scastie.scala-lang.org/FefGHXgUROaaX3HOBLyOsg

alexarchambault (commented):
Thanks @danielleontiev!

Let's keep `cs-maven` just in case, I should rather update it more often.

danielleontiev:
superceded by https://github.com/coursier/coursier/pull/2547

PR #2548, opened 2022-10-14
chore(dep): use scala-xml 2.x for everything

ckipp01:
This is a continuation or really a revert of what was done in #2274. Originally this was done because Scala 2.12 was on 1.x, but that is no longer the case. You can see the discussion for this in https://github.com/scala/bug/issues/12632. I'm assuming that other things like sbt that will use this will also hit on it as well if it's still on 1.x. See https://github.com/sbt/sbt/issues/6997 for more details on that. This also then bumps Scala 2.12 to 2.12.17 to make sure the scala-xml is aligned there too.


alexarchambault (approved):
Thanks @ckipp01! I should merge as soon as the CI is green.

PR #2576, opened 2022-11-09
CVE-2022-37866 prevent path-traversal with bogus module coordinates

adpi2:
[CVE-2022-37866](https://github.com/advisories/GHSA-wv7w-rj2x-556x)


eed3si9n (approved):
Thanks, @adpi2!

PR #2597, opened 2022-11-24
Support URI userinfo without password part

### LINKED ISSUE #2596: Repository parser doesn't support userinfo without a password
Hi!

I'm using `RepositoryParser` on the JVM and I've noticed that it doesn't seem to work with certain URLs. For example, I have an AWS CodeArtifactURL that looks like this:

```
https://aws@<redacted>.d.codeartifact.us-east-1.amazonaws.com/maven/<redacted>
```

now, there are several ways I could parse that:

- compile-time literal (not an option for me): this just checks for `java.net.Uri` validness, and runs `MavenRepository.apply`
- `MavenRepository.apply` - I'd rather not, unless I'm guaranteed to have a Maven repo as input
- `RepositoryParser.repository` - that's what I've been using... but it fails for this exact URL.

The error I'm getting from `RepositoryParser` is a `Left(s"No password found in user info of URL $url")` which seems to come from this line:

https://github.com/coursier/coursier/blob/9b115ece4ca3c221e2c8de3c0b1922108fdd1de3/modules/coursier/jvm/src/main/scala/coursier/internal/PlatformRepositoryParser.scala#L93

However, using the `aws@...` string without a password as a `MavenRepository` is enough to fetch artifacts... as [the spec considers passwords optional](https://www.rfc-editor.org/rfc/rfc3986#section-3.2.1).

I think `RepositoryParser` should support userinfos without a password (happy to make that change).

kubukoz:
Closes #2596.

Any way we could get this into a 2.0.x patch release?

### DIFF
--- a/modules/coursier/jvm/src/main/scala/coursier/internal/PlatformRepositoryParser.scala
+++ b/modules/coursier/jvm/src/main/scala/coursier/internal/PlatformRepositoryParser.scala
@@ -61,36 +61,34 @@ abstract class PlatformRepositoryParser {
           case None =>
             repo
           case Some(userInfo) =>
-            userInfo.split(":", 2) match {
-              case Array(user, password) =>
-                val baseUrl = new java.net.URL(
-                  url.getProtocol,
-                  url.getHost,
-                  url.getPort,
-                  url.getFile
-                ).toString
+            val authBase = userInfo.split(":", 2) match {
+              case Array(user, password) => Authentication(user, password)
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
Looks good! From what I checked specifying only user is ok.

@alexarchambault is there anything I might be missing?

alexarchambault (approved):
Thanks @kubukoz! (and @tgodzik for the review / heads up)

PR #2605, opened 2022-12-02
Fix Metals import instructions

adpi2:
For some reason, if I don't run `clean` before `mill.contrib.bloop.Bloop/install`, Mill only generates five Bloop configuration files, instead of 77. @julienrf was also able to reproduce this.

For now I think it is safer to ask people to always run `./mill clean` before `./mill mill.contrib.bloop.Bloop/install`

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -45,6 +45,7 @@ $ ./mill -w __.compile
 Using the default import functionality of Metals is _not_ recommended. Instead,
 generate bloop configuration files with
 ```text
+$ ./mill clean
 $ ./mill mill.contrib.bloop.Bloop/install
 ```
 


tgodzik (approved):
LGTM! I have managed to import via metals without issues previously, but not sure if I needed the clean.

lefou:
> For some reason, if I don't run `clean` before `mill.contrib.bloop.Bloop/install`, Mill only generates five Bloop configuration files, instead of 77. @julienrf was also able to reproduce this.
> 
> For now I think it is safer to ask people to always run `./mill clean` before `./mill mill.contrib.bloop.Bloop/install`

Hi, I'm interested in a reproduction of this. As maintainer of Mill, I'm curious, whether this is some issue in Mill or specific to `mill.contrib.bloop`.

adpi2:
@lefou thanks for jumping in.

This should reproduce the bug in coursier/coursier:
1. Run `./mill clean`
2. Run `./mill mill.contrib.bloop.Bloop/install` => it creates 77 Json file in `.bloop`
3. Run `rm -Rf .bloop`
4. Run `./mill mill.contrib.bloop.Bloop/install` => it does not create any Json file

lefou:
This looks like an design problem of the Bloop plugin. The `<module>.bloop.writeConfig` is designed as target, which is cached, but it writes a file outside of `T.dest`. Instead, it should be a command, which is not cached.

Issue #2624, opened 2022-12-14
coursier Linux launcher broken


randov:
```
bash-4.2# curl -fL https://github.com/coursier/launchers/raw/master/cs-x86_64-pc-linux.gz | gzip -d > cs
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 20.1M  100 20.1M    0     0  9511k      0  0:00:02  0:00:02 --:--:-- 37.0M
bash-4.2# chmod +x ./cs
bash-4.2# ./cs setup
./cs: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by ./cs)
./cs: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by ./cs)
```

The linux used:
```
bash-4.2# cat /etc/os-release 
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"
```

Previous version of the launcher was working on the same Linux container.

tgodzik:
Thanks for reporting! That looks like an issue with updated ubuntu image. We run the release on the latest. It's possible that if you update your system the error will go away.

In the meantime I'll update the jobs.

alexarchambault:
It seems we should do just like how Scala CLI does: it generates its Linux binaries from a docker container. mill-native-image should allow to do that easily.

tgodzik:
> It seems we should do just like how Scala CLI does: it generates its Linux binaries from a docker container. mill-native-image should allow to do that easily.

Those are already images, we can just pin it for now. Change like https://github.com/coursier/coursier/pull/2625 worked for bloop

andriygm:
Issue still seems to be present on my Debian 11 machine. Using v2.1.0-RC2 seems to work for now.

miamagana:
Issue still present when trying to build it using Ubuntu 20.04

lwronski:
> I confirm that the issue is still present in 2.1.0-RC4 (with Ubuntu on an x86_64 arch)

[...]

PR #2625, opened 2022-12-14
bugfix: Generate native launcher for linux on the previous ubuntu image

tgodzik:
Previously, we would generate native image on ubuntu 22, which would cause issues for users of older linux distros. Now, we generate them the same way we did previously on ubuntu 20.

Fixes https://github.com/coursier/coursier/issues/2624

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -192,7 +192,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


alexarchambault (commented):
Thanks @tgodzik! I'm trying a slightly different approach in https://github.com/coursier/coursier/pull/2626, so that this doesn't hamper bumping the ubuntu version used on the CI. https://github.com/coursier/coursier/pull/2626 should also allow to build so-called "static" or "mostly static" launchers.

PR #2635, opened 2022-12-21
Update GraalVM to 22.3.0

alexarchambault:


alexarchambault:
Seems the CI failure needs to be investigated on an actual aarch64 machine...

tgodzik:
> Seems the CI failure needs to be investigated on an actual aarch64 machine...

Would a docker image on M1 machine be a good place to debug? I can take a look next week, I need to anyway install docker on the machine in the office.

alexarchambault:
@tgodzik A Linux aarch64 VM on any cloud should do. That's what I usually use in such cases.

alexarchambault:
I removed the Linux / aarch64 job in https://github.com/coursier/coursier/pull/2658. We can always investigate issues in [VirtusLab/coursier-m1](https://github.com/VirtusLab/coursier-m1).

lwronski:
I found interesting [line](https://github.com/coursier/coursier/actions/runs/3751413250/jobs/6383988540#step:5:2867) in logs. It's seems, that proxy tests using docker, which is not available in docker now.

alexarchambault:
@lwronski Yeah, it seems it went away. The docker tests are disabled too in VirtusLab/coursier-m1, but could be re-instated (as we generate the Linux launcher there from docker, we'd need to be able to run "docker-in-docker", which needs a bit of setup…)

Issue #2636, opened 2022-12-22
Error when running coursier on Linux aarch64


mkurz:
```sh
$ curl -fL https://github.com/coursier/launchers/raw/master/cs-aarch64-pc-linux.gz | gzip -d > cs && chmod +x cs
...
$ ./cs 
Fatal error: Failed to create the main Isolate. (code 8)
```
I am running Asah Linux, which basically is Arch Linux ARM, on Apple Silicon (MacBook Pro M1):
```sh
$ uname -a
Linux mkurz-macbook-pro 6.1.0-asahi-2-1-edge-ARCH #2 SMP PREEMPT_DYNAMIC Wed, 14 Dec 2022 14:33:17 +0000 aarch64 GNU/Linux
```
The error message [originates in graal here](https://github.com/oracle/graal/blob/b6e1c8d0844c475c97ad7ef52796b192b59ef2fc/substratevm/src/com.oracle.svm.core/src/com/oracle/svm/core/JavaMainWrapper.java#L386-L388).
However, I think you don't use latest graal for building the native images, maybe it would work with the latest one?

mkurz:
Is it possible this is related to `sbtn` not having being available for aarch64 linux, like described here: #2358?

lwronski:
> Is it possible this is related to `sbtn` not having being available for aarch64 linux, like described here: #2358?

Hi @mkurz, I don't think so, in this issue #2358 user be able to run corusier but fail when trying to download some apps, but in your case it throws error when starting cs.

I think a good test would be if you tried to build `cs` at Asah Linux. 
```
$ git clone --recursive https://github.com/coursier/coursier.git
$ cd coursier
$ ./mill -i show cli.nativeImage
```

If it works, it means that Scala CLI and Corusier have some misconfiguration during the building native launcher for aarch64 linux.

You can also update `graalVMVersion` to `22.3.0` and check what happen - [Deps.graalVmVersion](https://github.com/coursier/coursier/blob/f730535d8d9c8bdb885d353c5665a6785510cf7f/project/deps.sc#L74)

mkurz:
So... I was able to build _and successfully run_ the cs binary on my machine, without changing anything. I just ran `./mill -i copyLauncher artifacts/` in a fresh clone (and git submodules initialized :wink:)

[...]

PR #2639, opened 2022-12-22
Use latest Ubuntu 22.04 LTS for building aarch64 binaries

mkurz:


mkurz:
I somehow try to solve #2636, so I guess using the latest Ubuntu is probably a good idea, to use a newer gcc and other toolkits.

tgodzik:
> I somehow try to solve #2636, so I guess using the latest Ubuntu is probably a good idea, to use a newer gcc and other toolkits.

Although it seems that it might break for anyone using an earlier ubuntu :thinking:

mkurz:
> Although it seems that it might break for anyone using an earlier ubuntu :thinking:

Why do you think so?

mkurz:
@tgodzik Can you once more approve the workflows? As you can see I uncommented a condition so the arm64 binaries gets uploaded as artifacts. I would like to download them and try and see if it works for me on my machine. Thanks!

tgodzik:
> > Although it seems that it might break for anyone using an earlier ubuntu thinking
> 
> Why do you think so?

I had that issue since newer gcc was not available for ubuntu 20.04

mkurz:
One workflow fails because jitpack is down... https://github.com/jitpack/jitpack.io/issues/5337
So we need to wait until it comes back up.

mkurz:
@tgodzik So because jitpack is down, I temporary removed the test workflow, so we are not blocked here any more.
Could you once more approve the workflows? Thanks!

mkurz:
Nevermind, I set up my own repo to test the workflows.

mkurz:
Using Ubuntu 22.04 does not really help, I still get an error:
```
Fatal error: Failed to create the main Isolate. (code 9)
```
This time however it's code 9, not code 8 (whatever that means...)

PR #2640, opened 2022-12-22
Fix instructions to install the native launcher on Linux

julienrf:
The current instructions don’t work on other architectures than x86.

We resolve the host architecture with `uname -m`.

### DIFF
--- a/doc/docs/cli-installation.md
+++ b/doc/docs/cli-installation.md
@@ -39,6 +39,9 @@ Currently, we support the following architectures: x86-64 and ARM64.
 On macOS, download and run the coursier installer with
 
 ```bash
+# On Apple M1:
+$ curl -fL https://github.com/VirtusLab/coursier-m1/releases/latest/download/cs-aarch64-apple-darwin.gz | gzip -d > cs
+# Otherwise:
 $ curl -fL https://github.com/coursier/launchers/raw/master/cs-x86_64-apple-darwin.gz | gzip -d > cs
 $ chmod +x cs
 $ ./cs setup
--- a/doc/docs/cli-installation.md
+++ b/doc/docs/cli-installation.md
@@ -27,10 +27,12 @@ After the setup, you can [start using Scala](https://docs.scala-lang.org/scala3/
 On Linux, download and run the coursier installer with
 
 ```bash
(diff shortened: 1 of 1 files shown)


julienrf:
I have added a second commit to update the instructions for the Apple M1 machines. This solution is not super satisfactory but it is temporary until we streamline the generation of OS-specific packages with https://github.com/VirtusLab/scala-packager

doc/docs/cli-installation.md
```diff
@@ -27,16 +27,21 @@ After the setup, you can [start using Scala](https://docs.scala-lang.org/scala3/
 On Linux, download and run the coursier installer with
 
 ```bash
-$ curl -fL https://github.com/coursier/launchers/raw/master/cs-x86_64-pc-linux.gz | gzip -d > cs
+$ curl -fL "https://github.com/coursier/launchers/raw/master/cs-$(uname -m)-pc-linux.gz" | gzip -d > cs
 $ chmod +x cs
 $ ./cs setup
 ```
 
+Currently, we support the following architectures: x86, AMD, and ARM.
```
mkurz, line 35: Shouldn't that be ARM64 instead of ARM?

mkurz: I am pretty sure this should by ARM64, not just ARM. ARM is IMHO a way to broad definition. ARM64 whoever is a synonym for AARCH64, also see https://en.wikipedia.org/wiki/AArch64
And currently the binaries you offer are aarch64 only, so arm64 is accurate here.

alexarchambault: I agree with @mkurz. Just ARM would be 32 bit ARM (Raspberry Pi 1-3).

alexarchambault: Rather than `x86, AMD`, I'd just say `x86-64`. Here too, just `x86` means 32 bit processors.

julienrf: Indeed, thank you for the suggestions, this is fixed by 96ada0ba9

julienrf: BTW, why do we generate both `cs-amd64-pc-linux` and `cs-x86_64-pc-linux` if the underlying architecture is the same?

alexarchambault (approved):
Thanks!

PR #2239, opened 2021-10-22
Use Java 17 based GraalVM distribution

alexarchambault:

PR #2641, opened 2022-12-23
Set page size for aarch64 Linux binaries to 64k

### LINKED ISSUE #2636: Error when running coursier on Linux aarch64

mkurz:
Fixes #2636, makes coursier run on Linux running on bar metal Apple silicon and also in Linux docker containers within macOS on M1/M2 machines.

Same like we did for scala-ci here: https://github.com/VirtusLab/scala-cli/pull/1726

### DIFF
--- a/project/launchers.sc
+++ b/project/launchers.sc
@@ -79,10 +79,11 @@ trait Launchers extends CsModule {
         if (usesDocker) s"/data/$staticLibDirName"
         else staticLibDir().path.toString
       val extraOpts =
-        if (Properties.isLinux)
+        if (Properties.isLinux && arch == "aarch64")
           Seq(
             // required on the Linux / ARM64 CI in particular (not sure why)
-            "-Djdk.lang.Process.launchMechanism=vfork" // https://mbien.dev/blog/entry/custom-java-runtimes-with-jlink
+            "-Djdk.lang.Process.launchMechanism=vfork", // https://mbien.dev/blog/entry/custom-java-runtimes-with-jlink
+            "-H:PageSize=65536" // Make sure binary runs on kernels with page size set to 4k, 16 and 64k
           )
         else
           Nil


project/launchers.sc
```diff
@@ -79,10 +79,11 @@ trait Launchers extends CsModule {
         if (usesDocker) s"/data/$staticLibDirName"
         else staticLibDir().path.toString
       val extraOpts =
-        if (Properties.isLinux)
+        if (Properties.isLinux && arch == "aarch64")
           Seq(
             // required on the Linux / ARM64 CI in particular (not sure why)
-            "-Djdk.lang.Process.launchMechanism=vfork" // https://mbien.dev/blog/entry/custom-java-runtimes-with-jlink
+            "-Djdk.lang.Process.launchMechanism=vfork", // https://mbien.dev/blog/entry/custom-java-runtimes-with-jlink
```
mkurz, line 85: @alexarchambault Based on this comment, I think it should be safe to just have this property set for the aarch64 builds? You introduced that in #2239. I think we should see CI complaining if the non-aarch64 Linux binaries need to be build this flag way as well, hopefully (?)

mkurz:
Can someone approve the workflows to run? Thanks!

project/launchers.sc
alexarchambault: Yeah, that must have been an oversight. Non-aarch64 binaries are [tested on PRs too](https://github.com/coursier/coursier/blob/f730535d8d9c8bdb885d353c5665a6785510cf7f/.github/workflows/native-tests.yml), so if the CI doesn't complain, that's fine.

mkurz:
@alexarchambault Can you approve once more? I force pushed because formatting was wrong.

alexarchambault (approved):
Thanks @mkurz! I just hope this won't have unintended consequences (performance-wise, ...)

mkurz:
@alexarchambault Besides more memory usage, which can be neglected IMHO, there should hopefully not be a performance penalty. Also this changes really only effects the Linux aarch64 binaries, so nothing changes for the "average" Linux, macOS or Windows users :wink: 

Would really love to see this in RC5, thanks!

mkurz:
Just to let you know, [sbt 1.8.1 just got released](https://github.com/sbt/sbt/releases/tag/v1.8.1), which contains the same page size setting like in this pull request and like scala-cli has now as well: https://github.com/sbt/sbt/pull/7108

mkurz:
Any reason not to merge this?

tgodzik:
@alexarchambault should we merge?

PR #2651, opened 2023-01-04
Fix platform - fallback arm64 to aarch64

lwronski:


tgodzik (approved):

modules/install/src/main/scala/coursier/install/Platform.scala
```diff
@@ -6,8 +6,11 @@ object Platform {
 
   def get(os: String, arch: String): Option[String] = {
 
-    val os0   = os.toLowerCase(Locale.ROOT)
-    val arch0 = if (arch == "amd64") "x86_64" else arch
+    val os0 = os.toLowerCase(Locale.ROOT)
+    val arch0 =
+      if (arch == "amd64") "x86_64"
+      else if (arch == "arm64") "aarch64"
```
mkurz, line 12: I don't think under windows that is called arm64... Should that be checked somehow? Like
```suggestion
      else if (arch == "arm64" && (os0.contains("linux") || os0.contains("mac"))) "aarch64"
```
but not 100% sure maybe I don't have all the context of this change, I didn't really had a look at the surroundings.

alexarchambault: @mkurz What is it called on Windows?

No need to check for the OS here though IMO, I guess if `arch` is `"arm64"`, it's pretty safe to change it to `"aarch64"`, no matter the OS;

alexarchambault (approved):

PR #2654, opened 2023-01-06
Update README.md

calvinvette:
Added first paragraph of what Coursier is from website as a brief description.


tgodzik (approved):
Thanks! Makes sense to add it.

PR #2655, opened 2023-01-07
Remove recursion from Complete.hasOrg

### LINKED ISSUE #1700: `cs complete` for snapsots not working like resolve or bootstrap
Is there a way to do the following:
```
cs complete -r sonatype:snapshots org.scalameta:metals_2.12:
```
This only returns stable version rather than the snapshots I'd expect.

However if I target a snapshot directly with resolve, it does work.

```
cs resolve -r sonatype:snapshots org.scalameta:metals_2.12:0.8.4+127-65087a2b-SNAPSHOT
```

If resolve is able to resolve this snapshot, I sort of expected `complete` to be able to pick up on all the published snapshots. Is there a way to do this that I'm missing by chance?

jackkoenig:
This is primarily to fix an issue where the Sonatype snapshots repository disallows listing the top-level directory, but the recursion appears to have no functional benefit. This should also make complete faster by reducing the number of queries.

Fixes https://github.com/coursier/coursier/issues/1700

Many thanks and full credit to @keynmol for doing the real legwork debugging the issue in https://github.com/coursier/coursier/issues/1700#issuecomment-912498601.

I'm opening this as a draft PR because I am not sure how best to test this, especially since it relies on peculiarities of the Sonatype snapshots repository, and my limited grepping did not find any tests for the behavior of Complete.


jackkoenig:
I also suspect that the `partial` argument to `hasOrg` may not be needed but I'm already over my skis a bit with this change so I figured I'd start here 🙂

modules/core/shared/src/main/scala/coursier/core/Repository.scala
```diff
@@ -162,31 +162,15 @@ object Repository {
...
-      val idx = orgInput.input.lastIndexOf('.')
-      if (idx > 0) {
-        val truncatedInput = Complete.Input.Org(orgInput.input.take(idx))
-        hasOrg(truncatedInput, partial = true).flatMap {
-          case false =>
-            F.point(false)
-          case true =>
-            check
-        }
+    ): F[Boolean] =
```
jackkoenig, line 165: I recommend hiding whitespace on the diff to see that this is just taking the expression assigned to `check` and returning it.

jackkoenig:
This isn't a complete fix for compete on sonatype snapshots, complete still doesn't work for organizations, but it works for artifacts and versions which seems like a pretty big improvement to me.

I proved this worked to myself with the following script (running it with and without this change):
```bash
#!/usr/bin/env bash -x

CS='mill -s --disable-ticker -i cli.run'
$CS complete-dep --cache localcache --no-default -r sonatype:snapshots edu.berkeley.cs:chisel3_2.12:3.2
$CS complete-dep --cache localcache --no-default -r sonatype:snapshots edu.berkeley.cs:chisel3_2.12:
$CS complete-dep --cache localcache --no-default -r sonatype:snapshots edu.berkeley.cs:chisel3
$CS complete-dep --cache localcache --no-default -r sonatype:snapshots edu.berkeley.cs:
$CS complete-dep --cache localcache --no-default -r sonatype:snapshots edu.berkeley
$CS complete-dep --cache localcache --no-default -r sonatype:snapshots edu
$CS complete-dep --cache localcache --no-default -r sonatype:releases edu.berkeley.cs:chisel3_2.12:3.2
$CS complete-dep --cache localcache --no-default -r sonatype:releases edu.berkeley.cs:chisel3_2.12:
$CS complete-dep --cache localcache --no-default -r sonatype:releases edu.berkeley.cs:chisel3
$CS complete-dep --cache localcache --no-default -r sonatype:releases edu.berkeley.cs:
$CS complete-dep --cache localcache --no-default -r sonatype:releases edu.berkeley
$CS complete-dep --cache localcache --no-default -r sonatype:releases edu
```

You'll notice that `edu.berkeley` and `edu` still don't complete correctly for `sonatype:snapshots` but the rest do.

alexarchambault (commented):
Thanks @jackkoenig!

Do you mind if I ponder that a little? IIRC, this recursion was related to caching, with things being properly cached with the recursion, and not correctly cached without it. I don't recall if this still applies though, which what I'd like to check…

jackkoenig:
Sorry I missed your response @alexarchambault! Yeah this isn't urgent by any means but I hope some solution eventually makes its way into coursier! Thanks for the great tool 🙂

PR #2666, opened 2023-01-19
Add generating sdkman packages with cs

lwronski:


lwronski:
I confirm that these changes generate the correct package for SDKMAN. [Here](https://github.com/coursier/coursier/actions/runs/3965873848) is a test CI where SDKMAN packages were successfully generated, they are available in artifacts.

alexarchambault (approved):
Thanks!

PR #2667, opened 2023-01-20
Always upload the generated launchers in the CI

julienrf:
We used to upload the generated launchers only in case of release. Instead, we propose do upload them on every CI run.


julienrf:
We can’t do it with the jar-based launchers because they require the .jar artifacts to have been published to Sonatype, which happens only when we cut a release.

.github/workflows/native-tests.yml
```diff
@@ -26,6 +26,15 @@ jobs:
         if: runner.os == 'Linux'
       - run: .github/scripts/maybe-with-graalvm-home.sh nativeTests
         shell: bash
+      - name: Copy artifacts
+        run: .github/scripts/maybe-with-graalvm-home.sh copyLauncher artifacts/
```
julienrf, line 30: That line fails on the Windows CI with a `FileAlreadyExistsException`. I don’t understand what is going on because what I do is similar to what is done in the `ci.yml`. Does `nativeTests` also copy the JVM?

alexarchambault:
Cherry-picking [that commit](https://github.com/alexarchambault/coursier/commit/d20552adb2eeaad910d5bb22ec68ef5777927ece) should fix that (seems I can't push on the PR branch here).

.github/workflows/native-tests.yml
alexarchambault: Cherry-picking [that commit](https://github.com/alexarchambault/coursier/commit/d20552adb2eeaad910d5bb22ec68ef5777927ece) should fix that (seems I can't push on the PR branch here).

alexarchambault (approved):

PR #2675, opened 2023-01-27
Check that all the apps are installed, document the prebuiltBinaries entries

julienrf:


modules/install/src/main/scala/coursier/install/InstallDir.scala
```diff
@@ -268,7 +268,10 @@ import scala.util.Properties
       )
 
       val appArtifacts = prebuiltOrNotFoundUrls0 match {
-        case Left(_)  => desc.artifacts(cache, verbosity)
+        case Left(_) =>
+          if (verbosity >= 0)
+            System.err.println("Prebuilt launcher not found, trying to build the app from classfiles")
```
julienrf, line 273: Now, `cs install --verbose --install-platform aarch64-apple-darwin sbtn` explains slightly better the cause of the failure:

~~~ text
Using install directory /home/julien/.local/share/coursier/bin
https://repo1.maven.org/maven2/io/get-coursier/apps/maven-metadata.xml
  No new update since 2022-12-21 14:05:06
Found app sbtn in channel io.get-coursier:apps
Checking prebuilt launcher at https://github.com/sbt/sbt/releases/download/v1.8.2/sbt-1.8.2.zip
Found prebuilt launcher at https://github.com/sbt/sbt/releases/download/v1.8.2/sbt-1.8.2.zip (4.0 KiB)
Checking prebuilt launcher at https://github.com/sbt/sbt/releases/download/v1.8.2/sbt-1.8.2.zip
Found prebuilt launcher at https://github.com/sbt/sbt/releases/download/v1.8.2/sbt-1.8.2.zip (4.0 KiB)
Checking prebuilt launcher at https://github.com/sbt/sbt/releases/download/v1.8.1/sbt-1.8.1.zip
Found prebuilt launcher at https://github.com/sbt/sbt/releases/download/v1.8.1/sbt-1.8.1.zip (4.0 KiB)
Checking prebuilt launcher at https://github.com/sbt/sbt/releases/download/v1.8.0/sbt-1.8.0.zip
Found prebuilt launcher at https://github.com/sbt/sbt/releases/download/v1.8.0/sbt-1.8.0.zip (4.0 KiB)
Checking prebuilt launcher at https://github.com/sbt/sbt/releases/download/v1.7.3/sbt-1.7.3.zip
Found prebuilt launcher at https://github.com/sbt/sbt/releases/download/v1.7.3/sbt-1.7.3.zip (4.0 KiB)
Prebuilt binary not found, trying to build the app from classfiles
…
No main class found
~~~

I think in the case of `sbtn` we don’t really want to build the launcher ourselves, do we? I’ve seen there is a setting `preferPrebuilt`, maybe we should have something finer-grained like `RequirePrebuilt`, `PreferBrebuilt`, and `NoPreference`?

julienrf: Ah, I see that you even suggested to fail in case the prebuilt launcher is not found instead of even trying to build it ourselves: https://github.com/coursier/apps/issues/175#issuecomment-1282045723

alexarchambault: Yeah, ideally I'd like to get rid of the code building launchers

julienrf: What would be the steps to get there?

alexarchambault: I guess we could just remove that code now…

alexarchambault (commented):
Looks good!

@julienrf Could you edit the first 3 commits, so that they're properly formatted upfront and the "scalafmt" commit isn't needed? That way, we can merge with a standard merge (and keep a clean git history in the main branch). (I'd be ok editing the commits myself, if I had the right to push commits to the PR branch - but PR edits by maintainers was disabled when you opened the PR it seems)

modules/install/src/main/scala/coursier/install/InstallDir.scala
julienrf: Do you mean removing the `println` I added, or removing the code that tries to build the launcher?

julienrf: I just removed the println.

alexarchambault (approved):

PR #2690, opened 2023-02-14
Fix installation instructions

julienrf:
Since the ARM64 binaries are built on a different repository, the current instructions don’t work.


julienrf:
Hopefully these changes are temporary, and soon we will be able to tell users to run `sudo apt-get install scala` instead :)

tgodzik (approved):
LGTM!

PR #2691, opened 2023-02-15
Fix link to coursier-s3

fthomas:
rtfpessoa/coursier-s3 does not exist anymore and seems to have been moved to alexarchambault/coursier-s3

### DIFF
--- a/doc/docs/extra.md
+++ b/doc/docs/extra.md
@@ -102,4 +102,4 @@ Coursier will search for your plugin with the following order on various classlo
 
 Real Word Examples:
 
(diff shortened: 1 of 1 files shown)


tgodzik (approved):

tgodzik:
Thanks @fthomas ! :tada:

tgodzik:
Och, my bad. But the other fork doesn't seem too well maintained either. Is it something we would want to have under coursier org then?

fthomas:
If you want to list a more recent protocol handler there, there is for example @987Nabil's [gar-coursier](https://github.com/987Nabil/gar-coursier) that handles `artifactregistry://` URLs.

PR #2701, opened 2023-02-26
Set defaults for http.proxyProtocol and https.proxyProtocol

cheister:
https://github.com/coursier/coursier/pull/2440 added the requirement to specify `http.proxyProtocol` or `https.proxyProtocol` when using a proxy.  

This PR sets default values for these options so they don't have to be set to use the proxy and keeps backwards compatibility with proxy settings from previous versions.


alexarchambault (approved):
Thanks!

PR #2706, opened 2023-03-06
Document Versions API, add helpers

kubukoz:
Just opening this to get some early feedback.


doc/docs/api.md
```diff
@@ -341,3 +341,21 @@ ioFiles: cats.effect.IO[Seq[java.io.File]]
 
 Note that this example requires the `coursier-cats-interop` module
```
kubukoz, line 342: I think this whole page could use a bit of a reorg, e.g. there's some duplication around the whole future/io part. I'd extract that to a separate section (outside Fetch and Resolve). What do you think?

alexarchambault (approved):
Thanks!

PR #2713, opened 2023-03-15
[doc] Clarify custom ivy repository usage

ex0ns:
I am not sure that this is required, but I struggle a bit to understand that this was possible and I though it could be useful to keep that information somewhere.


alexarchambault (approved):
Thanks!

PR #2718, opened 2023-03-17
Map packaging to non-optional

### LINKED ISSUE #2377: `cs fetch` caches a POM with no artifact, and fails to download on subsequent runs with different repositories
The maven package `javax.media:jai_core:1.1.3` [exists in an incomplete state](https://repo1.maven.org/maven2/javax/media/jai_core/1.1.3/) on `repo1.maven.org/maven2`, which is one of Coursier's default repositories.

To satisfy dependencies from `org.osgeo`, the package is hosted at `https://repo.osgeo.org/repository/release`. The JAR is available from there.

We've discovered that you can get Coursier into a state where it will permanently fail to download this artifact.

## Symptoms

There are three separate problematic behaviours that Coursier is exhibiting here:

1. If `repo1.maven.org` comes before `repo.osgeo.org` in Coursier's resolve order, Coursier will download the POM at `repo1.maven.org`, rather than trying the other configured repository to see if it possible to resolve there
2. If `--no-default` is not specified, `repo1.maven.org` is _always_ tested before user-supplied repositories, which leads to unexpected behaviour for an inexperienced user (re-ordering the repositories list will not do anything until `--no-default` is supplied, which will break other resolves)
3. Coursier _eagerly caches_ the first POM it finds for a given coordinate, regardless of whether a JAR can be fetched next to that POM, and regardless of which repositories are configured on subsequent runs. Consequently, once `javax.media:jai_core:1.1.3` is attempted to be fetched from `repo1.maven.org`, subsequent Coursier runs will always fail to fetch the JAR, regardless of what repositories are fetched. The current work-around for this is to blow away the entire Coursier cache, which is undesirable

## Expected behaviours

1. Coursier should try all configured repositories until an artifact is downloaded. Downloading just the POM for that artifact is not sufficient to show success.
2. Coursier should check user-supplied repositories before the default repositories
3. Coursier should not cache this POM because it does not supply a usable artifact, and causes Coursier to actively get in the way of fixing these symptoms
4. (BONUS) Coursier should cache each artifact according to the repository it was found at, so that if the repositories list ever changes (to get a different packaging of an artifact maybe?) it will download it more eagerly.

## How to reproduce

1. Work in a tmpdir: `mkdir /tmp/coursier && cd /tmp/coursier`

2.
Delete the coursier download cache, to start with clean state. E.g., on MacOS: `rm -rf ~/Library/Caches/Coursier/v1/`.

5. 
Fetch the broken artifact from the bad repository:
```
$ cs fetch -r=https://repo1.maven.org/maven2 --no-default --json-output-file=coursier_report.json javax.media:jai_core:1.1.3

$ cat coursier_report.json | jq -M .dependencies[0].file
null
```

6.
Try and refetch with the good repository:
```
$ rm -f coursier_report.json  # Just to be sure that isn't confusing us somehow
$ cs fetch -r=https://repo.osgeo.org/repository/release -r=https://repo1.maven.org/maven2 --no-default --json-output-file=coursier_report.json javax.media:jai_core:1.1.3

$ cat coursier_report.json | jq -M .dependencies[0].file
null
```

7.
Delete the coursier cache again, as above.

8.
Refetch with the good repository, as above:
```
$ rm -f coursier_report.json 
$ cs fetch -r=https://repo.osgeo.org/repository/release -r=https://repo1.maven.org/maven2 --no-default --json-output-file=coursier_report.json javax.media:jai_core:1.1.3

$ cat coursier_report.json | jq -M .dependencies[0].file
"/Users/benjyw/Library/Caches/Coursier/v1/https/repo.osgeo.org/repository/release/javax/media/jai_core/1.1.3/jai_core-1.1.3.jar"
```

dwijnand:
Fixes #2377.  Depends on https://github.com/coursier/test-metadata/pull/8.

### DIFF
--- a/modules/coursier/jvm/src/test/scala/coursier/FetchTests.scala
+++ b/modules/coursier/jvm/src/test/scala/coursier/FetchTests.scala
@@ -5,6 +5,7 @@ import java.io.File
 import coursier.core.{Activation, Configuration, Extension}
 import coursier.params.ResolutionParams
 import coursier.ivy.IvyRepository
+import coursier.util.Task
 import utest._
 
 import scala.async.Async.{async, await}
@@ -333,6 +334,33 @@ object FetchTests extends TestSuite {
       }
     }
 
-  }
+    test("jai_core") {
+      test("ko") {
+        val cache1 = cache match {
+          case cache: coursier.cache.MockCache[Task] =>
+            cache.withDummyArtifact(_ => false)
+        }
+        try {
+          val res = fetch
+            .withCache(cache1)
+            .addDependencies(dep"javax.media:jai_core:1.1.3")
+            .run()
+          println(s"res=$res")
+          assert(false)
+        }
+        catch { case _: coursier.error.FetchError.DownloadingArtifacts => () }
+      }
 
+      test("ok") - async {
+        val res = await {
+          val osgeo = MavenRepository("https://repo.osgeo.org/repository/release")
+          fetch
+            .withRepositories(Seq(osgeo, Repositories.central))
+            .addDependencies(dep"javax.media:jai_core:1.1.3")
+            .future()
+        }
+        assert(res(0) != null)
+      }
+    }
+  }
 }
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,7 +1,6 @@
 [submodule "tests/metadata"]
 	path = modules/tests/metadata
 	url = https://github.com/coursier/test-metadata.git
-	branch = failure-isnt-optional
 [submodule "modules/directories"]
 	path = modules/directories
 	url = https://github.com/coursier/directories-jvm.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,7 @@
 [submodule "tests/metadata"]
 	path = modules/tests/metadata
 	url = https://github.com/coursier/test-metadata.git
+	branch = failure-isnt-optional
 [submodule "modules/directories"]
 	path = modules/directories
 	url = https://github.com/coursier/directories-jvm.git
--- a/modules/core/shared/src/main/scala/coursier/maven/MavenRepositoryInternal.scala
+++ b/modules/core/shared/src/main/scala/coursier/maven/MavenRepositoryInternal.scala
@@ -422,7 +422,7 @@ private[coursier] class MavenRepositoryInternal(
           ) -> optional
       }
 
-      val allPubs = packagingPublicationOpt.map(_ -> true).toSeq ++ extraPubs
+      val allPubs = packagingPublicationOpt.map(_ -> false).toSeq ++ extraPubs
       val optional = allPubs
         .groupBy(_._1)
         .view
--- a/modules/coursier/jvm/src/test/scala/coursier/ArtifactsTests.scala
+++ b/modules/coursier/jvm/src/test/scala/coursier/ArtifactsTests.scala
@@ -322,13 +322,18 @@ object ArtifactsTests extends TestSuite {
           .future()
       }
 
-      val artifacts = Artifacts.artifacts(res, Set.empty, None, None, true).map(_._3).distinct
-      val groupedArtifacts = Artifacts.groupArtifacts(artifacts)
+      val databindUrl =
+        "https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.6.7.2/jackson-databind-2.6.7.2.jar"
+
+      val artifacts   = Artifacts.artifacts(res, Set.empty, None, None, true).map(_._3).distinct
+      val databindOpt = artifacts.find(_.url == databindUrl)
+      assert(databindOpt.isDefined)
+      val groupedArtifacts = Artifacts.groupArtifacts(artifacts :+ databindOpt.get)
 
       assert(groupedArtifacts.length == 2)
 
       val expectedDuplicatedUrls = Set(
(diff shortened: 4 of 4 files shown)


dwijnand:
Hey @alexarchambault (or @julienrf) any chance this could be reviewed, merged and released, please?

alexarchambault:
@dwijnand I'll have a closer look later, but at first sight, this is not a trivial change… Packaging has been optional on purpose. Some dependencies do miss a JAR, and they shouldn't fail.

About the repo linked in https://github.com/coursier/coursier/issues/2377, I'd need to have a closer look, but at first, I'd assume the repository to be broken.

alexarchambault:
That said, if I read https://github.com/coursier/coursier/issues/2377#issuecomment-1309691107 correctly, and if your change is only about the case where the dependency has an explicit `<packaging>jar</packaging>`, that might work…

modules/install/src/test/scala/coursier/install/InstallTests.scala
```diff
@@ -803,6 +803,7 @@ object InstallTests extends TestSuite {
 
         val scala2CompilerJars =
           Seq(
+            "https://repo1.maven.org/maven2/org/scala-lang/modules/scala-xml_2.12/1.0.6/scala-xml_2.12-1.0.6.jar",
```
julienrf, line 806: Is this change related to the PR?

modules/coursier/jvm/src/test/scala/coursier/FetchTests.scala
julienrf, line 1: Would it be possible to unify the code style of the tests? (I see one that uses `.future()` and the other uses `.run()`. I think we should use `.run()` everywhere. Also, there is a `println()` that should be removed I guess)

modules/tests/shared/src/test/resources/artifacts/com.splicemachine/splice_spark/2.8.0.1915-SNAPSHOT
```diff
@@ -223,7 +223,6 @@ https://repo1.maven.org/maven2/xml-apis/xml-apis/1.3.04/xml-apis-1.3.04.jar
 http://repository.mapr.com/maven/org/apache/hadoop/hadoop-yarn-server-nodemanager/2.7.0-mapr-1808/hadoop-yarn-server-nodemanager-2.7.0-mapr-1808.jar
 http://repository.splicemachine.com/nexus/content/groups/public/com/splicemachine/splice_encoding/2.8.0.1915-SNAPSHOT/splice_encoding-2.8.0.1915-20190427.010754-1.jar
 https://repo1.maven.org/maven2/software/amazon/ion/ion-java/1.0.1/ion-java-1.0.1.jar
-https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.6.7.1/jackson-databind-2.6.7.1.jar
```
julienrf, line 226: Why is this change necessary?

julienrf (commented):
IIUC https://github.com/coursier/coursier/issues/2377#issuecomment-1309691107 there is an inconsistency between the way Coursier reads a dependency from a pom or from the command-line.

From the command-line (e.g. `cs fetch javax.media:jai_core:1.1.3,type=jar`), when we set the type to `jar`, Coursier fails to resolve the dependency if the .jar file is absent.

However, when the dependency is read from a pom file with `<packaging>jar</packaging>`, Coursier marks the .jar file as optional and does not fail the resolution if the file is absent.

This PR fixes this inconsistency, so the approach looks good to me. Please see my other comments below.

alexarchambault:
I agree with @julienrf comments. I'd like to take the time to understand why the databind and the scala-xml changes are necessary

modules/tests/shared/src/test/resources/artifacts/com.splicemachine/splice_spark/2.8.0.1915-SNAPSHOT
dwijnand: The fix removes the duplicate jackson-databind jar (there's one on line 121).

dwijnand:
> AFAIU, this PR may not completely address the linked issue because it will not, by default, expand dependency descriptors to `type=jar,optional=false`.

I don't know what that means, outside of not making packaging=jar non-optional..

julienrf:
> I don't know what that means, outside of not making packaging=jar non-optional..

I might be missing something, but what I meant is that with this PR Courser will not set the packaging to jar by default when reading dependences from a pom. It will still have to be specified in the pom. Probably, this is fine and should not hold this PR.

julienrf (approved):

alexarchambault:
I'll try to find time to have a closer look in the coming days (until next Monday say, don't merge until then)

dwijnand:
Thank you ❤️

alexarchambault (approved):

alexarchambault:
So I had a closer look at the changes in test fixtures that looked suspicious. https://github.com/coursier/coursier/pull/2776 and https://github.com/coursier/coursier/pull/2777 deal with those. I just pushed a rebased version of the original code of this PR, and the diff looks much nicer now 🎉

I don't feel like writing a long explanation about this right now, given it's quite late for me 😬 I might tomorrow or later during the week.

In any case, thanks for having pushed for this, @dwijnand! I'll merge if the CI gets green (it should), and this'll be part of the next release.

dwijnand:
Very nice, thank you.  It's green, btw.

dwijnand:
@alexarchambault could we merge this, please?

PR #2722, opened 2023-03-20
Fix #2677 - instructions for zsh completions

### LINKED ISSUE #2677: Setup instructions for zsh completions don't reference a valid cs completions command
According to https://get-coursier.io/docs/cli-installation#zsh-completions we can do this:

```
mkdir -p ~/.zsh/completion
cs --completions zsh > ~/.zsh/completion/cs
echo 'fpath=(~/.zsh/completion $fpath)' >> ~/.zshrc
echo 'autoload -Uz compinit ; compinit' >> ~/.zshrc
```

However, the `cs --completions zsh` command doesn't actually seem to do anything any more:

```
% cs --completions zsh
Usage: /Users/james.johnston/Library/Application Support/Coursier/bin/.cs.aux <COMMAND>
Coursier is the Scala application and artifact manager.
It can install Scala applications and setup your Scala development environment.
It can also download and cache artifacts from the web.

Install application commands:
  install    Install an application from its descriptor.
  list       List all currently installed applications.
  setup      Setup a machine for Scala development.
  uninstall  Uninstall one or more applications.
  update     Update one or more applications.
<snip>
% cs version
2.1.0-RC5
```

The version is what was downloaded today when I installed cs using `curl -fL https://github.com/VirtusLab/coursier-m1/releases/latest/download/cs-aarch64-apple-darwin.gz | gzip -d > cs` command as documented.

Kevin-Lee:
Fix #2677 - instructions for zsh completions

I've never got zsh completions working with the existing instructions, but this is the only way that works for me.

### DIFF
--- a/doc/docs/cli-installation.md
+++ b/doc/docs/cli-installation.md
@@ -164,7 +164,15 @@ coursier completion data into your completions directory.
 You can install the completions with:
 ```bash
 mkdir -p ~/.zsh/completion
-cs --completions zsh > ~/.zsh/completion/cs
+```
+```bash
+echo '#compdef _cs cs
+
+function _cs {
+  eval "$(cs complete zsh-v1 $CURRENT $words[@])"
+}' > ~/.zsh/completion/_cs
+```
+```bash
 echo 'fpath=(~/.zsh/completion $fpath)' >> ~/.zshrc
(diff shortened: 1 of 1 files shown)


tgodzik (approved):

Kevin-Lee:
@tgodzik Thank you!

PR #2747, opened 2023-04-21
Use class-path-util library

alexarchambault:
Living [here](https://github.com/coursier/class-path-util)


alexarchambault:
Waiting for the sync to Maven Central

PR #2757, opened 2023-05-04
Extra check of recent Windows JNI issue 

alexarchambault:
Follow-up of https://github.com/coursier/coursier/pull/2756


alexarchambault:
Added in https://github.com/coursier/coursier/pull/2756

PR #2773, opened 2023-05-29
feat: offer basic support for fish for Java commands

ckipp01:
This PR offers some basic support for fish users that want to use coursier to manage their Java installations. Before this everything was `.profile`ish based, which isn't supported by fish. It works a bit differently and fish users basically have the choice to only use the CLI when setting env variables with things like `set -Ux JAVA_HOME <whatever>` or then never have to update any files, or they can update their fish configs, which is typically located at `XDG_CONFIG_HOME` under `fish/fish.config`. Since we already had everything set up to add to files I went that route.

So this offers support for `--env`, for example:

```
❯ mill -i cli.run java --jvm 17 --env
[503/503] cli.run
set -x CS_FORMER_JAVA_HOME "$JAVA_HOME"
set -x JAVA_HOME "/Users/ckipp/Library/Caches/Coursier/arc/https/github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.7%252B7/OpenJDK17U-jdk_aarch64_mac_hotspot_17.0.7_7.tar.gz/jdk-17.0.7+7/Contents/Home"
```

It also supports the `--disable` flag. And finally it can now update your fish config when you're using `--setup`. For example:

```
❯ mill -i cli.run java --jvm 17 --setup
[503/503] cli.run
Checking if ~/.config/fish/config.fish need(s) updating.
Some shell configuration files were updated. It is recommended to close this terminal once the setup command is done, and open a new one for the changes to be taken into account.
```

And you'll then notice your fish config has the following:

```
# >>> JVM installed by coursier >>>
set -gx JAVA_HOME "/Users/ckipp/Library/Caches/Coursier/arc/https/github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.7%252B7/OpenJDK17U-jdk_aarch64_mac_hotspot_17.0.7_7.tar.gz/jdk-17.0.7+7/Contents/Home"
# <<< JVM installed by coursier <<<
```

Which correct sets your `JAVA_HOME`:

```
❯ echo $JAVA_HOME
/Users/ckipp/Library/Caches/Coursier/arc/https/github.com/adoptium/temurin17-binaries/releases/download/jdk-17.0.7%252B7/OpenJDK17U-jdk_aarch64_mac_hotspot_17.0.7_7.tar.gz/jdk-17.0.7+7/Contents/Home
```

All the existing behavior for non-fish shell users should remain the same.


tgodzik:
Hmm... coursier doesn't use case app yet? I remember there was a lot of things done there for completions etc. (not for fish though yet)

ckipp01:
> Hmm... coursier doesn't use case app yet? I remember there was a lot of things done there for completions etc. (not for fish though yet)

Yea indeed, that's a whole other battle. I updated the description to make it clear, but this isn't offering completions, but rather just supporting being able to manage your java installation with fish.

alexarchambault (approved):
Nice! If you feel like it, you might want to add non-regression tests to [`ProfileUpdaterTests`](https://github.com/coursier/coursier/blob/929301cd078b6ba13ea78d5065cb07130576839a/modules/env/src/test/scala/coursier/env/ProfileUpdaterTests.scala)

ckipp01:
> Nice! If you feel like it, you might want to add non-regression tests to [ProfileUpdaterTests](https://github.com/coursier/coursier/blob/929301cd078b6ba13ea78d5065cb07130576839a/modules/env/src/test/scala/coursier/env/ProfileUpdaterTests.scala)

Yes let me try to add something, because with the changed logic, the shell is now picked up in various places where it wasn't before and now on fish the tests actually don't pass anymore on a machine using fish.

alexarchambault (approved):
Looks great, thanks @ckipp01!

PR #2791, opened 2023-06-23
Add `url` and `checksums` to JSON report.

### LINKED ISSUE #1487: add ability to emit a lock file
the coursier CLI has a resolve option and `-t` but the output is more useful for humans than tools.

Would you be open to a PR to add the ability to emit a JSON lock file?

I imagine it would list all the constraints that went in (all the artifacts + versions), and then list the minDependencies optionally including sha256 hashes of all the artifacts.

This is suitable for checking into a version control for a fully reproducible description of the classpath used to build.

This protects people from mutable changes in repositories (which are possible) causing hard to diagnose issues (not noticing that the resolution changed).

If this sounds like it would be useful, I can see about sending some PRs to implement this.

yilinwei:
This adds the `url` and `checksums` to the JSON report produced via `fetch`.

Fixes #1487.

Ideally I'd like to add the metadata files and the associated hashes so I can recreate the coursier cache within a sandbox, which would allow downstream build tools such as `sbt` or `mill` to build in a sandboxed environment, but that can be done in a follow up PR.


yilinwei:
By way of an update, this works quite well apart from the fact that the set of pom's needs to be extended to allow coursier to resolve the dependency properly.

What I mean is:

- Suppose we have dependencies foo, bar, and baz.
- foo and bar both depend on baz but different minor versions

then coursier needs the pom for both minor versions but not the jar for the evicted version.

cosmicexplorer:
Is this still something worth doing @yilinwei? I ask because I designed and implemented the `--report` json metadata format to `pip install`. The format we reproduced is actually not standardized into a PEP yet and it would be extremely cool if these fields could be standardized across tooling.

coursier's metadata resolve json output from @wisechengyi was a massive inspiration for my work on metadata  resolves in pip (which remains ongoing).

Issue #2678, opened 2023-01-28
cs setup can't find a JVM


w-r-z-k:
When I run the setup process from the website:

```
$ curl -fL https://github.com/VirtusLab/coursier-m1/releases/latest/download/cs-aarch64-apple-darwin.gz | gzip -d > cs
$ chmod +x cs
$ ./cs setup
```

it says:

```
Checking if a JVM is installed
JVM adopt@1.8+ not found in index: JVM adopt not found
```

I thought the point of coursier was so that you didn't need to deal with setting up java.  The website says "They will also install a JVM if none is found on the system."

tgodzik:
Thanks for reporting! Looks like we don't have adopt available in the indexes, we should most likely change the defaults there (I think Zulu should be present.)

julienrf:
I agree with you Tomasz, is this something you could take care of?

julienrf:
Possible workaround: run `cs setup --jvm 17`

jpassaro:
bumping this, ran into it today and it's pretty confusing on its face. fortunately @julienrf's workaround was effective. I might suggest, if it's not too hard, special-casing the situation where the default can't be found and making some explicit user recommendations (e.g. explicit `--jvm`, check network connection, report as a bug, etc)

PR #2796, opened 2023-06-30
Set default JVM to adoptium@11

julienrf:


julienrf:
May fix #2678

julienrf:
The CI is green but I am not sure how well this is covered by the tests :-/ @alexarchambault what do you think?

tgodzik (approved):
Looks good, but I'm also not sure if there will be any problems caused by this change,

alexarchambault (approved):
Thanks @julienrf! Merging, but we might as well update to Java 17…

PR #2817, opened 2023-08-04
Cross-compile CLI module to 2.13

keynmol:
For the purposes of programmatically accessing some coursier features, it would be cool to publish it for 2.13 as well, to use from Scala 2.13/3

1. Remove monadless dependency and transform back to monadic flow (very little change, I tried to make it nicer with better-monadic-for, but it didn't work - will try again later)
2. Modify the built to support cross-compiled modules
3. Remove launchers for Scala Native 0.3 and 0.4.0-M2

I kept the 2.12 in there in case there are people out there depending on the artifact with an explicit suffix.

The launchers also all refer explicitly to the 2.12 version of CLI, to keep old behaviour (specifically the default scala version when resolving/fetching/etc.)


tgodzik (approved):
Looks good from my side. @alexarchambault do you want to take a look?

tgodzik:
@keynmol how urgent  is this PR? Should we try merge and release? @alexarchambault I am not sure how comfortable are you with releasing?

alexarchambault:
Merged via https://github.com/coursier/coursier/pull/2834, thanks @keynmol!

PR #2823, opened 2023-08-24
CVE-2022-46751 external entity reference vulnerability

adpi2:
See [CVE-2022-46751](https://github.com/advisories/GHSA-2jc4-r94c-rp7h) to get all the details. The fix is inspired by https://github.com/apache/ant-ivy/commit/2be17bc18b0e1d4123007d579e43ba1a4b6fab3d.

Ping @eed3si9n @sjrd

### DIFF
--- a/modules/core/jvm/src/main/scala/coursier/core/compatibility/package.scala
+++ b/modules/core/jvm/src/main/scala/coursier/core/compatibility/package.scala
@@ -2,7 +2,9 @@ package coursier.core
 
 import java.io.CharArrayReader
 import java.util.Locale
+import javax.xml.parsers.ParserConfigurationException
 import javax.xml.parsers.SAXParserFactory
+import javax.xml.XMLConstants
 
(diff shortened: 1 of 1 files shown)


alexarchambault (approved):
Nice, thanks! If the CI is green, let’s merge

modules/core/jvm/src/main/scala/coursier/core/compatibility/package.scala
```diff
@@ -108,6 +110,30 @@ package object compatibility {
   private lazy val spf = {
     val spf0 = SAXParserFactory.newInstance()
     spf0.setNamespaceAware(false)
+
+    // Fixing CVE-2022-46751: External Entity Reference Vulnerability
```
jtjeferreira, line 114: since scala-xml is already in the classpath, why not use `Xml.parser`? https://github.com/scala/scala-xml/blob/main/shared/src/main/scala/scala/xml/factory/XMLLoader.scala#L28-L48

PR #2874, opened 2023-11-10
Update Mill to 0.10.13, cleanup and fix deprecations

lefou:
- Update Mill to 0.10.13
- Some obvious review fixes and updates
- Enable some warnings and deprecations in build scripts
- Fixed various compiler warnings
- Replaced deprecated use of `Tests` traits


project/modules/cache0.sc
```diff
@@ -59,4 +59,6 @@ trait CacheJvmBase extends Cache with CsMima with Shading {
       super.sources() ++ Seq(PathRef(dest))
     }
   }
+  @deprecated("Use CacheJvmBaseTests")
+  trait Tests extends CacheJvmBaseTests
```
tgodzik, line 63: Could we remove it altogether? It's only used in the coursier build, no?

tgodzik (commented):
Looks good! Just one question.

Do you want to add anything more here?

project/modules/cache0.sc
lefou: Yeah, I needed it in-between to let the compiler give me more precise messages. I'll remove later.

lefou:
> Looks good! Just one question.
> 
> Do you want to add anything more here?

I want to migrate to Mill 0.11, but this is still a bit far, so a merge now makes sense. I'll remove the draft state once I think I'm done.

tgodzik (approved):
LGTM!

PR #2880, opened 2023-11-13
Migrate to Mill 0.11.5

lefou:
This migrates the build system from Mill 0.10.13 to 0.11.5.


lefou:
I think this is good to go. I bumped most build dependencies. jarjarabrams isn't up-to-date due to some incompatible API changes. 

With the move to Mill 0.11, there is an opportunity to migrate all those included `.sc` files into the Mill meta build, where we can have ordinary `.scala` files under `mill-build/src/`, but this would be a rather large change, so this should be a separate PR.

alexarchambault:
Merging, thanks a lot @lefou, for having opened that!

Issue #1846, opened 2020-09-09
Allow fetching transitive provided dependencies


dwijnand:
From observations using Coursier as a library in MiMa it seems that transitive provided dependencies aren't returned from fetch.  Provided dependencies shouldn't be present when running, but my usage is at compile time (parsing classfiles in the jar).

In particular this means that sbt plugins, that (automatically) have a transitive provided dependency on sbt, don't get sbt as a dependency, resulting in bad MiMa reports.

alexarchambault:
> sbt plugins […] don't get sbt as a dependency, resulting in bad MiMa reports.

I'm not sure I understand how this is a problem.

Provided dependencies of transitive dependencies are always ignored during resolution.

dwijnand:
Provided dependencies are provided by something for a runtime, but they're non-optional dependencies outside of runtime.  MiMa is studying the bytecode of a jar (or a dir of classes), bytecode that can reference bytecode in another dependency.  As you say "Provided dependencies of transitive dependencies are always ignored during resolution" which is a problem for me/MiMa.

What MiMa does today, for reasons I haven't discovered yet, is it provides a stub "SytheticClassInfo" when it doesn't find a class on the classpath, which is a class with the given (looked up) name, extends Object and that's it, which means false "missing methods" (etc) errors... 😕

Does that make more sense?

olafurpg:
I just hit on a use-cases where it would be useful to be able to optionally include jars of provided transitive dependencies. My use-case is to compile source code from the `-sources.jar` of a library. Currently, I get compile errors from unresolved symbols that origin in provided dependencies.

For example, here is an error compiling the byte-buddy sources

```
/var/folders/mb/88qbygqj169byvv_68bw4fn80000gn/T/semanticdb-javac5269371498875463369/net/bytebuddy/ClassFileVersion.java:18: error: package edu.umd.cs.findbugs.annotations does not exist
import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
                                      ^
/var/folders/mb/88qbygqj169byvv_68bw4fn80000gn/T/semanticdb-javac5269371498875463369/net/bytebuddy/ClassFileVersion.java:291: error: cannot find symbol
    @SuppressFBWarnings(value = "REC_CATCH_EXCEPTION", justification = "Exception should not be rethrown but trigger a fallback")
     ^
  symbol:   class SuppressFBWarnings
  location: class ClassFileVersion
[...]

PR #2891, opened 2023-11-23
Add option to resolve transitive provided deps

adpi2:
Fix https://github.com/coursier/coursier/issues/1846

Implemented solution described in https://github.com/coursier/coursier/issues/1846#issuecomment-775526310


adpi2:
> The change here isn't really about making them transitive, it's about taking them into account in the first place (and dependencies, provided or not, are assumed to be transitive)

Actually it is really about transitive dependencies. If I depend directly on `foo:provided`, Coursier should and does resolve it but if I depend transitively on it, it does not resolve it. Which is correct based on the [Maven spec](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-scope).

The option changes that behavior to also resolve transitive provided dependencies.

So `keepTransitiveProvidedDependencies` would be more precise, but quite long. I am okay with `keepProvidedDependencies`.

alexarchambault:
Yeah, but coursier doesn't really know about "provided" input dependencies, they're just simple input dependencies for it. It's only the build tool that marks them as provided when it publishes them, so that they're later ignored. Overall, before this PR, coursier only ignores provided dependencies.

alexarchambault:
Just pushed a commit that makes `ResolutionParams.keepProvidedDependencies` an option. There's [some logic in the tests](https://github.com/coursier/coursier/blob/00b055fcca6d2fad694d14ea91a1c3bc23409e6e/modules/coursier/shared/src/test/scala/coursier/TestHelpers.scala#L72-L83) that makes the resolution hashes remain unchanged when adding an optional field to `ResolutionParams` (and having this param being a `Boolean` or an `Option[Boolean]` is kind of even…)

alexarchambault:
@adpi2 Do you need a release with that sooner rather than later? Or can this wait a bit for more changes to accumulate?

adpi2:
> @adpi2 Do you need a release with that sooner rather than later? Or can this wait a bit for more changes to accumulate?

It can wait a bit but thanks for asking!

lefou:
Hi, I just stumbled over this PR when I was in search for some way to check a set of dependencies for consistency, considering their transitive dependencies with `provided`-scope. 

My use case is to detect, whether a Mill plugin is compatible to the current Mill version. Mill plugins typically declare their Mill-dependencies with `provided`-scope.

In detail, I want to detect, that a plugin which was build against a newer version of the API (e.g. 0.11.6) is compatible to the currently used API (e.g. 0.11.5). The idea was to add some version range for the currently running Mill API, e.g. `[0.11.0,0.11.5]` and also add the plugins. In case a plugin was build against a newer version, the version reconciliation process as outlined in https://get-coursier.io/docs/other-version-handling should fail due to the mismatch of the version range and the actual required API version. But the whole idea is only working if coursier also takes the `provided`-scoped dependencies into account.

So here are my two questions:

1. Is the idea from above already possible with this PR? If not, what else is missing? 

2. Can we get a new release including this PR?

Thank you!

Issue #1832, opened 2020-08-24
Coursier fails to resolve all maven dependencies


kamilkloch:
Coursier `v2.0.0-RC6-24`.
```
> coursier resolve -t com.google.inject:guice:3.0
  Result:
└─ com.google.inject:guice:3.0
   ├─ aopalliance:aopalliance:1.0
   └─ javax.inject:javax.inject:1
```
Direct compile dependency ` org.sonatype.sisu.inject:cglib:2.2.1-v20090111` is missing: https://mvnrepository.com/artifact/com.google.inject/guice/3.0

alexarchambault:
@kamilkloch That seems to be expected. The [POM of guice](https://repo1.maven.org/maven2/com/google/inject/guice/3.0/guice-3.0.pom) has a `guice.with.jarjar` profile, which gets automatically activated (the `guice.with.jarjar` is not `false`, as it's undefined, which activates the `guice.with.jarjar` profile). This profile marks the cglib as optional.

Manually setting the property to `false` disables the profile, and pulls cglib:
```text
$ cs resolve com.google.inject:guice:3.0 --pom-property guice.with.jarjar=false
aopalliance:aopalliance:1.0:default
asm:asm:3.1:default
com.google.inject:guice:3.0:default
javax.inject:javax.inject:1:default
org.sonatype.sisu.inject:cglib:2.2.1-v20090111:default
```

Maven behaves the same: with a simple POM like
```xml
<?xml version='1.0' encoding='UTF-8'?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0">
  <modelVersion>4.0.0</modelVersion>

  <groupId>test</groupId>
  <artifactId>main</artifactId>
  <version>0.1.0</version>

  <organization>
    <name>test</name>
    <url>https://github.com</url>
  </organization>
  <name>main</name>

  <dependencies>
    <dependency>
      <groupId>com.google.inject</groupId>
      <artifactId>guice</artifactId>
      <version>3.0</version>
    </dependency>
  </dependencies>

</project>
```
we get
```text
$ mvn dependency:build-classpath
…
[INFO] Dependencies classpath:
[...]

PR #2929, opened 2024-02-15
Warn about use of  `~/.m2/repository`

ghostbuster91:
We have just run ourselves into a weird issue with coursier where it would fail with artifact not found error. After closer inspection it turned out that it was because of use of ~/.m2/repository as a coursier repository.

I think that it would be great to add a warning to the documentation about this.

This was also mentioned here https://github.com/coursier/coursier/issues/1832#issuecomment-1357764540

This directory is maven internal files-system cache and afaik its layout can change without prior notice. It also can contain POMs without corresponding jars (e.g. when a transitive dependency was evicted due to a version conflict) which confuses coursier.

See problem reproduction and more details in https://github.com/ghostbuster91/demos/tree/coursier-not-found

### DIFF
--- a/doc/docs/other-repositories.md
+++ b/doc/docs/other-repositories.md
@@ -29,6 +29,8 @@ $ cs fetch --no-default \
   org.scalameta:scalafmt-cli_2.12:2.0.0-RC4+29-f2154330-SNAPSHOT
 ```
 
+⚠️ _Avoid using `~/.m2/repository` as a repository as it is maven's internal file-system cache._ 
+
 ## Ivy Repositories
 
 It is possible to use a custom Ivy repository and pattern by concatenating them:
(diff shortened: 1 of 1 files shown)


lefou (commented):
I saw these kind of warnings earlier in the source code and must say I don't like them as they are. That means, they are not actionable, so the used has no clue _why_ they are risky, so they can't decide _what_ risks they face and whether they want to deal with it or not.

The option to integrate with a local Maven repository is often the only choice to use another build tool than Maven. Therefore, this is a good feature, although it comes with some issues.

So we either add some links, e.g. the the issue you mentioned in the PR comment, or some text outlining the kind of issue a user should anticipate when using it. Maybe, also some workaround or fix actions.

ghostbuster91:
I see your point. I expanded that section a bit, let me know if this is enough.

lefou (approved):
Looks good to me, thanks.

tgodzik (approved):
Thanks!

Issue #837, opened 2018-04-06
Repository order does not seem to apply between local and remote


wisechengyi:
```
$ rm -rf /Users/me/Library/Caches/Coursier/ && \
./run fetch -t junit:junit:4.12 -r https://repo1.maven.org/maven2 -r file:///Users/me/.m2/repository --no-default
  Result:
└─ junit:junit:4.12
   └─ org.hamcrest:hamcrest-core:1.3
/Users/me/.m2/repository/junit/junit/4.12/junit-4.12.jar
/Users/me/.m2/repository/org/hamcrest/hamcrest-core/1.3/hamcrest-core-1.3.jar
```
In this case we want to check from `https://repo1.maven.org/maven2` first then local.

I think it's important to enforce the order because in most cases, we want developers to resolve against a central system, then fallback to their local maven repo used to publish testing jars. If the local repo took priority that could pollute the resolve, since it is the least trusted place.

alexarchambault:
This is expected, and originates from the [cache "policies"](https://github.com/coursier/coursier/blob/595f7d7d48c0c38bded88531d1fe1cdeb3ccc07c/cache/jvm/src/main/scala/coursier/CachePolicy.scala) or ["modes"](https://github.com/coursier/coursier/blob/595f7d7d48c0c38bded88531d1fe1cdeb3ccc07c/cli/src/main/scala-2.12/coursier/cli/options/CommonOptions.scala#L13).

The modes are tried successively. For each mode, the repositories are tried successively.

By [default](https://github.com/coursier/coursier/blob/595f7d7d48c0c38bded88531d1fe1cdeb3ccc07c/cache/jvm/src/main/scala/coursier/CachePolicy.scala#L68)), the first policy looks at what's in the cache, only trying to do fetch / update the changing things (snapshots…) that are already in cache. The second policy only looks at what's already in cache (this one may be redundant with the first one…). The third one tries to actually fetch things.

So if we have 3 repositories say, and a dependency is already in cache for the third one, it will be peaked in cache from the third repo (by the first two policies). Even if the dependency was actually available in the first or second repositories.

That trades off some reproducibility (we may get different jars depending on the previous cache state) for less IO overhead (if something is in cache for any of the repositories, no network IO will happen). It's unlikely a non-snapshot version of a JAR is different on different repositories, so I guess it's worth it…

[...]

PR #2930, opened 2024-02-15
Add notes about repository resolution order

ghostbuster91:
It is not obvious that when using local repository coursier will use it first regardless of the specified order, so I think that it is a good idea to document it.

This is mostly based on this comment https://github.com/coursier/coursier/issues/837#issuecomment-379746716

### DIFF
--- a/doc/docs/other-repositories.md
+++ b/doc/docs/other-repositories.md
@@ -36,3 +36,33 @@ It is possible to use a custom Ivy repository and pattern by concatenating them:
 ```
 $ cs fetch --repository  "https://corporate.com/ivy-repo/[org]/[module]/[baseRev](-[folderItegRev])/[module]-[baseRev](-[fileItegRev])(-[classifier]).[ext]" sh.almond:scala-kernel_2.12.8:0.2.2
 ```
+
+## Repository resolution order
+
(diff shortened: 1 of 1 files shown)


lefou (approved):

doc/docs/other-repositories.md
```diff
@@ -36,3 +36,33 @@ It is possible to use a custom Ivy repository and pattern by concatenating them:
...
+Modes are tried successively and for each mode, repositories themselves are tried successively.
+
+By [default](https://github.com/coursier/coursier/blob/595f7d7d48c0c38bded88531d1fe1cdeb3ccc07c/cache/jvm/src/main/scala/coursier/CachePolicy.scala#L68), 
+the first policy looks at what's in the cache, only trying to fetch / update the changing things (snapshots…) that are already in cache. 
+
+The second policy only looks at what's already in cache (this one may be redundant with the first one…). 
+
+The third one tries to actually fetch things.
+
+If we have 3 repositories and a dependency is already in cache for the third one, it will be peaked in cache from the third repo (by the first two policies).
```
tgodzik, line 53:
```suggestion
If we have 3 repositories and a dependency is already in cache for the third one, it will be picked in cache from the third repo (by the first two policies). 
```
? I am not sure I understand the sentence :thinking:

ghostbuster91: Thanks, good catch:+1:

tgodzik (approved):
Thanks!

PR #2957, opened 2024-04-08
Fix unquoted variable for java path in boostrap script

zaneduffield:
The unquoted `$JAVA_CMD` breaks when `$JAVA_HOME` contains a space (e.g. in `C:\Program Files` on Windows).


tgodzik (approved):
Makes sense, thanks!

PR #2973, opened 2024-05-02
Update the sbt-coursier version on site to actual

danicheg:
Hey there. I observed a minor issue: the version of `sbt-coursier` on site is outdated and generally not automatically updated by Scala Steward / dependabot when new versions arrive. In this PR, we update the version and add a caution on where to pick the latest version on site.

### DIFF
--- a/doc/docs/sbt-coursier.md
+++ b/doc/docs/sbt-coursier.md
@@ -41,6 +41,8 @@ not handle publishing for now.
 
 ## Setup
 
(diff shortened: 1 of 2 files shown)


tgodzik (approved):
Thanks!

tgodzik:
I think rebasing on top of main should fix the CI issue.

danicheg:
Yeah, definitely. Or maybe we can move on without that, since I changed only site related things?

tgodzik:
sure, should be fine!

Issue #2944, opened 2024-03-06
Start building Apple Silicon native launcher on coursier main repo


keynmol:
Github now has built-in apple silicon support - it's available by setting os to `macos-14` (`macos-12` will remain forever on x86).

Doing so will allow getting rid of VirtusLab/coursier-m1 repository altogether, though it might remain for redirects.

alexarchambault:
Fixed by #3148

PR #2974, opened 2024-05-02
Pin the macos-12 for generate-native-launchers job in CI

danicheg:
It's an attempt to make CI green.


danicheg:
Ah, CI requires a bit of another fix. Will rework this.

danicheg:
In light of https://github.com/coursier/coursier/issues/2944, we perhaps have nothing but to pin the `macos-12` version in CI for now.

alexarchambault (approved):
Thanks!

PR #2975, opened 2024-05-02
support mix of launcherType in version overrides

bishabosha:
needed for https://github.com/scala/scala3/issues/20098 and https://github.com/VirtusLab/scala-cli/issues/2838

enable a mixture of launcherType in version overrides, this allows to have a prebuilt scala launcher for 3.5.0+, and then bootstrap launcher for anything below to preserve backwards compatibility.

here is the proposed new `scala.json` app descriptor:

```json
{
  "repositories": [
    "central"
  ],
  "dependencies": [
    "org.scala-lang:scala3-compiler_3:latest.stable"
  ],
  "name": "scala",
  "launcherType": "prebuilt",
  "prebuilt": "zip+https://github.com/scala/scala3/releases/download/${version}/scala3-${version}.zip!scala3-${version}/bin/scala",
  "versionOverrides": [
    {
      "versionRange": "[3.1.0,3.4.max]",
      "launcherType": "bootstrap",
      "properties": {
        "scala.usejavacp": "true"
      },
      "mainClass": "dotty.tools.MainGenericRunner"
    },
    {
      "versionRange": "[3.0.0,3.0.max]",
      "launcherType": "bootstrap",
      "properties": {
        "scala.usejavacp": "true"
      },
      "mainClass": "dotty.tools.repl.Main"
    },
    {
      "versionRange": "(,2.max]",
      "launcherType": "bootstrap",
      "properties": {
        "scala.usejavacp": "true"
      },
      "dependencies": [
        "org.scala-lang:scala-compiler:latest.stable"
      ],
      "mainClass": "scala.tools.nsc.MainGenericRunner"
    }
  ]
}
```

we would actually have to update this app descriptor again when we want a true native launcher built by graalvm - but version overrides support that already.

cc/ @adpi2 @tgodzik @kasiaMarek @Gedochao @hamzaremmal


bishabosha:
Everyone involved, I added a field to AppDescriptor so I guess we should have a new minor release I put 2.2.0 in the since annotation, but is that right?

adpi2 (approved):
LGTM

bishabosha:
CI fails due to https://github.com/coursier/coursier/pull/2974

alexarchambault:
> Everyone involved, I added a field to AppDescriptor so I guess we should have a new minor release I put 2.2.0 in the since annotation, but is that right?

Precisely, the `@since` annotation allows not to break binary compatibility upon field addition

modules/install/src/test/scala/coursier/install/InstallTests.scala
```diff
@@ -835,12 +839,14 @@ object InstallTests extends TestSuite {
             "scala.usejavacp=true",
             "scala-compiler.version=2.12.8"
           )
+
         testRun(scala2CompilerJars, scala2Properties)
 
         val updated = installDir0.createOrUpdate(appInfo0)
         assert(updated.exists(identity))
 
-        testRun(scala3CompilerJars, scala3Properties)
```
alexarchambault, line 843: Can't we keep this check (`testRun`), alongside the new `testOutput`?

bishabosha: so I removed this one because `testRun` is specifically looking for files inside a bootstrapped launcher jar, but with the change in the test it is now a prebuilt launcher, so these entries don't exist

bishabosha: maybe I can do a check that the file exists and matches some shape?

bishabosha: I added a test that searches for the path to the launcher in the script file

modules/install/src/test/scala/coursier/install/InstallTests.scala
```diff
@@ -791,6 +797,13 @@ object InstallTests extends TestSuite {
         val launcher = installDir0.actualDest(id)
         assert(Files.isRegularFile(launcher))
 
+        // for a prebuilt launcher, we expect the path of the launcher to appear somewhere in the script
+        val scala3path = {
+          val original = "github.com/scala/scala3/releases/download/3.3.3/scala3-3.3.3.zip/scala3-3.3.3/bin/scala"
+          if (currentOs == "windows") original.replace('/', '\\')
```
bishabosha, line 803: I use `currentOs` here because if ran from e.g. macOS then the Mac-formatted path will actually be in the launcher.

tgodzik (approved):
LGTM from me, since it's needed for Scala CLI as Scala I am going to merge it and work on a release. Hopefully, this is not an issue @alexarchambault ?

Thanks @bishabosha !

Issue #2985, opened 2024-05-23
Review and replace use of git submodules


lefou:
Currently used submodules:

* [ ] modules/directories
* [ ] modules/tests/handmade-metadata
* [ ] modules/tests/metadata
* [ ] modules/windows-ansi

Find out, which submodule is used for what. Depending on their usage, replace with either:

1. Use some proper dependency
2. Checkout in a build step (e.g. a custom Mill task) and use as resources or sources

lefou:
@alexarchambault Can you please comment, if there are any reasons to *not* vendor the test metadata into this repository? I'm speaking about [test-metadata](https://github.com/coursier/test-metadata) and [handmade-metadata](https://github.com/coursier/handmade-metadata).

lefou:
In these three pull requests, I removed all git submodules by applying different strategies (1-3):

* https://github.com/coursier/coursier/pull/2986 - Strategy 1: Use a regular library dependency, fetch the source-artifact and use it in the build
* https://github.com/coursier/coursier/pull/2987 - Strategy 2: just fetch the sources and resources from the original repository and use then in the build
* https://github.com/coursier/coursier/pull/2988 - Strategy 3: vendored the files into this repository

After the last of these PRs is merged, the developer documentation needs a cleanup, since there are no git submodules left and the build workflow gets easier.

Please review!

lefou:
Since the question came up, why git submodules are a problem: Creating pull requests by users that don't have write permissions to all referenced git repositories, but require modifications to files in these submodules are simply impossible. Especially, when you aim for WIP/draft PRs.

And since coursier code base is complicated, most potential contributors that are brave enough to still insist to tackle it, will be going to write and adapt tests first. But part of the test data is located in submodules, so we can be almost sure that for every feature PR we got in the past there were many more attempts who gave up.

alexarchambault:
> @alexarchambault Can you please comment, if there are any reasons to not vendor the test metadata into this repository? I'm speaking about [test-metadata](https://github.com/coursier/test-metadata) and [handmade-metadata](https://github.com/coursier/handmade-metadata).

IIRC, [test-metadata](https://github.com/coursier/test-metadata) is managed via a submodule because of its size. I'd rather keep this one this way, I don't see a more handy way to manage it. If it's being kept in a separate repo, downloading it from the build won't help much contributors.

About [handmade-metadata](https://github.com/coursier/handmade-metadata), this one is small enough to be vendored, yes.

[...]

PR #2986, opened 2024-05-23
Replace git submodule with `windows-ansi-ps` source dependency

lefou:
Instead of using a git submodule, we just use a normal library dependency on the source artifact and use it in the build.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -85,15 +85,12 @@ object paths extends JavaModule {
     Deps.jniUtils
   )
 }
-object `windows-ansi` extends Module {
-  object ps extends JavaModule
-}
 
 object `custom-protocol-for-test` extends CsModule {
   def scalaVersion = ScalaVersions.scala213
 }
 
-object `bootstrap-launcher` extends BootstrapLauncher { self =>
+object `bootstrap-launcher` extends BootstrapLauncher with CoursierModule { self =>
   def proxySources = T.sources {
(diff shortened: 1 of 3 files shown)


tgodzik (approved):
Looks good, I just wonder why are submodules a problem? I am also not a big fan, but I am not aware on any problems other than they are more complicated to use.

tgodzik:
Thanks for explaining! I think it's good to merge it then.

build.sc
```diff
@@ -109,12 +106,25 @@ object `bootstrap-launcher` extends BootstrapLauncher { self =>
       }
     Seq(PathRef(dest))
   }
-  def sources = T.sources {
+  def windowsAnsiPsSources = T {
```
alexarchambault, line 109: It might be possible to use standard dependencies here. We just have to make sure that they end up in the various assemblies we generate around [here](https://github.com/coursier/coursier/blob/13b6faa5936b838d73ba773ba90a23ec5574997f/project/modules/bootstrap-launcher0.sc#L85-L171) (I think they should end up there out-of-the-box… The use of source dependencies is, among others, a remnant of me not wanting to take the time to set things up in sbt, when coursier was built with sbt. This is not an issue anymore, since the switch to Mill.)

alexarchambault (commented):
Thanks a lot for this, @lefou! (and @tgodzik for the review)

PR #2987, opened 2024-05-23
Replace git submodule `directories` with Mill based download/unzip

lefou:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "tests/metadata"]
 	path = modules/tests/metadata
 	url = https://github.com/coursier/test-metadata.git
-[submodule "modules/directories"]
-	path = modules/directories
-	url = https://github.com/coursier/directories-jvm.git
 [submodule "modules/tests/handmade-metadata"]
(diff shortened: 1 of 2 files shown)


alexarchambault (changes requested):
I'm not sure checking out things manually from the build help much here. Given that if users want to change things in the sub-repo, they'll still have to send a PR to it too. I'd rather publish coursier/directories-jvm to Maven Central, then use it as a standard dependency.

lefou:
> Given that if users want to change things in the sub-repo, they'll still have to send a PR to it too. 

Or they can be lazy and just place/override the changed file, temporarily.

lefou:
I'm fine with both. IMHO this PR improves the git submodules aspect. But a published `directories-jvm` dependency would be even better, but is out of my control.

* #2989

alexarchambault:
FYI, I'm trying to setup publishing for coursier/directories-jvm [here](https://github.com/coursier/directories-jvm/pull/1)

lefou:
Can't you just upstream your changes (which seem to be only one commit) and use upstream instead?

PR #2988, opened 2024-05-24
Vendor test metadata (removed git submodules)

lefou:
- Remove git submodules for metadata and handmade-metadata
- Add metadata files previously located in a git submodule
- Add handmade-metadata files previously located in a git submodule

I plainly added all test metadata files to the repository, since this is the easiest way to work with them (and edit them). I had no evidence, that these data where used outside the test workflows in coursier.

If having all test metadata in isolation is a requirement, we could also create an orphan branch for the metadata, and create a merge commit initially and on each change. Possible, but it has less potential to ease the DX.


tgodzik:
I am not sure why it was done this way. @alexarchambault can confirm if it's ok to change.

alexarchambault:
To be honest, I'd rather keep this one apart from the main repo, given the amount of clutter it brings
Capture d’écran 2024-05-24 à 15 53 08

It seems I can't even review this PR, it makes the GitHub freeze 😬

alexarchambault (changes requested):
See my comment above

lefou:
@alexarchambault Can you please outline how a new contributor is support to submit a draft PR which involves changed test data, that is supposed to possibly change again while working on the PR? Creating a draft PR is essential to get early feedback from CI and maintainers.

alexarchambault:
I can document that, this mainly involves (temporarily) changing `https://github.com/coursier/test-metadata` in `.gitmodules` to the address of the contributor's fork, and then making the submodule point to a commit in the fork.

alexarchambault:
User should also send a PR to the test-metadata repo at the same time. Prior to merging the PR against the main repo, we merge the one against test-metadata, and wait for users to revert the address change in `.gitmodules` prior to merging.

alexarchambault:
Several contributors did that in the past…

lefou:
Excuse my language, but I think that workflow sucks. 

If we could at least stay in the same repo, on an orphan branch, to avoid all the hassle of (multiple) pull requests just to get to the first CI result. I don't know if git `submodules` can do that. Maybe git `subtree` works that way. Or we check out an orphan branch with Mill, so it's a simple line change in the `build.sc` while working on a PR.

PR #3068, opened 2024-09-09
Update Scala.JS to 1.16.0

alexarchambault:


alexarchambault:
Seems required for https://github.com/coursier/coursier/pull/3048

PR #3069, opened 2024-09-09
Retry on FileNotFoundException when accessing "structure lock" file

alexarchambault:


alexarchambault:
See https://github.com/com-lihaoyi/mill/issues/3465#issuecomment-2338638113 for context

PR #3086, opened 2024-10-02
Fix #3085: Ignore attributes for conflict resolution

### LINKED ISSUE #3085: Eviction for transitive sbt plugin dependencies not working
cross-posting:
- https://github.com/sbt/sbt/issues/7710

adpi2:


modules/core/shared/src/main/scala/coursier/core/ResolutionProcess.scala
```diff
@@ -417,7 +417,7 @@ object ResolutionProcess {
 
     def uniqueModules(modVers: Seq[(Module, String)]): LazyList[Seq[(Module, String)]] = {
 
-      val res = modVers.groupBy(_._1).toSeq.map(_._2).map {
+      val res = modVers.groupBy(_._1.withAttributes(Map.empty)).toSeq.map(_._2).map {
```
eed3si9n, line 420: In theory, there could be multiple versions of sbt within the same Scala version, so some attributes are part of the resolution address, while others are informational?

adpi2: Yes, in theory, `sbtVersion` and `scalaVersion` should both be part of the resolution address. But in practice we should never have two versions of the same artifacts with different `sbtVersion` or `scalaVersion`. The only 2 meaningful `sbtVersion`s are 0.13 and 1.0, which are built on incompatible Scala 2.10 and 2.12. It makes no sense to resolve them in the same resolution job.

So I feel like it is useless to special case `scalaVersion` and `sbtVersion` attributes. I prefer to keep it simple and ignore all attribues.

lefou: And sbt 2.x will encode the binary platform in the artifact ID, as all other projects do nowadays.

adpi2:
Superseded by #3088

PR #3094, opened 2024-10-11
Remove argonaut-shapeless dependency from jvm module

alexarchambault:
Seems it was unused.

I think dropping dependencies is supposed to break binary compatibility though 🤔 (if a library compiled with coursier `2.1.14` relies on shapeless via it, and users bump coursier to `2.1.15` at runtime, shapeless is likely to be missing)

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -525,7 +525,6 @@ trait Jvm extends CrossSbtModule with CsModule
     Deps.svm
   )
   def ivyDeps = super.ivyDeps() ++ Agg(
-    Deps.argonautShapeless,
     Deps.jsoniterCore
   )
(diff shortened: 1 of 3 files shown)


lefou:
> I think dropping dependencies is supposed to break binary compatibility though 🤔 (if a library compiled with coursier `2.1.1`

This isn't a binary compatibility isssue but a incomplete declared dependencies issue. Dropping unused dependencies should always be ok, unless you claim to provide some kind of framework that provides all that is needed, which shouldn't be the case for coursier as a library.

tgodzik (approved):
I think that's fine. Unless anyone depends on a library transitively, but that's their own problem really.

lefou (approved):

PR #3095, opened 2024-10-11
Add `Deps.collectionCompat` to `coursier-core`

Friendseeker:
### Issue

https://github.com/coursier/coursier/blob/880a2220015cd9aefd436977af76ca4c841c1e8e/modules/core/shared/src/main/scala/coursier/core/Resolution.scala#L8

`coursier-core` relies on `collectionCompat`, yet the dependency is not declared. As a result user of `coursier-core` has to manually add `collectionCompat` to their build config.

### Fix

Add `Deps.collectionCompat` to `coursier-core`

### DIFF
--- a/project/modules/core0.sc
+++ b/project/modules/core0.sc
@@ -11,7 +11,8 @@ trait Core extends CsModule with CsCrossJvmJsModule with CoursierPublishModule {
     Deps.dataClass
   )
   def ivyDeps = super.ivyDeps() ++ Agg(
-    Deps.fastParse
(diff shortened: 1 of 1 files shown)


Friendseeker:
The reason for this is to investigate the failure that is happening with `sbt` [build](https://github.com/sbt/sbt/pull/7759#discussion_r1797412428), but now looking more closely this should be unnecessary and this is not the root cause for failure... `coursier-core` depends on  `coursier-util`  which depends on `collectionCompat`.

Issue #1390, opened 2019-10-01
Support BOMs/dependencyManagement


mandrean:
Correct me if I'm wrong, but to me it seems that the Coursier CLI doesn't support fetching/resolving the `dependencyManagement` dependencies in BOMs?

For example: https://github.com/FasterXML/jackson-bom/blob/master/pom.xml#L75

I've tried all kinds of variants
```
$ coursier resolve com.fasterxml.jackson:jackson-bom:2.9.10 -t
$ coursier fetch com.fasterxml.jackson:jackson-bom:2.9.10 -j dep-tree.json && cat dep-tree.json
```

and more.

Also see https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Management

>The dependency management section is a mechanism for centralizing dependency information. When you have a set of projects that inherits a common parent it's possible to put all information about the dependency in the common POM and have simpler references to the artifacts in the child POMs.

Commonly used in large companies with a lot of teams developing various services in order to align them on external dependency versions!

alexarchambault:
You mean, you'd like the CLI to fetch all the dependencies of the `dependencyManagement` section? It's not something that's supported as is. What would be the use case for that?

Resolution does take into account parent POMs / import scope during resolution though.

mandrean:
Partially, the use-case is: https://github.com/bazelbuild/rules_jvm_external/issues/229

But also more broadly speaking:

Many companies (including Spotify, where I'm at) use BOMs to centrally manage versions of various company-wide internal dependencies. Also some open-source projects have support for that, like: https://github.com/FasterXML/jackson-bom

As it is today, if you run `coursier fetch` on some artifact that happens to rely on a BOM to resolve some of its dependencies, then `coursier`, would make an incomplete job at downloading all the transitive dependencies  (`mvn` can do this, however).

Same if you use `--json-output-file` - the dependency resolution list there is in fact incomplete!

kohsuke:
@alexarchambault I'm worried that the point of this feature request was misunderstood, so let me try to clarify.

What @mandrean is asking isn't so much "fetching all the dependencies of the `dependencyManagement` section.

The request is that the resolution should take `dependencyManagement` into account, in particular the dependency management included via the import scope, aka the BOM pattern. And that the equivalent capability be exposed via CLI:
* command line option to contribute an entry to `ResolutionParams.addForceVersion`, which IIUC is the equivalent of `dependencyManagement` in `pom.xml`
[...]

PR #3097, opened 2024-10-16
Add better support for dependency management (aka BOM)

alexarchambault:
## What these changes are about

This PR changes things related to the so called Maven "BOMs" or "dependency management" support. It changes the way versions are picked for many dependencies of various JVM ecosystems (Android, Spark, Spring, …), hence the "sensitiveness" of these changes.

## What BOMs are

BOMs are basically POM files listing versions to be used for dependencies. They look like [this](https://repo1.maven.org/maven2/io/quarkus/quarkus-bom/3.9.5/quarkus-bom-3.9.5.pom): this POM contains a `dependencyManagement` section, listing many dependencies. These dependencies' version and sometimes exclusions matter to us here.

A standard Maven dependency can pull a BOM either by declaring it as a "parent", like at the top of [this POM](https://repo1.maven.org/maven2/org/apache/spark/spark-core_2.13/3.5.3/spark-core_2.13-3.5.3.pom), or by declaring an "import" dependency in its `dependencies` or `dependencyManagement` sections, like in [this POM](https://repo1.maven.org/maven2/io/quarkus/quarkus-build-parent/3.9.5/quarkus-build-parent-3.9.5.pom). Note that parents POMs can themselves have parents and "import" dependencies, whose values are taken into account too. Note also that POMs can also have a `<dependencyManagement>` section, acting as a BOM too, like [this POM](https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.15.2/jackson-databind-2.15.2.pom) - BOM values don't always come from other POM files.

When added to a Maven project, Maven BOMs are basically [expected to be](https://reflectoring.io/maven-bom/) dependency override lists. If project A has B:1.2 in its BOM, it is expected either: not to pull B (if neither A or its transitive dependencies depend on B), or to pull B version `1.2` (and no other version). No matter what version of B a transitive dependency of A depends on, if pulled via A, B should have version `1.2`, because of A's BOM.

Also, it's common practice in Maven projects having a BOM not to put versions in dependencies in their `dependencies` section. These versions are supposed to be filled in with values from the BOM. We can see that clearly in [this POM](https://repo1.maven.org/maven2/io/quarkus/quarkus-rest/3.9.5/quarkus-rest-3.9.5.pom).

## BOM support in coursier

Up to now, coursier used BOM values for the latter case (filling up empty versions of direct dependencies of the dependency that pulls the BOM), not the former (dependency overrides, applying to transitive dependencies too). Early checks of mine (around… 2015 😅) didn't display the former behavior, so I assumed all those BOM values where only meant for the latter case, when one pulls Maven dependencies.

This PR changes that: it uses the BOM values as "dependency overrides" too, applying to transitive dependencies.

## Impact of those changes

One can see in the diff of this PR how many resolutions are impacted (or not impacted…). Many resolutions' results are hard-coded under `modules/tests/shared/src/test/resources/resolutions`.  We can see in the diff here that Spark or Quarkus dependencies are impacted, among others. Overall, these changes are consistent with the new BOM values handling (all individual values I looked at in detail were fine at least).

## Should these changes be enabled by default right now?

The new way of computing versions gives more consistent dependencies to me, more in line with what Maven users use when building their projects. These changes ought to be welcomed, yet they might surprise some users.

I'm thinking these could be enabled by default, but advertized loudly in release notes, and users need to be able to disable them if they want to.

## Dependency overrides handling in build tools

BOMs give a "first class citizen" status to "dependency overrides". BOMs are dependency overrides that can be published, and pulled by downstream users.

In sbt for example, the values of `dependencyOverrides` are equivalent to BOM values. As a consequence, these can be published. These can go in a `<dependencyManagement>` section of a module POM file, like in [this POM](https://repo1.maven.org/maven2/com/fasterxml/jackson/core/jackson-databind/2.15.2/jackson-databind-2.15.2.pom). Then, when users pull that sbt-published module, coursier applies the dependency overrides from the `<dependencyManagement>` section to transitive dependencies pulled by the module.

AFAIK, Mill doesn't have an equivalent target.

A simple setting like `dependencyOverrides` in sbt or an equivalent target in Mill only correspond to `dependencyManagement` sections in library POMs. They don't allow to publish shared dependency overrides via "parent POMs" and "dependency imports", nor to pull such overrides from Maven Central. Gradle apparently [allows such uses](https://docs.gradle.org/current/userguide/platforms.html).


alexarchambault:
Given this impacts the way versions gets picked during resolution, I'm going to write more in length about these changes, and ask some people who know about these topics too their opinion

alexarchambault:
FYI, if anyone's looking at this, I'm pretty happy with these changes now (the way test data are changed is consistent overall, and these look fine in the final commit). I should write about it soon here.

alexarchambault:
cc @eed3si9n who [dived a quite a bit in dependency resolver intricacies](https://eed3si9n.com/dependency-resolver-semantics/) and @lefou who manifested interest for BOM support in https://github.com/coursier/coursier/issues/1390. Also cc-ing @lihaoyi and @adpi2 given this should impact both Mill and sbt, and @jin for [rules_jvm_external](https://github.com/bazel-contrib/rules_jvm_external)

lihaoyi:
I haven't reviewed the code, but the PR description looks like exactly what we'd want, since it makes Coursier resolutions behave the same as Maven or Gradle or other tools

alexarchambault:
> AFAIK, Mill doesn't have an equivalent target.

I'm probably going to have a look at this next: adding an equivalent to `dependencyOverrides` in Mill, and make sure its values are added to published POM files (and checking that these overrides are taken into account when coursier pulls these)

alexarchambault:
One note: I didn't consider yet the impact of this on performance. This involves passing around and updating many maps of dependency overrides (`Dependency#dependencyOverrides`). This looks a bit like the handling of exclusions or dependency sets. Both benefitted from optimizations over the initial naive approaches (https://github.com/coursier/coursier/pull/2442, https://github.com/coursier/coursier/pull/1257, https://github.com/coursier/coursier/pull/2420)

lefou:
Thank you @alexarchambault for working on it. Here is Mill's tracker issue:

* https://github.com/com-lihaoyi/mill/issues/1975

> AFAIK, Mill doesn't have an equivalent target.

Mill has the `CoursierModule.mapDependencies` task, returning a function of type `Dependency => Dependency`, which can be used to map dependencies. I used this function often to apply something like Maven `dependencyManagement` in my projects when needed. It's not published in the `pom` though. I'd love to deprecate it in favor of a more explicit dependency management, which is not a function.

Additionally, in Mill, dependencies can set the `force` flag with `forceVersion`. It's internally handled by the `mapDependencies` task. But I find it's less usable and more hidden, as well when writing the dependencies as also when reading projects. I'd like to deprecate it as well. We also explain all those shortcoming of `Dep.forceVersion` in our documentation, so deprecating it in favor of a better and more consistent solution will most likely be welcomed by our users and should become the default behavior.

What needs to be done in Mill:

* Decide, how we do declare managed dependencies? We should add a new task to `CoursierModule`. We need to agree on a good name. The type should be `T[Seq[Dep]]`.

* Decide, how we do import a BOM dependencies in `import`-scope? In Mill dependencies don't have scopes, but we manage multiple scope-like dependency sets: `ivyDeps` ~= `compile`-scope, `runIvyDeps` ~= `runtime`-scope, `compileIvyDeps` ~= `provided`-scope. (To import BOMs via `ivyDeps`, we need to find out, if Maven makes a distinction between these scopes and if there are any situations, where mixing `compile` and `import` scope for BOM management is problematic.) We better add a new task to import those BOMs. The task name might match the other new task name but add a suffix like `Includes` or `Imports`. The type is also `T[Seq[Dep]]`.

* How can we produce BOM artifacts? I think this should already work by changing the `PublishModule.pomPackagingType` to `Pom`.

eed3si9n:
Thanks @alexarchambault for this work! I'd echo @lihaoyi on the point that it's great to have Coursier resolution in line with Maven, or giving users the capability to do so.

> In sbt for example, the values of `dependencyOverrides` are equivalent to BOM values. As a consequence, these can be published.

I think this is mostly true, but there is a small details of Maven dependency quirk, which we might have to be aware: I think Maven prioritizes direct dependency over dependencyManagement.

### example of dependency vs dependencyManagement

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.example</groupId>
  <artifactId>hi</artifactId>
  <version>1.0.0</version>
  <packaging>jar</packaging>

  <dependencyManagement>
    <dependencies>
     <dependency>
       <groupId>junit</groupId>
       <artifactId>junit</artifactId>
       <version>4.10</version>
     </dependency>
    </dependencies>
  </dependencyManagement>

  <dependencies>
   <dependency>
     <groupId>junit</groupId>
     <artifactId>junit</artifactId>
     <version>4.9</version>
   </dependency>
  </dependencies>
</project>
```

```bash
$ mvn dependency:tree
[INFO] --- dependency:3.7.0:tree (default-cli) @ hi ---
[INFO] com.example:hi:jar:1.0.0
[INFO] \- junit:junit:jar:4.9:compile
[INFO]    \- org.hamcrest:hamcrest-core:jar:1.1:compile
```

This shows that direct `dependency` JUnit 4.9 won over `dependencyManagement` JUnit 4.10.

lefou:
> > In sbt for example, the values of `dependencyOverrides` are equivalent to BOM values. As a consequence, these can be published.
> 
> I think this is mostly true, but there is a small details of Maven dependency quirk, which we might have to be aware: I think Maven prioritizes direct dependency over dependencyManagement.

I think that matches my perception. Local dependency declarations override management overrides parent poms. There are a lot of precedence rules in Maven with regard to dependency management. I think the following page list some or all (?) of them:

https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#dependency-management

It also mentions a new packaging type `bom` (since Maven 4.0 and the POM Model 4.1.0).

In case Coursier will follow the Maven semantics, we should name the new task in Mill rather `dependencyManagement` than `dependencyOverrides`, as "overrides" implies a simple rule. Too simple.

jin:
cc @shs96c for rules_jvm_external's BOM support. 

We currently ask folks who need BOMs to switch to a different Aether based resolver (https://github.com/bazel-contrib/rules_jvm_external?tab=readme-ov-file#experimental-support-for-maven-bom-files), so it would definitely be nice to have this.

We will be happy to try out a build of coursier with our tests (https://github.com/bazel-contrib/rules_jvm_external/blob/master/MODULE.bazel#L432)

alexarchambault:
> What needs to be done in Mill:
> 
> * Decide, how we do declare managed dependencies? We should add a new task to `CoursierModule`. We need to agree on a good name. The type should be `T[Seq[Dep]]`.
> * Decide, how we do import a BOM dependencies in `import`-scope? In Mill dependencies don't have scopes, but we manage multiple scope-like dependency sets: `ivyDeps` ~= `compile`-scope, `runIvyDeps` ~= `runtime`-scope, `compileIvyDeps` ~= `provided`-scope. (To import BOMs via `ivyDeps`, we need to find out, if Maven makes a distinction between these scopes and if there are any situations, where mixing `compile` and `import` scope for BOM management is problematic.) We better add a new task to import those BOMs. The task name might match the other new task name but add a suffix like `Includes` or `Imports`. The type is also `T[Seq[Dep]]`.
> * How can we produce BOM artifacts? I think this should already work by changing the `PublishModule.pomPackagingType` to `Pom`. We just need to generate the `<dependencyManagement>` node in the `pom` task. Don't know what the equivalent to `ivy` is, if there is any.

@lefou Thanks for all these details. Let's discuss these Mill-specific points in an upcoming Mill issue or early PR that I'll open about this

alexarchambault:
Going to merge this, with the feature enabled by default. I should mention it in the next release notes, but those changes can be easily disabled, and you might want to allow sbt / Mill / rules_jvm_external users to do so.

From the API, it can be disabled via the `ResolutionParams`, like
```scala
Resolve()
  .mapResolutionParams(_.withEnableDependencyOverrides(Some(false)))
// or
Fetch()
  .mapResolutionParams(_.withEnableDependencyOverrides(Some(false)))
```

On the CLI, `--enable-dependency-overrides=false` allows to disable it.

alexarchambault:
For `scala-{library,reflect,compiler}`, most of the time, these versions are just forced during dependency resolution. We might want to manage those via a dependency management section or even a BOM now 🤔 That BOM could even be managed by the Scala compiler team itself, especially for Scala 3, where the scala-library version (2.13 scala-library that Scala 3 pulls) cannot be mapped directly from the Scala 3 version.

alexarchambault:
cc @Gedochao, as Scala CLI resolves dependencies and publishes them too. IIRC, Scala CLI uses coursier via [coursier-interface](https://github.com/coursier/interface). Options to disable the feature here should be added to it too, to allow Scala CLI user to disable it too. Maybe some BOM support could be added to Scala CLI too

PR #3107, opened 2024-10-23
Add more resolution tests

alexarchambault:
Tests involving BOMs / parent POMs in particular, so that the impact of https://github.com/coursier/coursier/pull/3097 can be better captured later on


alexarchambault:
(About to add Spring / Quarkus stuff)

PR #3112, opened 2024-10-24
Add resolution consistency check

alexarchambault:
There are two main ways to get dependency lists out of a `Resolution` instance:
- `orderedDependencies` (main one), which should be used to compute an ordered list of artifacts, and walks the dependency graph to compute its result
- `minDependencies`, used internally by `Resolution` itself, that contains a more compact list (removing some redundancies)

This adds checks in the tests that the two are consistent with each other (no extra dependencies on one side or the other).

This is being helpful in https://github.com/coursier/coursier/pull/3097 to debug new developments in `Resolution`.


alexarchambault:
Hopefully, these checks will be green here

PR #3137, opened 2024-11-04
Build things with JDK 17, but ensure Java 8 compat anyway

alexarchambault:
Right now, publishing fails because:
- Mill `0.12.x` publish tasks require JDK >= 11
- using Java >= 11 but building for Java 8 requires passing `--require 8` to javac and scalac
- we're using "unsupported" JDK APIs (`sun.misc.Signal` stuff), and neither javac nor scalac accept options to enable access to it alongside `--release 8` (javac doesn't accept `--add-exports jdk.unsupported/sun.misc=ALL-UNNAMED` alongside `--release`, and scalac doesn't allow passing such an option altogether anyway, see https://github.com/scala/scala/pull/10801)

To work around that, this PR:
- uses JDK 17 everywhere on the CI
- passes `--release 8` to scalac
- moves the (Java) file using "unsupported" JDK APIs to a separate pure Java `cache-util` module
- passes `-source 8 -target 8 -bootclasspath some-jdk-8/jre/lib/rt.jar` to javac in order to allow access to unsupported JDK APIs


som-snytt:
Scala 2:
```
  -Yrelease:<packages>                  Expose platform packages hidden under --release
```
in case that is useful. Not ported to Scala 3 (yet?), but maybe because there was no feedback.

PR #3147, opened 2024-11-05
Add extra installation instruction for ARM64 Windows

Friendseeker:
Currently the Windows installer installs x86_64 build of Coursier, which then installs x86_64 version of Java by default causing every coursier installed JVM app to be slower as they use the installed x86_64 Java.

To mitigate this we should warn against using x86_64 Coursier installer to avoid emulation overhead.


alexarchambault (approved):
Thanks!

PR #3172, opened 2024-11-26
Remove unused snapshots

alexarchambault:
These _seem_ to be unused, although I'm not 100% sure…


alexarchambault:
Seems these are used… 🤔

PR #3178, opened 2024-11-30
Fix '_' version handling by rules

palanski:


palanski:
I see test failures because `org.apache.logging.log4j:log4j-bom:2.24.2` can't be found locally. I had to introduce this BOM because it introduces dependencies that do not interfere with the rest of dependencies tests use. I ran all the tests with `FETCH_MOCK_DATA=1`, so something like this has to be done here. If needed, I can add a commit with missing metadata.

palanski:
* Changed the version of `org.apache.logging.log4j:log4j-bom` to `2.23.1`, it is [present in metadata](https://github.com/coursier/test-metadata/tree/1e1a6a3827607addb5ffed2f92f6d719f6784358/https/repo1.maven.org/maven2/org/apache/logging/log4j/log4j-bom/2.23.1).|
* Fixed formatting.

This should fix the build.

palanski:
```
Checking out the ref
  "C:\Program Files\Git\bin\git.exe" checkout --progress --force refs/remotes/pull/3178/merge
  Error: error: unable to create file modules/tests/shared/src/test/resources/resolutions/com.github.alexarchambault/argonaut-shapeless_6.2_2.11/1.2.0-M11_depf167dfebbad7975d2ae448ab8dd9b3081f511e6a_bomse061b3f216e9c034d7ff893451c376784c2ffc37_paramsf59b950c0c661129fff44bfa4db5052893a4eb0a: Filename too long
```
This file was created by Coursier code while running tests:
`FETCH_MOCK_DATA=1 ./mill "coursier.jvm[2.12.20].test.testOnly" -- coursier.tests.ResolveRulesTests`

tgodzik:
> ```
> Checking out the ref
>   "C:\Program Files\Git\bin\git.exe" checkout --progress --force refs/remotes/pull/3178/merge
>   Error: error: unable to create file modules/tests/shared/src/test/resources/resolutions/com.github.alexarchambault/argonaut-shapeless_6.2_2.11/1.2.0-M11_depf167dfebbad7975d2ae448ab8dd9b3081f511e6a_bomse061b3f216e9c034d7ff893451c376784c2ffc37_paramsf59b950c0c661129fff44bfa4db5052893a4eb0a: Filename too long
> ```
> 
> This file was created by Coursier code while running tests: `FETCH_MOCK_DATA=1 ./mill "coursier.jvm[2.12.20].test.testOnly" -- coursier.tests.ResolveRulesTests`

Looks like we need to change how the names are being created or just make sure they are shortened, though not sure where the code resides. @alexarchambault did you have issues like that?

palanski:
It's [BOMs hash part](https://github.com/coursier/coursier/blob/main/modules/coursier/shared/src/test/scala/coursier/tests/TestHelpers.scala#L92-L105) making filenames longer.

palanski:
> Maybe we should combine the hashes into one? I doubt this would break anything

Good option. Combining dependencies and BOMs together should not break caching because BOMs are dependencies too.

palanski:
> Never ran into that, but I guess these checksums can be shortened, yes. That might be cumbersome to do, though (all existing test data file names with a checksum have to be updated)

This is probably because there have been no cases where BOMs, dependencies, and parameters are present. For example, files including BOM hashes: [one](https://github.com/coursier/coursier/blob/main/modules/tests/shared/src/test/resources/resolutions/ch.epfl.scala/bsp4j/2.2.0-M2_boms4d5b01f7a140549f06d97104b8cae8b186491053), [two](https://github.com/coursier/coursier/blob/main/modules/tests/shared/src/test/resources/resolutions/com.thesamet.scalapb/scalapbc_2.13/0.9.8_boms55f35177bfe89c38caddefa936b7a48d73b4d23f), [three](https://github.com/coursier/coursier/blob/main/modules/tests/shared/src/test/resources/resolutions/com.google.protobuf/protobuf-java/__boms55f35177bfe89c38caddefa936b7a48d73b4d23f).

In order to not change all the files, file names can be formed as the following:
1. If there are dependencies, but no BOMS -- `_deps` hash is used.
2. If there are BOMs, but no dependencies -- `_boms` hash is used.
3. Both BOMs and dependencies are present -- `_deps` hash including both dependencies and BOMs is used.

What do you think?

alexarchambault:
That could work, yes

palanski:
Ready for review?

alexarchambault (approved):
Thanks!

PR #3189, opened 2024-12-11
Tweak Maven scopes handling

alexarchambault:
Seems these were hard-coded in Resolution, making it impossible for users to customize some aspects of configuration handling


alexarchambault:
(Probably an oversight by 2015 me lol)

alexarchambault:
Might add a non-regression test later on

PR #3193, opened 2024-12-19
Optimize BOM handling

alexarchambault:
Fixes https://github.com/coursier/coursier/issues/3190

PR #3195, opened 2024-12-26
Check - cache some hash codes

alexarchambault:
Suspicious thing spotted in https://github.com/coursier/coursier/pull/3193. Caching these hash codes (and slightly changing how they're computed along the way) ~doesn't break CI on the current main~ (see below). It should break an obscure test case once https://github.com/coursier/coursier/pull/3193 is there. I'm not sure why yet…


alexarchambault:
So it seems this breaks when using Scala 2.12, but not when using 2.13:
```text
$ ./mill 'tests.jvm[2.12.20].test' "coursier.tests.CentralTests" # fails
$ ./mill 'tests.jvm[2.13.15].test' "coursier.tests.CentralTests" # succeeds
```

Might involve some undefined behavior when handling sequences and maps…

modules/core/shared/src/main/scala/coursier/core/DependencyManagement.scala
```diff
@@ -13,6 +13,10 @@ object DependencyManagement {
     `type`: Type,
     classifier: Classifier
   ) {
+
+    override lazy val hashCode: Int =
+      tuple.hashCode()
```
He-Pin, line 18: how about `if (cachedHashCode ne null) cachedHashCode else cachedHashCode = tuple.hashCode`

alexarchambault: That might work too

PR #3238, opened 2025-01-18
feat: add maven 2 meta version support

yzia2000:
There are some maven poms out there with transitive dependencies using (RELEASE and LATEST).

Would be great if we can support resolving:
RELEASE (non-snapshot latest) -> "latest.release"
LATEST (snapshot + non-snapshot latest) -> "latest.integration"


yzia2000:
Attempt at https://github.com/coursier/coursier/issues/3239

Issue #3236, opened 2025-01-17
Dependency tree generated for `org.apache.pulsar:bouncy-castle-bc:4.0.1` is missing dependencies


shs96c:
Steps to reproduce:

`coursier fetch "org.apache.pulsar:bouncy-castle-bc:4.0.1" --artifact-type "jar,json,aar,bundle,eclipse-plugin,exe,orbit,test-jar,hk2-jar,maven-plugin,scala-jar,dylib,so,dll,pom,src,doc" --verbose --no-default --json-output-file "dep-tree.json" --checksum "SHA-1,MD5" --repository 'https://repo1.maven.org/maven2'`

Examine the generated `dep-tree.json` file and notice that `"org.bouncycastle:bcprov-ext-jdk18on:1.78.1"` is listed as a direct dependency, but there's no matching entry in the `dependencies` array:

```json
{
  "conflict_resolution": {},
  "dependencies": [
    {
      "coord": "org.apache.pulsar:bouncy-castle-bc:4.0.1",
      "file": "/Users/shs/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/apache/pulsar/bouncy-castle-bc/4.0.1/bouncy-castle-bc-4.0.1.jar",
      "directDependencies": [
        "org.bouncycastle:bcpkix-jdk18on:1.78.1",
        "org.bouncycastle:bcprov-ext-jdk18on:1.78.1"
      ],
      "dependencies": [
        "org.bouncycastle:bcpkix-jdk18on:1.78.1",
        "org.bouncycastle:bcprov-ext-jdk18on:1.78.1",
        "org.bouncycastle:bcprov-jdk18on:1.78.1",
        "org.bouncycastle:bcprov-jdk18on:jar:1.78.1",
        "org.bouncycastle:bcutil-jdk18on:jar:1.78.1"
      ]
    },
    {
      "coord": "org.bouncycastle:bcpkix-jdk18on:1.78.1",
      "file": "/Users/shs/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/bouncycastle/bcpkix-jdk18on/1.78.1/bcpkix-jdk18on-1.78.1.jar",
      "directDependencies": [
        "org.bouncycastle:bcprov-jdk18on:jar:1.78.1",
        "org.bouncycastle:bcutil-jdk18on:jar:1.78.1"
      ],
      "dependencies": [
        "org.bouncycastle:bcprov-jdk18on:jar:1.78.1",
[...]

PR #3242, opened 2025-01-20
Use VersionConstraint and Version rather than String for versions

alexarchambault:
Goal of this PR is to replace the use of `String` as a type for versions by two types:
- `VersionConstraint`, that can model things like specific versions (`1.2.3`), but also intervals (`[1.0,2)` or `2.2+`) and mixes of intervals and specific versions,
- `Version`, that corresponds to a specific version, like `1.2.3`, and not to an interval

Making the distinction allowed to debug (and fix, I think) things around printing dependency trees in particular.

That refactoring prepares the ground for upcoming Gradle module support, and its [dependency constraints](https://docs.gradle.org/current/userguide/dependency_constraints.html) in particular, which should be handled via `VersionConstraint`.

These two types keep both the original string they're built from, and the result of parsing it (that doesn't fail, by construction). They lazily parse this input string. So this adds a nice optimization: versions and version constraints should now be basically parsed only when they're read from inputs (user input, POM / ivy files, etc.). The result of parsing should be passed around transparently, and no parsing needs to happen later on.

This refactoring should also open the way to further optimizations for https://github.com/coursier/coursier/pull/3097 (more refined version constraints should allow to shrink the number of relevant dependencies exposed by `DependencySet`, which should lower the memory usage and amount of calculations during resolution).

Lastly, it should also open the way for support of version constraints from other ecosystems (NPM's `^X.Y.0` or `~X.Y.Z` or pip's version ranges), if ever that's needed in the future.

Note that both `VersionConstraint` and `Version` live in the [coursier-versions](https://github.com/coursier/versions) repository.

Fixes https://github.com/coursier/coursier/issues/3236


alexarchambault:
This breaks binary compatibility for now, that should be fixed prior to merging

alexarchambault:
Will probably add more non-regression tests for reports prior to merging that…

alexarchambault:
Will probably ponder those changes a few days… Beyond the newly introduced variables (that might need renaming), this also changes quite a bit JSON report generation. The latter should be less buggy, but we actually can't really add a flag to enable the former way of generating a report (the `Version` / `VersionConstraint` refactoring changed too many things). I'd like to think about the best way to handle that change.

alexarchambault:
Note that this PR overhauls JSON report generation. The updated reports in test data in the diff of this PR illustrate how this PR changes reports. Basically, for Android dependencies, it makes sure both the jar and the aar files are present in the report (if both file types are asked). In other cases (quarkus stuff in the tests), it makes sure dependencies aren't duplicated. It also doesn't add `jar` in the dependency strings, which is assumed to be the default type (`"org.bouncycastle:bcprov-jdk18on:jar:1.78.1"` is now just `"org.bouncycastle:bcprov-jdk18on:1.78.1"`). It seems reports were also missing dependencies (missing items in the `dependencies` field of individual entries, not missing entries in the global `dependencies` list).

alexarchambault:
Fixes https://github.com/coursier/coursier/issues/3236

alexarchambault:
> Note that this PR overhauls JSON report generation. The updated reports in test data in the diff of this PR illustrate how this PR changes reports.
> 
> * for Android dependencies, it makes sure both the jar and the aar files are present in the report (if both file types are asked)
> * in other cases (quarkus stuff in the tests), it makes sure dependencies aren't duplicated
> * it also doesn't add `jar` in the dependency strings, which is assumed to be the default type (`"org.bouncycastle:bcprov-jdk18on:jar:1.78.1"` is now just `"org.bouncycastle:bcprov-jdk18on:1.78.1"`)
> * it seems reports were also missing dependencies (missing items in the `dependencies` field of individual entries, not missing entries in the global `dependencies` list), they don't any more
> 
> An option, `--legacy-report-no-guarantees`, allows to fall back to the former code that generates reports. But that code seems more buggy than before this PR, because of the large refactoring that uses `Version` / `VersionConstraint` to handle versions.

FYI @shs96c

PR #3250, opened 2025-01-21
Build launchers for Linux ARM64

alexarchambault:


alexarchambault:
cc @tgodzik @Gedochao I'm going to cut some milestone releases to see how this goes. If this goes well, there might be no need to cut new releases in [VirtusLab/coursier-m1](https://github.com/VirtusLab/coursier-m1)

PR #3261, opened 2025-02-03
Use more standard syntax to define tests

alexarchambault:


alexarchambault:
Should help for https://github.com/coursier/coursier/pull/3258

PR #3308, opened 2025-02-18
Stop using argonaut / argonaut-shapeless

alexarchambault:


alexarchambault:
These dependencies have been dropped now

Issue #3267, opened 2025-02-04
Getting `OverlappingFileLockException` for Windows in Github actions


PriyanshC:
Proof of concept [repository](https://github.com/PriyanshC/coursier-windows-bug)

First noticed in [our project](https://github.com/j-mie6/parsley-debug-app) where we worked around it with manually removing files, enforcing UTF-8 encoding, always running `sbt clean` and setting a different tempdir

```yml
      - name: Clean SBT caches and set env (Windows only)
        if: matrix.platform == 'windows-latest'
        shell: bash
        run: |
          rm -rf "$GITHUB_WORKSPACE/.coursier-cache"
          rm -rf "C:/Users/runneradmin/.ivy2/cache"
          rm -rf "C:/Users/runneradmin/.sbt"
          echo "SBT_OPTS=-Dfile.encoding=UTF-8" >> $GITHUB_ENV

      - name: Setup Java
        uses: actions/setup-java@v4
        with:
          distribution: temurin
          java-version: 11
          
      - name: Setup sbt
        uses: sbt/setup-sbt@v1

      - name: Update sbt
        shell: bash
        run: |
          sbt clean
          sbt -Djava.io.tmpdir=$RUNNER_TEMP update


```

eed3si9n:
I'm seeing this as well on sbt/io, sometimes many times in a row on Windows. The repo uses sbt 1.10.7, which updated Coursier from 2.1.19 → 2.1.22, although I'm not sure if this only started with sbt 1.10.7.

The stacktrace I'm getting is:
<details>

<summary>

```
[error] Caused by: java.nio.channels.OverlappingFileLockException
[error] 	at sun.nio.ch.SharedFileLockTable.checkList(FileLockTable.java:255)
[error] 	at sun.nio.ch.SharedFileLockTable.add(FileLockTable.java:152)
[error] 	at sun.nio.ch.FileChannelImpl.lock(FileChannelImpl.java:1062)
[error] 	at java.nio.channels.FileChannel.lock(FileChannel.java:1053)
[error] 	at lmcoursier.internal.shaded.coursier.paths.CachePath.withStructureLockOnce(CachePath.java:191)
[error] 	at lmcoursier.internal.shaded.coursier.paths.CachePath.withStructureLock(CachePath.java:154)
[error] 	at lmcoursier.internal.shaded.coursier.cache.CacheLocks$.withStructureLock(CacheLocks.scala:23)
[...]

PR #3318, opened 2025-02-24
Retry on OverlappingFileLockException

eed3si9n:
Fixes https://github.com/coursier/coursier/issues/3267

## Problem
On Windows, Coursier is hitting OverlappingFileLockException reliably in some repos. It seems like the the function in question has a retry logic, but OverlappingFileLockException is not checked.

## Solution
This catches OverlappingFileLockException, so it will be retried.


alexarchambault (approved):
Thanks @eed3si9n!

PR #3327, opened 2025-03-05
Add credentials from env var in PlatformRepositoryParser

mikeshepherd:
When loading repositories via the coursierapi (https://github.com/coursier/interface/blob/76af353ae186de1bfcd30c910ec935df4399fd9b/interface/src/main/java/coursierapi/Repository.java#L12) it loads the repositories from the `COURSIER_REPOSITORIES` environment variable, but not the credentials from `COURSIER_CREDENTIALS`.
This PR adds that functionality.

This is the behaviour of the cs command line utility and I believe it makes sense for the coursierapi SDK to do the same, especially given that it already fetches the repositories from the environment.

I can write a test for this, but the only way I can see to do it is to explicitly call `System.setProperty` in the test. This doesn't appear to cause any other test failures but I don't know if this is a behaviour you want to encourage.

### DIFF
--- a/modules/coursier/jvm/src/main/scala/coursier/internal/PlatformRepositoryParser.scala
+++ b/modules/coursier/jvm/src/main/scala/coursier/internal/PlatformRepositoryParser.scala
@@ -4,7 +4,7 @@ import java.io.File
 import java.net.MalformedURLException
 
 import coursier.LocalRepositories
-import coursier.cache.CacheUrl
+import coursier.cache.{CacheDefaults, CacheUrl}
 import coursier.core.{Authentication, Repository}
 import coursier.ivy.IvyRepository
 import coursier.maven.MavenRepositoryLike
@@ -56,15 +56,27 @@ abstract class PlatformRepositoryParser {
         }
       }
 
+      val credentials = CacheDefaults.credentials.flatMap(_.get())
+
       validatedUrl.flatMap { url =>
-        Option(url.getUserInfo) match {
-          case None =>
-            repo
-          case Some(userInfo) =>
-            val authBase = userInfo.split(":", 2) match {
-              case Array(user, password) => Authentication(user, password)
-              case Array(user)           => Authentication(user)
-            }
+        val authBase = Option(url.getUserInfo).map(userInfo =>
+          userInfo.split(":", 2) match {
+            case Array(user, password) => Authentication(user, password)
+            case Array(user)           => Authentication(user)
+          }
+        ).orElse {
+          credentials.find(_.autoMatches(url.toExternalForm, None)).flatMap {
+            c =>
+              (c.usernameOpt, c.passwordOpt) match {
+                case (Some(user), Some(password)) => Some(Authentication(user, password.value))
+                case (Some(user), None)           => Some(Authentication(user))
+                case _                            => None
+              }
+          }
+        }
+
+        authBase match {
+          case Some(authBase) =>
 
             val auth = authBase.withHttpsOnly(url.getProtocol != "http")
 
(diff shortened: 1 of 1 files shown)


tgodzik:
Looks like this is and issue that causes https://github.com/scalameta/metals/issues/7293

@alexarchambault what do you think? Might be super useful to merge.

alexarchambault (commented):
Sorry I didn't review that earlier. Credential stuff are usually quite sensitive, I prefer to take the time to review them carefully. Here, I don't understand why this needs to be added that deep in the code. The parser should be free of externally-read-credentials concerns.

Plus, if this is meant to address an issue with coursier-api (and only there?), I guess the fix should be added in it, not here. Likely in [this file](https://github.com/coursier/interface/blob/dc7be73f3a3d6543dc5bec4590a2026dccd0aa32/interface/src/main/scala/coursier/internal/api/ApiHelper.scala).

Either the bug you're trying to address is tricky, or something in the configuration of credentials needs to be tweaked (or maybe in the way we handle `COURSIER_CREDENTIALS`?). Because the credentials you're adding here should actually be taken into account basically everywhere in coursier (including coursier-api), given they're [pulled by `FileCache` by default](https://github.com/coursier/coursier/blob/8e864315c6879cbccab964323eedf21251906a95/modules/cache/jvm/src/main/scala/coursier/cache/FileCache.scala#L37).

tgodzik:
So essentially this means we would need to implement a lot of the functionality anew in coursierapi then, no? We would need to implement reading the file, validating the url and matching to the repository, which I guess we would mostly copy from here and turn to Java?

tgodzik:
For now I am using this as a workaround in Metals https://github.com/scalameta/metals/pull/7314

mikeshepherd:
> Looks like this is and issue that causes https://github.com/scalameta/metals/issues/7293

I did go looking for an issue there but couldn't find it, but that is exactly the issue that lead me here.

> The parser should be free of externally-read-credentials concerns.

I'm happy to move it, but 

>  address an issue with coursier-api (and only there?)

it's not only there so I don't think that coursier-api would be the right place.

>  actually be taken into account basically everywhere in coursier

This is what I'm trying to address :grin: 
The api [calls](https://github.com/coursier/interface/blob/dc7be73f3a3d6543dc5bec4590a2026dccd0aa32/interface/src/main/scala/coursier/internal/api/ApiHelper.scala#L27) `Resolve.defaultRepositories` which is [here](https://github.com/coursier/coursier/blob/5911c1fecd3549442c97b208314fda8376f5b2ad/modules/coursier/jvm/src/main/scala/coursier/PlatformResolve.scala#L50) in PlatformResolve (jvm) which goes straight through to the file I've changed.

Would moving this into PlatformResolve be a better solution?

> should actually be taken into account basically everywhere in coursier
There seems to be a number of places where this isn't true, there's a scattering of lines that use the same `Resolve.defaultRepositories` function like the api does. Of course maybe they are only referenced from `FileCache` so it doesn't matter that they don't have the credentials if that later adds them :shrug:?

The other option being I update the api to use File Cache instead of the static call it makes now?

tgodzik:
@alexarchambault this seems to be breaking now in every tool that uses coursier API, do you think we could try merging this one? I did some workarounds, but I would have to add them in a lot of places.

tgodzik:
Thanks!

tgodzik:
Isn't that what invoking `cs` manually from commandline does? It adds the credentials blindly, while the API doesn't. So this is a bit inconsistent

alexarchambault:
Seems the initial problem should be fixed by #3801, which should be part of the upcoming `2.1.25`

PR #3341, opened 2025-03-18
Substitute properties in module name

### LINKED ISSUE #3340: Coursier does not resolve properties in `artifactId` for Maven
Hello!

We are using coursier via Bazel rules_jvm_external, and are stumbling across an issue where coursier isn't fetching a file corresponding to a particular (internal) artifact we depend on. It turns out it is due to a weird structure of the `artifactId` in the pom.

The pom is structured as follows:

```xml
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
    <modelVersion>4.0.0</modelVersion>
        <parent>
            <groupId>com.organisation</groupId>
            <artifactId>root-pom</artifactId>
            <version>76</version>
            <relativePath>../pom.xml</relativePath>
        </parent>
    <properties>
        <protocol_name>example-protocol</protocol_name>
    </properties>
    <groupId>com.organisation</groupId>
    <artifactId>protocols-${protocol_name}-${project.version}</artifactId>
</project>
```

And we are referencing the dependency as `com.organisation:protocols-example-protocol-76:76` (i.e., we run `coursier fetch com.organisation:protocols-example-protocol-76:76`)

In the verbose mode we see the logs:

```
Downloading https://maven-repository/repository/public/com/organisation/protocols-example-protocol-76/76/protocols-${protocol_name}-${project.version}-76.jar
Failed to download https://maven-repository/repository/public/com/organisation/protocols-example-protocol-76/76/protocols-${protocol_name}-${project.version}-76.jar
```

Which shows that it does not substitute the properties correctly in the artifact id when fetching.

We are using coursier version 2.1.8, but I can also reproduce this on the latest main branch at time of writing. Interestingly, I had a random coursier install available with version 2.0.16, and it was working correctly.

I saw a similar issue https://github.com/coursier/coursier/issues/1063 and fix https://github.com/coursier/coursier/pull/1213

JohnnyMorganz:
Fixes #3340 

Very similar to the fix in https://github.com/coursier/coursier/pull/1213

My first time writing scala, let me know if there is a better way to do this!


JohnnyMorganz:
Not sure how / where is the best place to write a test, let me know what would be recommended

alexarchambault:
@JohnnyMorganz Can this the issue you're trying to solve be reproduced with artifacts from public repositories? It might be trickier to add non-regression tests else

JohnnyMorganz:
Unfortunately I don't know of any. A public code search of `file:pom.xml artifactId>.*$` does not seem to return anything

PR #3351, opened 2025-03-24
Better default variant attributes

alexarchambault:
.


alexarchambault:
Now part of https://github.com/coursier/coursier/pull/3353

PR #3352, opened 2025-03-24
Enable Gradle Modules by default

alexarchambault:
Not sure I'll go on and actually enable it by default. But this PR show the impact of it on the tests.


alexarchambault:
Includes https://github.com/coursier/coursier/pull/3353 for now

alexarchambault:
(Not sure why the CI is red. Can't reproduce that error locally, and can't find anything the failing jobs' output that would explain it with certainty. That might need bisecting…)

PR #3124, opened 2024-10-28
Stop using FileOutputStream

alexarchambault:

PR #3364, opened 2025-04-04
Use java.nio.file.Path in CoursierPaths instead of java.io.File

vasilmkd:
Hello. The Scala Plugin for IDEA has a copy of the CoursierPaths source code in our project.
https://github.com/JetBrains/intellij-scala/blob/idea251.x/scala/scala-impl/src/org/jetbrains/plugins/scala/project/sdkdetect/repository/CoursierPaths.java

I'd like to remove this copy and use the version provided by coursier directly, since coursier is itself a dependency of the Scala Plugin for IDEA.

Unfortunately, we've recently started a migration to `java.nio.file.Path` and away from `java.io.File`. This will be necessary for proper support of remote development and WSL support in IDEA. Ideally, the CoursierPaths implementation would also use `java.nio.file.Path` behind the scenes.

This draft PR is the first step towards that, to kickstart the process. As an initial idea, I've added new methods which return a `java.nio.file.Path` and deprecated the already existing methods returning a `java.io.File`. I've also replaced the private implementation with `java.nio.file.Path`. I've not touched anything else.

Should we continue along this way, or do you have some other suggestions? Thank you.

### DIFF
--- a/modules/paths/src/main/java/coursier/paths/CoursierPaths.java
+++ b/modules/paths/src/main/java/coursier/paths/CoursierPaths.java
@@ -5,6 +5,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.function.Supplier;
 
 import dev.dirs.ProjectDirectories;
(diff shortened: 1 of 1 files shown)


alexarchambault:
Hi @vasilmkd, about these changes, I assume you'd like the methods returning `Path`s not to rely on `File`-based ones under-the-hood, right?

I've been wanting to switch entirely to NIO already. I ran into some issues, so I didn't fully make the switch (#3120, #3132 are merged, while #3124 isn't). Could you not deprecate the methods returning `File`s for now? I think I'll deprecate them later, maybe cutting a release with solely the switch to the new `Path`-based methods. Some code that relies on them is quite sensitive (addressing issues seen only in large CIs, so not tested here), I'd rather make the switch later.

vasilmkd:
> I assume you'd like the methods returning Paths not to rely on File-based ones under-the-hood, right?
Ideally, yes. I assume you're referring to other usages of `java.io.File` within coursier in general, right? I understand that will take time and it is not urgent for us at all. For the initial step I wish to do (remove the copy pasted code we have), only converting `CoursierPaths` is enough. I believe I've covered that, right?
> Could you not deprecate the methods returning Files for now?
Of course! This is exactly why I opened this PR as a draft. I will push the change as soon as possible.

alexarchambault (approved):
Looks good, thanks!

PR #3426, opened 2025-06-02
Read repositories in config files from Resolve.defaultRepositories

alexarchambault:
This makes the repositories added in `config.json` be used instead of the default repositories (the hard-coded ivy2local and Maven Central), but not instead of the user-added repositories. This was a problem in https://github.com/com-lihaoyi/mill/issues/5229, where an in-memory repository was added, but ended up being ignored when users had a `config.json`.


alexarchambault:
Meant to address https://github.com/com-lihaoyi/mill/issues/5229 ultimately

PR #3435, opened 2025-06-20
Don't pull jsoniter-scala-core_3 from scala-cli-config

alexarchambault:
We already have jsoniter-scala-core_2.13 in the class path, that should be fine


alexarchambault:
I'd expect / hope the CI will fail because the jsoniter-scala version we pull doesn't support Java 8

alexarchambault:
We should run the tests with Java 8, switching to Mill 1.0 should make that easier…

PR #3457, opened 2025-07-15
Update to Mill 1.0

alexarchambault:


mill-build/src/millbuild/Cs.scala
```diff
@@ -0,0 +1,37 @@
+package millbuild
```
lefou, line 1: May I suggest to use `coursierbuild`, to make it more obvious it's custom stuff for coursier.

alexarchambault: Should be fixed

PR #3514, opened 2025-11-01
Cache default coursier config file in Coursier

lihaoyi:
Attempts to fix https://github.com/com-lihaoyi/mill/issues/6049

Not tested, but given the stack traces involved this seems like it should fix the problem by avoiding redundant calls to the slow path on Windows-ARM

CC @alexarchambault


lihaoyi:
ping @alexarchambault

alexarchambault:
This commit fixed the class loader issues: https://github.com/com-lihaoyi/mill/pull/5793/commits/dbcd4b2927194e89ffc2a1a73579ea9e876e276d

alexarchambault:
But IIRC, [this branch](https://github.com/alexarchambault/mill/tree/coursier-launcher) helped too with class loader issues at the time. I wanted to start investigating https://github.com/com-lihaoyi/mill/issues/6049 from it.

Issue #3329, opened 2025-03-06
Resolution.susbtitute is performance bottleneck during resolution


retronym:
I've been profiling coursier resolution in a very large project. We've upgraded to 2.1.24 from a slightly outdated version and noticed the improvement -- in particular the caching in `Dependency.{clearVersion, clearExclusion}`.

Looking at the profiles after those changes, I notice a major remaining hotspot is in `Resolution.substitute` as called from `Resolution.projectProperties`. It looks like this is called at least a few times for each `Project` so one improvement might be as simple as caching the results?

https://github.com/coursier/coursier/compare/main...retronym:coursier:faster/substitute?expand=1

I'm happy to submit a PR but would appreciate some guidance about how you'd like things structure.

<img width="1991" alt="Image" src="https://github.com/user-attachments/assets/c924b7e3-f93b-40a6-ba67-44cb2b68d94e" />

<img width="1975" alt="Image" src="https://github.com/user-attachments/assets/8b3f9757-058a-498c-b6a6-ae224ced36f3" />

retronym:
@alexarchambault I've spent a few days looking at coursier performance and getting an understanding of the internals.

Results are promising, I've got about a 2x speedup on the sparkSql benchmark:

https://github.com/retronym/coursier/pull/1

Still need to clean up the patches, measure impact of the component changes, and break into reviewable chunks.

The big ideas are:
 - Lazily substiutite properties
 - Cache parsed interpolated strings (holding cached values in places where we don't need a new map lookup to access them)
 - Speed up hashmap building in some places (immutable.HashMap.transform is way faster than map), immutable.HashMap.equals also better than Map.equals
 - and more....

retronym:
@alexarchambault I'm ready to carve some pull requests out of my branch.

I'd welcome your input on how to proceed and any constraints (API compatibility, JVM/JS, Scala 2 vs 3) that I should keep in mind.

## Top Performance Issues

- **Redundant Property Substitution:** Switching to `LazyProperties` (deferring substitution and caching parsed forms directly in `Organization`, `Configuration`, etc.) nearly halved execution time. The existing `map(f: String => String)` API is preserved, but a specialized function type routes to a fast path that avoids repeated parsing.

[...]

PR #3549, opened 2026-01-22
feat: Parallelize reverseDependencies computation for faster resolution

honnix:
This is more of a discussion to see how to potentially improve the performance of resolving a large set of dependencies.

We operate a monorepo using Bazel and rules_jvm_external which uses Coursier to resolve a dependency tree. Since we have a very large number of dependencies, the resolution part has been slow. When examining how Coursier works, we find that a single threaded operation takes very long to finish. Pairing up with Claude, we get this change and would like to discuss with you folks. As we are not familiar with this code base, we have no idea whether Claude is doing the right thing or just messing it up.

From Claude:
-------------------------
Problem:
- Thread dump showed single thread spending 168+ seconds in reverseDependencies
- Scala 2 lazy val holds object lock via synchronized(this) during entire computation
- Sequential iteration through hundreds of dependencies calling finalDependencies0
- Each call invokes substituteProps (CPU-intensive string parsing)
- No parallelism possible due to lazy val locking semantics

Solution:
- Add platform-specific parFlatMap helper to compatibility package
- JVM: Creates FixedThreadPool sized to min(numElements, availableProcessors)
- JS: Sequential fallback (Scala.js doesn't support parallel execution)
- Threads named "coursier-resolution-worker-{id}" for jstack visibility
- Uses executor.invokeAll() to run tasks in parallel, collect results after

Concurrency safety analysis:
- substituteProps: Pure function, local variables only - safe
- finalDependencies: Immutable inputs, local variables - safe
- projectCache/finalDependenciesCache: Immutable Scala Maps - safe
- finalDependenciesCache0: Per-instance ConcurrentHashMap - safe
- parFlatMap: ArrayBuffer accessed sequentially after invokeAll - safe
- Benign race in finalDependencies0: Multiple threads may compute same result for same dep, but computation is deterministic so result is identical. ConcurrentHashMap.put() is atomic. Wasteful but correct.

Files changed:
- modules/core/jvm/.../compatibility/package.scala: Add parallel parFlatMap
- modules/core/js/.../compatibility/package.scala: Add sequential parFlatMap
- modules/core/shared/.../Resolution.scala: Use parFlatMap in reverseDependencies

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

### DIFF
--- a/modules/core/js/src/main/scala/coursier/core/compatibility/package.scala
+++ b/modules/core/js/src/main/scala/coursier/core/compatibility/package.scala
@@ -7,6 +7,13 @@ import org.scalajs.dom.{Node, NodeList}
 import coursier.util.{SaxHandler, Xml}
 
 package object compatibility {
+
+  /** Sequential flatMap for JS - Scala.js doesn't support parallel collections. This provides the
+    * same API as the JVM version for cross-platform compatibility.
+    */
+  def parFlatMap[A, B](seq: Seq[A])(f: A => Seq[B]): Seq[B] =
+    seq.flatMap(f)
+
   def option[A](a: js.Dynamic): Option[A] =
     if (js.typeOf(a) == "undefined") None
     else Some(a.asInstanceOf[A])
--- a/modules/core/js/src/main/scala/coursier/core/compatibility/package.scala
+++ b/modules/core/js/src/main/scala/coursier/core/compatibility/package.scala
@@ -14,6 +14,9 @@ package object compatibility {
   def parFlatMap[A, B](seq: Seq[A])(f: A => Seq[B]): Seq[B] =
     seq.flatMap(f)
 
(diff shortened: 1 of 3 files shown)


honnix:
@alexarchambault Wdyt of this? Thanks.

alexarchambault:
@honnix What do you mean by `2.4.14`? `2.1.24`?

alexarchambault:
@honnix It often does in practice. The more dependencies rely on parent POMs / BOMs / property substitution, the more they put a burden on dependency resolution. And different ecosystems have different habits regarding those.

retronym:
@honnix I'm working on a set of improvements to improve resolution performance. The main areas are property substitution, overrides merging, XML processing and version constraint solving.

https://github.com/coursier/coursier/issues/3329#issuecomment-4296809636

Overall these show 3x speedup in a resolution benchmark (4x reduction of allocation). These changes are starting to land.

PR #3568, opened 2026-03-04
Cache hashCode in ReverseModuleTree.Node

### LINKED ISSUE #3561: Strict rule hangs on large dependency graphs due to exponential hashCode in ReverseModuleTree.Node
## Summary

The `Strict` resolution rule hangs indefinitely when checking for evicted dependencies on projects with large dependency graphs (~600+ transitive dependencies). The root cause is an exponential `hashCode()` computation on `ReverseModuleTree.Node` case class instances during tree rendering.

## Steps to reproduce

1. Have a project with ~600+ transitive JVM dependencies and a few version conflicts (evictions)
2. Enable the `Strict` rule via `ResolutionParams`:
   ```scala
   ResolutionParams().addRule(
     new Strict(
       include = Set(ModuleMatcher.all),
       exclude = Set(/* some excludes */)
     )
   )
   ```
3. Resolve dependencies — the process hangs indefinitely

In our case, the resolution hangs at the final dependency (task 1447/1447 in Mill) and never completes. A JVM thread dump shows all CPU time consumed in `ReverseModuleTree$Node.hashCode`.

## Root cause analysis

The hang occurs in this call chain:

```
Strict.EvictedDependencies.<init>     (Strict.scala:97-101)
  → evicted.map(_.repr)               calls Conflicted.repr
  → Conflicted.repr                    (Conflict.scala:77-109)
    → Tree(Seq(tree))(_.dependees)     creates tree of ReverseModuleTree nodes
    → tree0.render(show)               calls Tree.customRender
      → Tree.customRender              (Tree.scala:17-60)
        → recursivePrint               (Tree.scala:34-55)
          → ancestors.contains(elem)   (line 41) — requires hashCode on Node
          → ancestors + elem           (line 49) — requires hashCode on Node
```

### The `hashCode` problem

[`ReverseModuleTree.Node`](https://github.com/coursier/coursier/blob/main/modules/core/shared/src/main/scala/coursier/graph/ReverseModuleTree.scala#L237-L249) is a `case class` with these fields:

```scala
private[graph] final case class Node(
  module: Module,
  reconciledVersionConstraint: VersionConstraint,
  retainedVersion0: Version,
  dependsOnModule: Module,
  dependsOnVersionConstraint: VersionConstraint,
  dependsOnRetainedVersion0: Version,
  excludedDependsOn: Boolean,
  endorsedDependsOn: Boolean,
  allDependees: Map[Module, Seq[(Module, VersionConstraint, Boolean, Boolean)]],
  versions: Map[Module, (VersionConstraint, Version)],
  rootDependencies: Map[Module, Seq[VersionConstraint]]
) extends ReverseModuleTree
```

The critical fields are `allDependees`, `versions`, and `rootDependencies` — these are **shared maps representing the entire dependency graph**. For a project with 600+ dependencies, `versions` has 600+ entries and `allDependees` has hundreds of entries with nested tuples.

Since `Node` is a `case class`, its auto-generated `hashCode()` recursively hashes **all** fields, including these large maps. Every call to `hashCode()` traverses the entire dependency graph data structure.

### How this becomes exponential

In [`Tree.customRender`](https://github.com/coursier/coursier/blob/main/modules/core/shared/src/main/scala/coursier/util/Tree.scala#L34-L55):

```scala
def recursivePrint(elems: Seq[A], ancestors: Set[A], ...): Unit = {
  val unseenElems = elems.filterNot(ancestors.contains)  // hashCode on each elem
  for ((elem, idx) <- unseenElems.iterator.zipWithIndex) {
    // ...
    recursivePrint(children(elem), ancestors + elem, ...) // hashCode again to add to Set
  }
}
```

- `ancestors` is a `Set[ReverseModuleTree.Node]`
- `ancestors.contains(elem)` calls `elem.hashCode()` — which hashes the entire dep graph maps
- `ancestors + elem` also calls `elem.hashCode()`
- Each child `Node` (created by [`Node.actualDependees`](https://github.com/coursier/coursier/blob/main/modules/core/shared/src/main/scala/coursier/graph/ReverseModuleTree.scala#L267-L283)) carries the **same** `allDependees` and `versions` maps
- So every node at every level of the tree triggers a full hash of the entire dependency graph

For a tree with depth D and branching factor B, this results in **O(B^D × N)** hash operations where N is the number of dependencies (~600). In practice, even moderate trees cause the computation to take hours or never finish.

## Proposed fix

There are two complementary fixes:

### Fix 1 (immediate): Use `Conflict.repr` instead of `Conflicted.repr` in error messages

In [`Strict.scala` line 100](https://github.com/coursier/coursier/blob/main/modules/coursier/shared/src/main/scala/coursier/params/rule/Strict.scala#L100):

```diff
-          evicted.map(_.repr + System.lineSeparator()).mkString
+          evicted.map(_.conflict.repr + System.lineSeparator()).mkString
```

`Conflicted.repr` renders an expensive reverse dependency tree. `Conflict.repr` produces a simple one-line message like `"com.squareup.okio:okio:3.16.4 selected, but com.box:box-java-sdk:10.5.0 wanted 3.5.0"` — this is much more useful in error messages anyway, and avoids the tree rendering entirely.

### Fix 2 (structural): Override `hashCode`/`equals` on `ReverseModuleTree.Node`

The `Node` case class should not include the shared graph maps in its identity:

```scala
private[graph] final case class Node(...) extends ReverseModuleTree {
  override def hashCode(): Int =
    (module, reconciledVersionConstraint, dependsOnModule, dependsOnVersionConstraint).hashCode()
  override def equals(obj: Any): Boolean = obj match {
    case other: Node =>
      module == other.module &&
      reconciledVersionConstraint == other.reconciledVersionConstraint &&
      dependsOnModule == other.dependsOnModule &&
      dependsOnVersionConstraint == other.dependsOnVersionConstraint
    case _ => false
  }
}
```

This ensures `hashCode()` is O(1) regardless of dependency graph size, while still correctly identifying unique nodes for cycle detection in `Tree.customRender`.

## Environment

- Coursier version: 2.1.25-M23 (bundled with Mill 1.1.2)
- Scala: 3.8.2
- ~670 resolved JVM dependencies, 4 version conflicts (evictions)
- macOS, JDK 25

## Workaround

Until fixed, users can work around this by adding the conflicting modules to the `Strict` rule's `exclude` set so no evictions are detected.

nguyenyou:
## Summary

- Cache `hashCode` in `ReverseModuleTree.Node` as a `lazy val`, following the existing pattern in `Module`, `Dependency`, `Project`, etc.
- Hash only identity fields (module, versions, booleans), excluding the large shared `Map` fields (`allDependees`, `versions`, `rootDependencies`)
- Add regression test verifying hashCode caching and Set operation performance with large maps

## Context

Fixes #3561

`ReverseModuleTree.Node` is a `case class` with large shared `Map` fields. The auto-generated `hashCode()` recursively hashes these maps on every call. `Tree.render` (used by `Conflicted.repr`) uses `Set[Node]` for cycle detection, calling `hashCode()` on every `contains`/`+` operation. This causes the Strict rule to hang on large dependency graphs.

Minimal reproduction: https://github.com/nguyenyou/coursier-issue-3561

## Test plan

- [x] Existing `core.jvm[2.13.16].test` passes (206 tests)
- [x] New `ReverseModuleTreeTests.nodeHashCodeIsCached` verifies caching
- [x] New `ReverseModuleTreeTests.nodeHashCodePerformance` verifies Set operations with 100 nodes + 500-entry maps complete in < 1s
- [x] Compiles on both Scala 2.12 and 2.13

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/modules/core/shared/src/test/scala/coursier/graph/ReverseModuleTreeTests.scala
+++ b/modules/core/shared/src/test/scala/coursier/graph/ReverseModuleTreeTests.scala
@@ -88,7 +88,7 @@ object ReverseModuleTreeTests extends TestSuite {
         set.contains(node)
         set = set + node
       }
-      val elapsedMs = (System.nanoTime() - start) / 1_000_000
+      val elapsedMs = (System.nanoTime() - start) / 1000000
 
       // With cached hashCode, this should complete in well under 1 second.
       // Without caching, 100 nodes * 500-entry maps would take much longer.


alexarchambault (approved):
Merging, thanks @nguyenyou!

PR #3579, opened 2026-03-17
Run shutdown hooks right before doing execve

alexarchambault:


alexarchambault:
On hold until I can ensure the native CLI tests are actually run on the CI (via https://github.com/coursier/coursier/pull/3584)

PR #3628, opened 2026-04-06
Add .gitattributes to recognize .mill files as Scala

xuwei-k:
For GitHub syntax highlight 

https://github.com/com-lihaoyi/mill/blob/b79ae7dd4bd3576ef95223f218e9a9127cc332cc/.gitattributes

image

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.mill linguist-language=Scala


alexarchambault (approved):
Thanks @xuwei-k!

PR #3630, opened 2026-04-19
Integrate benchmarks into build and address bitrot

retronym:
I'm not very familiar with Mill, nor with the history of the SAX/DOM parser options in Coursier, but this compiles and allows some benchmarks to run again.

```
./mill benchmark.runJmh coursier.benchmark.ResolutionTests.sparkSql -prof jfr:configName=$PWD/out/Profiling.jfc -wi 5 -i 5 -f1

```

Preview of what I'm using this for: 

https://github.com/retronym/coursier/pull/1#issuecomment-4275024457
https://0f693c45.branch-bench-epoch-27.pages.dev/report

Issue #3557, opened 2026-02-22
Timeouts during resolutions on CI


alexarchambault:
One reported [here](https://github.com/coursier/coursier/discussions/3506). Two others reported [here](https://github.com/com-lihaoyi/mill/issues/6823#issuecomment-3897354970) and [here](https://github.com/com-lihaoyi/mill/issues/6823#issuecomment-3900990929)

All 3 reports provide very little info about what could be going on.

alexarchambault:
Got OOMs too now:

[First one](https://github.com/coursier/coursier/actions/runs/23324057572/job/67841398260)
```text
0011] mill.javalib.JvmWorkerModule/classpath
0011] Exception in thread "coursier-pool-1-thread-2" java.lang.OutOfMemoryError: Java heap space
0011] 	at scala.collection.immutable.Vector.iterator(Vector.scala:135)
0011] 	at scala.collection.IterableOps.groupBy(Iterable.scala:561)
0011] 	at scala.collection.IterableOps.groupBy$(Iterable.scala:137)
0011] 	at scala.collection.AbstractIterable.groupBy(Iterable.scala:946)
0011] 	at coursier.core.Resolution$.merge0(Resolution.scala:356)
0011] 	at coursier.core.Resolution.nextDependenciesAndConflicts$lzycompute(Resolution.scala:1589)
0011] 	at coursier.core.Resolution.nextDependenciesAndConflicts(Resolution.scala:1586)
0011] 	at coursier.core.Resolution.missingFromCache$lzycompute(Resolution.scala:1635)
0011] 	at coursier.core.Resolution.missingFromCache(Resolution.scala:1629)
0011] 	at coursier.core.Resolution.nextIfNoMissing(Resolution.scala:1741)
0011] 	at coursier.core.ResolutionProcess$.apply(ResolutionProcess.scala:408)
0011] 	at coursier.core.ResolutionProcess$.$anonfun$apply$1(ResolutionProcess.scala:413)
0011] 	at coursier.core.ResolutionProcess$$$Lambda/0x00007f3f59b13cd8.apply(Unknown Source)
0011] 	at coursier.core.Continue.next(ResolutionProcess.scala:208)
0011] 	at coursier.core.Continue.nextNoCont(ResolutionProcess.scala:211)
0011] 	at coursier.core.ResolutionProcess.run0(ResolutionProcess.scala:35)
[...]

PR #3635, opened 2026-04-28
Improve performance of Overrides

retronym:
- Convert all instances to s.c.i.HashMap (faster equality)
 - Cache an expensive call site to addAll that has a high cache hit ratio
 - Avoid tuple allocation and prefer Map.transform
 - Where available use s.c.i.HashMapBuilder non-public API to
   lookup bindings during building and avoid tuple overhead
   
 Context -- this is part 1/N of a set of performance improvements described https://github.com/coursier/coursier/issues/3329#issuecomment-4296809636
 
 This branch builds on top of https://github.com/coursier/coursier/pull/3630 -- that could be reviewed/merged first and I'll rebase.

### DIFF
--- a/mill-build/src/coursierbuild/Deps.scala
+++ b/mill-build/src/coursierbuild/Deps.scala
@@ -97,7 +97,9 @@ object Deps {
     def scala3   = "3.3.6"
     def scala213 = "2.13.16"
     def scala212 = "2.12.20"
-    val all      = Seq(scala213, scala212)
+    // TODO SCALA_213_BASELINE search for this TODO in the codebase
+    // for cleanup tasks when we move to Scala 2.13 as as the baseline
+    val all = Seq(scala213, scala212)
 
     def scalaJs = "1.19.0"
   }
--- a/modules/core/jvm/src/main/scala/coursier/util/HashMapBuilderFactory.scala
+++ b/modules/core/jvm/src/main/scala/coursier/util/HashMapBuilderFactory.scala
@@ -1,0 +1,74 @@
+package coursier.util
+
+import java.lang.invoke.{MethodHandle, MethodHandles, MethodType}
+import scala.collection.immutable.HashMap
+import scala.collection.compat._
+
+private[coursier] object HashMapBuilderFactory {
+  def apply[A, B <: AnyRef]: HashMapBuilder[A, B] =
+    if (getOrElseMH == null || addOneMH == null) new CompatibleHashMapBuilder[A, B]
+    else new FastHashMapBuilder[A, B]
+  // Uses non-public API of scala.collection.immutable.HashMapBuilder
+  // to allow lookup of values mid-build and to enable tuple-free addOne.
+  final class FastHashMapBuilder[A, B <: AnyRef] extends HashMapBuilder[A, B] {
+    import HashMapBuilder.toGrowableExtensionMethods
+    lazy val delegate = HashMap.newBuilder[A, B]
+
+    // TODO SCALA_213_BASELINE Uncomment when Scala 2.13 becomes the baseline
+    // override def knownSize: Int = delegate.knownSize
+
+    def clear(): Unit = delegate.clear()
+
+    def result(): HashMap[A, B] = delegate.result()
+
+    def getOrNull(a: A): B =
+      getOrElseMH.invoke(delegate, a, null).asInstanceOf[B]
+
+    override def add(a: A, b: B): FastHashMapBuilder.this.type = {
+      addOneMH.invoke(delegate, a, b)
+      this
+    }
+
+    def addOne(elem: (A, B)): FastHashMapBuilder.this.type = {
+      delegate.addOne(elem)
+      this
+    }
+
+    def addAll(elems: IterableOnce[(A, B)]): FastHashMapBuilder.this.type = {
+      delegate.addAll(elems)
+      this
+    }
+  }
+
+  private val lookup = MethodHandles.lookup()
+  private final val getOrElseMH: MethodHandle =
+    try
+      lookup.findVirtual(
+        Class.forName("scala.collection.immutable.HashMapBuilder"),
+        "getOrElse",
+        MethodType.methodType(
+          classOf[java.lang.Object],
+          classOf[java.lang.Object],
+          classOf[java.lang.Object]
+        )
+      )
+    catch {
+      case e: ReflectiveOperationException => null
+    }
+  private final val addOneMH: MethodHandle =
+    try {
+      val builderClass = Class.forName("scala.collection.immutable.HashMapBuilder")
+      lookup.findVirtual(
+        builderClass,
+        "addOne",
+        MethodType.methodType(
+          builderClass,
+          classOf[java.lang.Object],
+          classOf[java.lang.Object]
+        )
+      )
+    }
+    catch {
+      case e: ReflectiveOperationException => null
+    }
+}
--- a/modules/core/shared/src/main/scala/coursier/core/DependencyManagement.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/DependencyManagement.scala
@@ -210,36 +210,55 @@ object DependencyManagement {
       b.result().toMap
     }
 
+  case class AddAllResult(map: Map, mayContainGlobal: Boolean)
+
   def addAll(
     initialMap: Map,
     entries: Seq[GenericMap],
     composeValues: Boolean = true
-  ): GenericMap =
-    if (entries.forall(_.isEmpty))
-      initialMap
-    else {
-      val b = new mutable.HashMap[Key, Values]
-      b.sizeHint(entries.iterator.map(_.size).sum)
-      val it = entries.iterator.flatMap(_.iterator)
-      while (it.hasNext) {
-        val (key0, incomingValues) = it.next()
-        val newValuesOpt = b.get(key0).orElse(initialMap.get(key0)) match {
-          case Some(previousValues) =>
-            if (composeValues)
-              Some(previousValues.orElse(incomingValues))
-                .filter(_ != previousValues)
-            else
-              None
-          case None =>
-            Some(incomingValues)
+  ): AddAllResult = {
+    var mayContainGlobal = false
+    val builder: coursier.util.HashMapBuilder[Key, Values] =
+      coursier.util.HashMapBuilderFactory.apply
+
+    val allEntries = if (initialMap.isEmpty) entries.toList else initialMap :: entries.toList
+    allEntries match {
+      case head :: tail =>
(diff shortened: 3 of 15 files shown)


alexarchambault:
@retronym Just sent you a collaborator invite to the coursier repo so that you can freely re-run the CI if needed

retronym:
This is ready for review.

Here are updated benchmark results for this PR.

https://5200eabe.coursier-benchmark-results.pages.dev/

mill-build/src/coursierbuild/Deps.scala
```diff
@@ -97,7 +97,9 @@ object Deps {
     def scala3   = "3.3.6"
     def scala213 = "2.13.16"
     def scala212 = "2.12.20"
-    val all      = Seq(scala213, scala212)
+    // TODO SCALA_213_BASELINE search for this TODO in the codebase
+    // for cleanup tasks when we move to Scala 2.13 as as the baseline
+    val all = Seq(scala213, scala212)
```
retronym, line 102: @alexarchambault I'd vote for dropping Scala 2.12 support next semi-major release. While it wasn't a huge issue to stay compatible in this PR, it does add friction to the dev cycle, complicate CI etc.

modules/core/jvm/src/main/scala/coursier/util/HashMapBuilderFactory.scala
```diff
@@ -0,0 +1,74 @@
...
+        MethodType.methodType(
+          classOf[java.lang.Object],
+          classOf[java.lang.Object],
+          classOf[java.lang.Object]
+        )
+      )
+    catch {
+      case e: ReflectiveOperationException => null
+    }
+  private final val addOneMH: MethodHandle =
```
retronym, line 58: An alternative would be to place facade classes in the `scala.collection` package to access the `private[collection]` methods.

Its just a matter of which you find the least-bad from a style perspective. Performance will be the same either way.

modules/core/shared/src/main/scala/coursier/core/DependencyManagement.scala
```diff
@@ -210,36 +210,55 @@ object DependencyManagement {
       b.result().toMap
     }
 
+  case class AddAllResult(map: Map, mayContainGlobal: Boolean)
+
   def addAll(
     initialMap: Map,
     entries: Seq[GenericMap],
     composeValues: Boolean = true
```
retronym, line 216: Unrelated to my change, `composeValues` is always true in the internals calls to this method.  Possible bug, or possible opportunity to simplify this?

modules/core/shared/src/main/scala/coursier/core/DependencyManagement.scala
```diff
@@ -210,36 +210,55 @@ object DependencyManagement {
...
+        val headValuesIt = head.valuesIterator
+        while (headValuesIt.hasNext && !mayContainGlobal)
+          if (headValuesIt.next().global)
+            mayContainGlobal = true
+        val it = tail.iterator
+        while (it.hasNext) {
+          val map: collection.Map[Key, Values] = it.next()
+          import scala.collection.compat._
+          map.foreachEntry {
+            (key, incoming) =>
```
retronym, line 242: `foreachEntry` avoids `Tuple2` allocations during iteration.

modules/core/shared/src/main/scala/coursier/core/DependencyManagement.scala
```diff
@@ -210,36 +210,55 @@ object DependencyManagement {
...
+      coursier.util.HashMapBuilderFactory.apply
+
+    val allEntries = if (initialMap.isEmpty) entries.toList else initialMap :: entries.toList
+    allEntries match {
+      case head :: tail =>
+        builder.addAll(head)
+        val headValuesIt = head.valuesIterator
+        while (headValuesIt.hasNext && !mayContainGlobal)
+          if (headValuesIt.next().global)
+            mayContainGlobal = true
```
retronym, line 236: The `filter { case (k, v) => v.global }` below that i removed was a hotspot. I noticed this value is typically false, we can tag such `Overrides` during creation during an existing iteration pass, and use this later to bypass an iteration.

modules/core/jvm/src/main/scala/coursier/util/HashMapBuilderFactory.scala
```diff
@@ -0,0 +1,74 @@
...
+
+    // TODO SCALA_213_BASELINE Uncomment when Scala 2.13 becomes the baseline
+    // override def knownSize: Int = delegate.knownSize
+
+    def clear(): Unit = delegate.clear()
+
+    def result(): HashMap[A, B] = delegate.result()
+
+    def getOrNull(a: A): B =
+      getOrElseMH.invoke(delegate, a, null).asInstanceOf[B]
```
retronym, line 25: Wish that we'd got this method into the API in 2.13.0 as its super useful for performance sensitive code. As it came later we hid it behind `private[collection]` but intended it to be used externally.

modules/core/shared/src/main/scala/coursier/core/Resolution.scala
```diff
@@ -438,6 +442,46 @@ object Resolution {
     )
   }
 
+  private def dictForOverrideMapImpl(
```
retronym, line 445: Extracted to a non-nested method so it can't accidentally use closed-over data that isn't part of the cache key.

mill-build/src/coursierbuild/Deps.scala
alexarchambault: I think the main motivation for keeping it is its use by sbt 1.x, via [sbt-coursier](https://github.com/coursier/sbt-coursier)

modules/core/jvm/src/main/scala/coursier/util/HashMapBuilderFactory.scala
alexarchambault: Wouldn't using `private[collection]` rather than reflection work more smoothly when using native-image?

Although I see the integration tests seem to be fine here…

modules/core/shared/src/main/scala/coursier/core/DependencyManagement.scala
alexarchambault: I think it can be simplified. It's a part of the code I've been wanting to overhaul recently, in https://github.com/coursier/coursier/pull/3618.

As of now, https://github.com/coursier/coursier/pull/3618 is more a refactoring than an optimization, as I went for correctness first, and it makes the time to run `ResolveTests` (via `./mill 'coursier.jvm[2.13.16].test.testForked' "coursier.tests.ResolveTests`), or just its `large resolution` test, increase.

I'll rebase it on top of your changes.

modules/core/jvm/src/main/scala/coursier/util/HashMapBuilderFactory.scala
retronym: Note that GraalVM Native Image should still be able to call these as the method handles are constants visible at image build time:

https://github.com/oracle/graal/issues/2761#issuecomment-700058883

modules/core/jvm/src/main/scala/coursier/util/HashMapBuilderFactory.scala
retronym: Constant method handles are supported: https://github.com/oracle/graal/issues/2761#issuecomment-700058883

modules/core/shared/src/main/scala/coursier/core/DependencyManagement.scala
retronym: The biggest challenge I found to reworking the representation of `Overrides` is that it needs to have efficient `hashCode` / `equals`. My initial thought was to make `Overrides.add` just create a stack of `Map`-s, loop over them and compose values on lookup (betting that laziness is a win).

But, given of the way that `@data class Dependency(...., overridesMap: Overrides, ... )` is used in `Set[Dependency]` or `Map[Dependency, ... ]`, this falls apart because AFAICT you would need to fully force all the lazy lookups/composition to properly implement `hashCode`/`equals`.

retronym: BTW, looks like #3618 will break binary compat somewhat. My next PR that refactors property substitution will quite hard to implement without a bincompat change. For instance I need to convert some value classes to regular classes to store the pre-parsed representation of Strings. If there an opportunity to schedule a major release where some of these bincompat breaking changes would be permitted?

alexarchambault (approved):

alexarchambault:
Thanks a lot @retronym, will merge as soon as the CI is green

modules/core/shared/src/main/scala/coursier/core/DependencyManagement.scala
alexarchambault: I was thinking about cutting `2.1.25`, and then go for `2.2.0-M??` (and allow ourselves to break bin compat during these milestones).

One thing that made me hesitating was the perf regression that appeared in the `2.1.25-M??` versions. I've seen OOMs during resolutions (https://github.com/coursier/coursier/issues/3557), and suspect timeouts occasionally seen during resolution on Mill projects' CI are due to OOMs too. There's the perf issue in https://github.com/coursier/coursier/pull/3549 too. Your changes merged by then could help, but I might just recommend some users to be cautious and wait for `2.2.0`.

alexarchambault: A [new one](https://github.com/coursier/coursier/actions/runs/25218291464/job/73943871080?pr=3635#step:7:1072) just happened in this PR's CI

retronym: I'll take a look. Perhaps the new cache hanging off `Overrides.Impl` should have weak refs or be size limited.

retronym: Oh, you mean that might be a pre-existing OOME. I'll still take a look :)

retronym: Maybe worth trying to use `Caffiene` (or at least create a SPI for users to opt in to that) for caches with eviction policies.

retronym: I've submitted a binary compatible subset of the change to property substitution that gets most of the benefits: https://github.com/coursier/coursier/pull/3643

Once the next major release branch is open, I'll see what incremental benefit we get from caching pre-parsed versions of the strings directly alongside those strings in the object model (e.g. in `Configuration.value` would have `Configuration.parsedValue`)

PR #3636, opened 2026-04-30
Improve performance of XML processing

retronym:
- Reduce allocation in entity substitition
 - Reuse a parser instance in a ThreadLocal (discard on error)
 - Reduce allocations in SAX handler lookup

Benchmark results: https://9d1e25dd.coursier-benchmark-results.pages.dev/

PR 2/N from https://github.com/coursier/coursier/issues/3329#issuecomment-4296809636

### DIFF
--- a/modules/core/jvm/src/main/scala/coursier/core/compatibility/package.scala
+++ b/modules/core/jvm/src/main/scala/coursier/core/compatibility/package.scala
@@ -1,17 +1,18 @@
 package coursier.core
 
-import java.io.CharArrayReader
+import java.io.{CharArrayReader, StringWriter}
 import java.util.Locale
 import javax.xml.parsers.ParserConfigurationException
 import javax.xml.parsers.SAXParserFactory
 import javax.xml.XMLConstants
-
 import coursier.util.{SaxHandler, Xml}
 import org.xml.sax
-import org.xml.sax.InputSource
+import org.xml.sax.{InputSource, XMLReader}
 import org.xml.sax.helpers.DefaultHandler
 
+import java.lang.ThreadLocal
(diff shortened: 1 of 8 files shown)


modules/core/jvm/src/main/scala/coursier/core/compatibility/package.scala
```diff
@@ -45,48 +47,163 @@ package object compatibility {
...
+      val hasBom = s.charAt(0) == '\uFEFF'
+      val start = if (hasBom) 1 else 0
+
+      var dst: Array[Char] = null
+      var readPtr = start
+      var writePtr = 0
 
-    var j       = 0
+      // Leapfrog from ampersand to ampersand
+      var nextAmp = s.indexOf('&', readPtr)
```
retronym, line 79: `indexOf` is faster than a while loop with `charAt`. Perhaps with the right set of JIT inlinings the loop can get closer, but with `indexOf` you end up with a single check for LATIN-1 vs UTF-8, bounds checks, and land in the `@IntrinsicCandidate` `StringLatin1.indexOf`.

modules/core/jvm/src/main/scala/coursier/core/compatibility/package.scala
```diff
@@ -45,48 +47,163 @@ package object compatibility {
...
+
+      if (nextAmp == -1) {
+        // Fast path: No ampersands at all
+        if (hasBom) {
+          val reader = new StringReader(s)
+          reader.skip(1)
+          reader.mark(len - 1)
+          return reader
```
retronym, line 87: Avoid a string copy by setting the reader's cursor to start beyond the BOM.

alexarchambault (approved):

PR #3639, opened 2026-05-02
OutOfMemoryException hunt

retronym:
Drop max heap for forked tests down to 2G and enable JFR and hprof capture

Actual recent OOME in CI: https://github.com/coursier/coursier/actions/runs/25218291464/job/73943871080?pr=3635#logs

I think the JVM default max heap for that run was likely 4GB (1/4 RAM size of the runner). AFAICT it wasn't explicity configured in the mill build


retronym:
Investigation led to #3640

PR #3642, opened 2026-05-02
Performance micro-optimisations in resolution hot paths

retronym:
A collection of small allocation- and CPU-reduction improvements identified
in the resolution hot paths:
- **Cache `MinimizedExclusions.toSet`** — pre-compute the singleton set in  `ExcludeAll` and lazily memoize it in `SetBasedExclusions`, replacing the  repeated `++` chain.
- **Eliminate `Seq`/`fold` overhead in `validateCoordinate`** — replace the  generic `foldLeft` over a two-element `Seq` with direct `if/else if` checks,  avoiding boxing and lambda allocation on a hot path.
- **Avoid `Dependency` copy when version constraint is unchanged** — introduce  withVersionConstraintConserve` and use it at all call sites in `Resolution`,  returning `this` when the constraint is already equal.
- **Avoid intermediate collections in `missingFromCache`** — replace the  `boms ++ modules ++ nextModules` chain (two intermediate sets) with a lazy  `Iterator.flatMap` before the final `.toSet`.
  
##  Benchmarks
  
https://78cd20f7.coursier-benchmark-results.pages.dev/?sec=%C2%B7gc.alloc.rate.norm

These improvements are still minor relative to my planned change to property substitution, but [that change](https://github.com/retronym/coursier/pull/2) is not binary compatible so needs to wait until a [2.2.x branch is open for business](https://github.com/coursier/coursier/pull/3635#discussion_r3173659660).

This is PR 3/N from the work summarised in https://github.com/coursier/coursier/issues/3329#issuecomment-4296809636

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
@@ -1017,9 +1017,12 @@ object ArtifactSource {
 
 private[coursier] object Validation {
   def validateCoordinate(value: String, name: String): Either[String, String] =
-    Seq('/', '\\').foldLeft[Either[String, String]](Right(value)) { (acc, char) =>
-      acc.filterOrElse(value => !value.contains(char), s"$name $value contains invalid '$char'")
-    }
+    if (value.contains('/'))
+      Left(s"$name $value contains invalid '/' character")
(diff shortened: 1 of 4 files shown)


alexarchambault (approved):
Thanks @retronym!

PR #3643, opened 2026-05-04
Improve performance of property substitution

retronym:
Lazily substitute properties -- most are never queried.

Cache the parsed form of property values themselves and of
other strings that support property substitution.

Minimize API changes by making the new `LazyProperties` structure
implement `Seq[(String, String)]`

Future work:
  - Change classes like `Organization` to wrap `PropertyExpr` rather
    than raw strings, further reusing parsing of values. This incrementally
    improves performance but is a binary incompatible change
    
## Benchmark:

https://7d8db12a.coursier-benchmark-results.pages.dev/?sec=%C2%B7gc.alloc.rate.norm

## Context

This is part 4/N of the changes described in https://github.com/coursier/coursier/issues/3329#issuecomment-4296809636 . This is the most important change for performance.

### DIFF
--- a/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
+++ b/modules/core/shared/src/main/scala/coursier/core/Definitions.scala
@@ -6,10 +6,15 @@ import coursier.error.VariantError
 import coursier.util.Artifact
 import coursier.version.{Version => Version0}
 import dataclass.data
+import scala.util.hashing.MurmurHash3
 
 final case class Organization(value: String) extends AnyVal {
-  def map(f: String => String): Organization =
-    Organization(f(value))
+  private[coursier] def parsedValue = PropertyExpr.parse(value)
+  def map(f: String => String): Organization = {
+    val o = PropertyExpr.applySubstitution(value, f)
+    if (o eq value) this
+    else Organization(o)
+  }
 }
 
 object Organization {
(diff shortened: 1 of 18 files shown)


modules/core/shared/src/main/scala/coursier/core/Definitions.scala
```diff
@@ -19,8 +24,12 @@ object Organization {
 }
 
 final case class ModuleName(value: String) extends AnyVal {
-  def map(f: String => String): ModuleName =
-    ModuleName(f(value))
+  private[coursier] def parsedValue = PropertyExpr.parse(value)
```
retronym, line 26: Example of a spot that could be a lazy vals when we next break binary compat. We can then change:

```diff
-PropertyExpr.applySubstitution(value, f)
+parsedValue.applySubstitution(value, f)
```

retronym: ... but until then, I'm passing the *unparsed* value only into `PropertyExpr.applySubstitution`, which leaves a central place that parses all of these. That is a viable spot to add a Map-based cache as an alternative to the this caching strategy.

retronym:
@alexarchambault I've rebased this PR.

 Updated benchmark run: https://fcb7bcf0.coursier-benchmark-results.pages.dev/?sec=%C2%B7gc.alloc.rate.norm

alexarchambault:
@retronym I should review that shortly (maybe tomorrow, or on Monday). I don't mind fixing conflicts myself, to speed up merging things.

retronym:
It's a big change so no rush!

retronym:
🛎️ @alexarchambault  -- let me know if I can do anything to help make this easily reviewable.

alexarchambault:
> 🛎️ @alexarchambault -- let me know if I can do anything to help make this easily reviewable.

@retronym Sorry, I was pondering in what extent I wanted to get familiar with these changes before or after merging. The code this touches is very heavily tested, so I think I'll merge in any case. I'll merge and cut a release during the week I think.

retronym:
Ideally cut a milestone release first so we can do a bit of real world testing

alexarchambault:
Merged via https://github.com/coursier/coursier/pull/3662, thanks a lot @retronym!

PR #3650, opened 2026-05-05
Cache allConfigurations

retronym:
This PR makes the cache global without eviction/expiry, but the cache size is typically tiny.

https://ed130ebd.coursier-benchmark-results.pages.dev/?sec=%C2%B7gc.alloc.rate.norm


alexarchambault (approved):

alexarchambault:
Thanks @retronym!

PR #3651, opened 2026-05-07
Tweaking

alexarchambault:
Adjusts the changes of https://github.com/coursier/coursier/pull/3642


alexarchambault:
cc @retronym (2nd commit in particular, which should be innocuous)

PR #3662, opened 2026-05-25
Improve performance of property substitution, by @retronym 

alexarchambault:
Same as @retronym's #3643, with a slightly reworked git history and additional tweaking


retronym:
@alexarchambault Are you planning a milestone release soon? Would be good to get some real world feedback on this and other recent changes.

alexarchambault:
@retronym Trying to get a few more things in, then I'll cut a release. It shouldn't take long now.

PR #3671, opened 2026-06-09
Fix OOM in reverse dependency tree via global visited set

retronym:
Tree.recursivePrint used a path-local `ancestors` set for cycle detection, which prevents infinite loops but allows the same node to be fully re-expanded each time it is reached via a different branch. In the reverse tree this produces exponential output for any node with high fan-in (e.g. slf4j-api with 50+ dependees after the 2.1.17 BOM changes).

Add a `deduplicateNodes: Boolean = false` flag to `customRender`. When enabled, a global `visited` HashSet is maintained; nodes already expanded are printed with a ` (*)` elision marker and their subtree is skipped. The flag defaults to false so forward-tree callers are unaffected — the forward tree intentionally shows the same package on every path.

The reverse-tree call site in Print.scala opts in via `customRender(deduplicateNodes = true)`.

Fixes OutOfMemoryError reported on large dependency sets with coursier 2.1.17+ (BOM support PRs #3097/#3143 increased dependees map fan-out).

Manually tested with https://github.com/retronym/coursier-oome

Opening a draft PR initially as this needs discussion of the behaviour change and automated testing if its deemed the right path.


lefou:
An example tree can be seen in the comment: https://github.com/com-lihaoyi/mill/issues/6823#issuecomment-4659035583

I think the `(*)`-marked reduced tree nodes are a reasonable compromise. I think `maven-dependency-plugin:tree` does the same kind of tree reductions in its text format.

alexarchambault (commented):
Thanks! Seems (most) CI errors are legit

retronym:
@alexarchambault I've updated test expectations

alexarchambault:
I disabled this by default, out of cautiousness for now. There might be things that could be tweaked in the resulting trees (like ` (*)` that should appear at places closer to the root of the tree, it seems). But this can be enabled by default in Mill if you'd like.

PR #3715, opened 2026-06-16
Ensure all modules have class file versions no higher than expected

alexarchambault:


alexarchambault:
Will probably make a Mill plugin for that at some point

PR #3725, opened 2026-06-22
Don't report Gradle BOM endorseStrictVersions pins as dependency edges

Dolfik1:
> This investigation and patch were prepared with the help of an AI assistant. The original cycle
> symptom and the fix were validated by the author against their own `rules_jvm_external` setup.

When a resolution uses Gradle Module Metadata (`--enable-modules`) and a node imports a
BOM with `endorseStrictVersions`, the strict-version pins from that BOM were surfaced as
**dependency edges** in the two "forward" graph outputs — the `fetch --json-output-file`
report and the `--tree` view. These pins are not real edges: they point sideways/backwards
(e.g. `kotlinx-coroutines-core-jvm` listing `kotlinx-coroutines-core`,
`kotlinx-coroutines-android`, `kotlinx-coroutines-play-services`, and even itself), which
downstream consumers that reject cycles (e.g. Bazel `rules_jvm_external` / `jvm_import`) read
as dependency cycles.

The fix filters these endorsed nodes out of the forward graph outputs only. Version
reconciliation is untouched, so the resolved module set / classpath is unchanged.

### Background

Gradle Module Metadata lets a dependency import a platform/BOM and *endorse its strict
versions*. coursier already models this: support was added in
`be55c4cd5` (#3402, "Take Gradle Modules' endorseStrictVersions into account"). That change
introduced an `endorseStrictVersions` flag on dependencies which, during resolution, turns
the BOM's managed versions into **global version constraints** used by version reconciliation
(`Resolution`'s `merge0` / `Overrides.enforceGlobalStrictVersions`). This part is correct and
matches Gradle's semantics: a platform contributes *constraints*, not dependency edges.

The same change also tags these endorsed entries in the dependency graph so they can be
distinguished from real edges. The **reverse** tree makes use of this
(`ReverseModuleTree.endorsedDependsOn`, and `ResolutionError` renders them as
`(endorsed dependency)` and can filter them out). The two **forward** consumers predate this
distinction and still walk `DependencyTree#children` directly, without consulting the
`endorsed` flag.

As a result, with `--enable-modules`, a node such as `kotlinx-coroutines-core-jvm` (whose
Gradle variant imports `kotlinx-coroutines-bom` with `endorseStrictVersions = true`) ends up
reporting, as direct dependencies, every BOM-managed coroutines module that happens to be
present in the resolution — including back-edges to `kotlinx-coroutines-core` and a self-edge
to `kotlinx-coroutines-core-jvm`. The same nodes also appear (and recurse) in `--tree`.

### Root cause

- `DependencyTree#children` appends the BOM strict-version pins as children created with
  `endorsed = true` (introduced in #3402).
- The forward consumers traverse `_.children` without consulting that flag:
  - `JsonReport` walks `_.children` for both the `directDependencies` set and the
    transitive `dependencies` closure.
  - `Print.dependencyTree0` (the forward `--tree`) renders `_.children` directly.

So the pins were emitted as ordinary edges, producing the reverse/self edges that read as
cycles downstream.

### Fix (forward outputs only)

Filter `endorsed` children out of the two forward consumers, mirroring how the reverse tree
already treats them:

- `modules/cli/src/main/scala/coursier/cli/fetch/JsonReport.scala`
  - `directDependencies`: `flatMap(_.children.iterator.filterNot(_.endorsed))`
  - transitive `dependencies` recursion: `filter(tr => !tr.endorsed && !seen(tr.dependency))`
- `modules/core/shared/src/main/scala/coursier/util/Print.scala`
  - forward `--tree`: `Tree(t.toVector)(_.children.filterNot(_.endorsed))`

Decision — for `--tree` we **filter** rather than label. The endorsed child is created
`withTransitive(false)`, but `children` still recurses into it (re-adding endorsed children at
every level), so a label alone would not stop the combinatorial expansion. Filtering matches
the end result of the reverse tree, which already excludes endorsed dependees from expansion.

### Why this is safe

- It is a **reporting-layer change only**. Version reconciliation
  (`merge0` / `enforceGlobalStrictVersions`) is intentionally left untouched, so the resolved
  module set / classpath does not change. The existing `endorseStrictVersions` resolution test
  (which pins `orderedDependencies`) stays green, confirming the classpath is unchanged.
- It does **not change which nodes exist** in the report. The report's node set
  (`JsonReport`'s `keys` / `depEntries`) is derived from the resolved artifacts list, which the
  patch never touches; the change only affects which *edges* are reported. This was verified
  empirically (see below).
- No resolved module is lost: modules that are genuinely in the graph (e.g.
  `kotlinx-coroutines-android`, `kotlinx-coroutines-core-jvm`) still appear as their own
  entries, reached via their real edges. Only the spurious endorsement edges are removed.
- The `directDependenciesMap` fallback branch in `JsonReport`
  (`resolution.dependenciesOf0`) never materializes endorsed entries, so both branches behave
  consistently after the change.

### Tests

Neither forward surface had Gradle-module / endorsed coverage, so each gets a new regression
test. Both use `androidx.test.ext:junit:1.2.1` (the same coordinate as the existing
`endorseStrictVersions` resolution test), which already has recorded metadata — no new
fixtures are added.

- `JsonReportTests` (`endorseStrictVersions`): asserts that the
  `kotlinx-coroutines-core-jvm` node's `directDependencies` and transitive `dependencies`
  contain its real variant deps (`org.jetbrains:annotations`, `kotlin-stdlib*`) and **not** the
  endorsed siblings (`kotlinx-coroutines-core`, nor a self-edge to `kotlinx-coroutines-core-jvm`).
  It builds the report from a resolve-only path (artifacts with empty files) so it runs against
  the offline test cache.
- `ResolveTests` (`gradle modules` → `endorseStrictVersions tree`): asserts that the forward
  `--tree` rooted at `kotlinx-coroutines-core-jvm` no longer contains a `kotlinx-coroutines-core`
  child.

### Verification

- Both new tests pass with the fix; with the production change reverted (tests kept), both
  fail — confirming they reproduce the issue and that `junit:1.2.1` exhibits it.
- The existing `endorseStrictVersions` resolution test stays green (classpath unchanged).
- The full `JsonReportTests` suite (42 tests, including the recorded Gradle-module Android
  snapshots) and the existing `CentralTests` forward/reverse tree tests stay green — the
  forward `--tree` change is a no-op on graphs that have no endorsed nodes.
- No recorded fixture changes: the working tree contains only the source files (two production,
  two test).

### Scope / known limitation (separate, pre-existing issue)

While validating on a full Android graph, we found a *separate* inconsistency that this PR does
**not** address and does **not** introduce: when a POM-only artifact (e.g.
`com.google.android.gms:play-services-location`) depends on a relocated Gradle umbrella module
(`kotlinx-coroutines-core`, whose `.module` redirects via `available-at` to
`kotlinx-coroutines-core-jvm`), the umbrella is resolved but, being a relocation with no
artifact of its own, produces no report entry — while the POM consumer's edge still points at
the umbrella coordinate. We verified this behavior is **identical with and without this patch**
(the report's node set and that edge are byte-identical before/after), so it is independent of
this change; removing the endorsement cycle simply lets downstream tooling progress far enough
to encounter it. It is best handled in its own change (e.g. having the forward report follow
relocation when emitting edges, so a coordinate with no entry is rewritten to its
`available-at` target). Happy to follow up separately.


Dolfik1:
The macos-15 Intel native failure looks unrelated. A network timeout downloading `net.java.dev.jna:jna:5.9.0` from the Maven mirror in the `clitests` fetch tests, not anything this PR changes (report/tree edge filtering only).

Dolfik1:
Let me know if any changes are needed here. Also happy to rebase if that helps with re-running CI.

alexarchambault (approved):
@Dolfik1 Thanks, that looks good!

Let me tweak the added tests a bit, then I'll merge

PR #3766, opened 2026-08-01
Support version ranges in parent POM coordinates

### LINKED ISSUE #3765: Version ranges in <parent> coordinates no longer resolve (regression since 2.1.25-M1)
Resolving an artifact whose POM declares a Maven version range in its <parent> coordinate fails, requesting the literal range string as if it were a version.

Minimal repro:

```sh
> cs version
2.1.25-M26

> cs fetch dev.openfeature.contrib.providers:env-var:0.0.12
Resolution error: Error downloading dev.openfeature.contrib:parent:[1.0,2.0)
  not found: /Users/shu.cheng/.ivy2/local/dev.openfeature.contrib/parent/[1.0,2.0)/ivys/ivy.xml
  not found: https://repo1.maven.org/maven2/dev/openfeature/contrib/parent/[1.0,2.0)/parent-[1.0,2.0).pom
```
coursier is requesting a POM literally named parent-[1.0,2.0).pom instead of resolving the range against maven-metadata.xml (which would select 1.0.0)

Tested with coursier 2.1.23 and 2.1.24 and both fetch parent/maven-metadata.xml and select 1.0.0. From 2.1.25-M1 to 2.1.25-M26 (current) it fails.

Also affects sbt 2 users transitively, since sbt 2 vendors coursier shaded and can't have this overridden from a build. @eed3si9n

shuch3ng:
Store the parent version as a `VersionConstraint` parsed from the POM, so it is never narrowed through `fromVersion`. This removes five `fromVersion` call sites in `Resolution` and restores the pre-2.1.25 lookup path.

Closes #3765

### DIFF
--- a/modules/core/jvm/src/test/scala/coursier/maven/PomParserTests.scala
+++ b/modules/core/jvm/src/test/scala/coursier/maven/PomParserTests.scala
@@ -1,7 +1,9 @@
 package coursier.maven
 
 import utest._
-import coursier.core.Info
+
+import coursier.core.{Info, Module, ModuleName, Organization}
+import coursier.version.VersionConstraint
 
 object PomParserTests extends TestSuite {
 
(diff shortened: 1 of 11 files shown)


alexarchambault (approved):
Thanks!

PR #3772, opened 2026-08-06
Add non-regression tests for many FileCache aspects

alexarchambault:


alexarchambault:
Don't have time to review in detail right now, but these probably can't hurt

PR #3777, opened 2026-08-13
Publish empty doc JARs

alexarchambault:
Not to flood Maven Central with heavy and almost useless JARs

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -982,16 +982,6 @@ trait Cli extends CsModule
       Name.IMPLEMENTATION_VENDOR.toString  -> "io.get-coursier"
     )
   }
-  def docJar = Task {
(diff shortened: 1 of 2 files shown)


lefou:
I find empty doc jars way more useless then one with actual docs.

alexarchambault:
I know, but with Sonatype new publishing limits, I'd rather not publish them for now, until they can be made less heavy.

lefou:
> I know, but with Sonatype new publishing limits, I'd rather not publish them for now, until they can be made less heavy.

I'd say, it's Sonatype own rule to add a docJar in the first place. They could raise an issue against Scaladoc tooling, if they don't like the size of them. Adding empty doc jars is a disservice to the whole community.

PR #3837, opened 2026-09-09
Make dependency management key substitution deterministic

alexarchambault:


alexarchambault:
(Agents stumbled upon that while working on https://github.com/coursier/coursier/pull/3742)

PR #3842, opened 2026-09-09
Allow User Agent to be customized

eed3si9n:
## Problem/Solution
We want to debug the Maven Central usage by overriding the User-Agent.


alexarchambault:
@eed3si9n What do you mean, by "[you] want to debug the Maven Central usage"?

eed3si9n:
Hi. Here's the relevant discussion, basically trying to subdivide Coursier users based on different scenarios I think - https://github.com/coursier/coursier/discussions/3841.

alexarchambault (approved):
Seems that can be useful, thanks!

PR #3844, opened 2026-09-10
Update default user agent and give more options to users for customizing it

alexarchambault:


eed3si9n:
@alexarchambault Could we get a new release of Coursier with this change plz?

PR #3869, opened 2026-09-22
Switch `cats-free` at `_2.13` to `cats-core`; add a check for cross-version dependencies

### LINKED ISSUE #3865: 2.1.25: all resolution commands fail with NoSuchMethodError — released jar bundles a mismatched cats
# Steps to reproduce

curl -fLo coursier.jar https://github.com/coursier/coursier/releases/download/v2.1.25/coursier.jar
java -jar coursier.jar fetch foo:bar:1.0

## Observed Behavior

Exception in thread "main" java.lang.NoSuchMethodError: 'cats.Align cats.implicits$.catsStdInstancesForList()'
	at coursier.cli.options.CacheOptions.params(CacheOptions.scala:162)
	at coursier.cli.options.CacheOptions.params(CacheOptions.scala:92)
	at coursier.cli.resolve.SharedResolveParams$.apply(SharedResolveParams.scala:32)
	at coursier.cli.fetch.FetchParams$.apply(FetchParams.scala:42)
	at coursier.cli.fetch.Fetch$.run(Fetch.scala:96)
This happens during argument parsing, before any network access. launch and resolve fail with the same error via their own *Params$.apply frames. --help and version still work, so the jar loads fine.

Seems to be independent of jdk version

Gedochao:
Fixes https://github.com/coursier/coursier/issues/3865

Copied the cross-Scala-version check from https://github.com/VirtusLab/scala-cli/blob/main/.github/scripts/check-cross-version-deps.sc


Gedochao:
~~Hmm... it seems that broke some Scala 2 checks... lemme take another look~~
Actually, just a bad commit on my side. Should be fixed.

alexarchambault:
Thanks @Gedochao, I already had a fix for that locally, pushed to https://github.com/coursier/coursier/pull/3879 now

